PR #3891, opened 2014-12-30
Cache Record Constructor in CursorImpl

arnaudroger:
When I executed the benchmark with sfm vs generated record the sfm was still faster because of the constructor lookup. 
Note that I cache the RecordInstantiator in the CursorImpl but very likely that ideally you would want to generate the RecrodInstantiator as par as generated the specified record.
That move the perf with the specified record higher that sfm as expected.


lukaseder:
I've slightly adapted the implementation. There is also room for broader re-use of this new "`RecordFactory`".

lukaseder:
This will also be useful for #3954

Issue #3923, opened 2015-01-09
Four octects are sent to Postgres for each negative byte in a byte[]


gortiz:
When JOOQ sends array of bytes to Postgres, it uses the octal format. To do the conversion from binary byte to octal, it uses java.lang.Integer#toOctalString(int) and the byte is automatically casted to integer. It works fine with positive bytes, but not with negative ones.

For example, when [-1] is inserted, [377] should be sent to Posgresql. Instead of that, Java transforms the byte -1 to the integer -1, represented in octal as 377 377 377 377, and jOOQ sends the array [377, 377, 377, 377]!

lukaseder:
This issue is dealt with in the relevant pull request here: #3924

PR #3924, opened 2015-01-09
[#3923] Four octects are sent to Postgres for each negative byte in a byte[]

gortiz:
When JOOQ sends array of bytes to Postgres, it uses the octal format. To do the conversion from binary byte to octal, it uses java.lang.Integer#toOctalString(int) and the byte is automatically casted to integer. It works fine with positive bytes, but not with negative ones.

For example, when [-1] is inserted, [377] should be sent to Posgresql. Instead of that, Java transforms the byte -1 to the integer -1, represented in octal as 377 377 377 377, and jOOQ sends the array [377, 377, 377, 377]!

### DIFF
--- a/jOOQ/src/main/java/org/jooq/impl/DefaultBinding.java
+++ b/jOOQ/src/main/java/org/jooq/impl/DefaultBinding.java
@@ -767,7 +767,8 @@ public class DefaultBinding<T, U> implements Binding<T, U> {
 
         for (byte b : binary) {
             sb.append("\\\\");
-            sb.append(leftPad(toOctalString(b), 3, '0'));
+            int octal = b & 0x000000ff;
+            sb.append(leftPad(toOctalString(octal), 3, '0'));
         }
 
         return sb.toString();


lukaseder:
Thank you very much for this pull request! Could you maybe show an example interaction with the database where this went wrong before and how your fix actually fixes the issue?

We don't host the integration tests on GitHub anymore, but it would still be interesting to have a hint, so we can cover this case in the integration tests...

gortiz:
We detected the bug in ToroDB (https://github.com/torodb/torodb). We are using arrays with 20 bytes to represent the MongoDB type ObjectId. We noticed that the insertion works fine (or at least no error was thrown) but when we read the stored value, sometimes the it had more than 20 elements.

After some analysis, we discovered this small bug in jOOQ. It is totally repeatable. You can create a table with a bytea column, then insert an array of bytes with at least one negative byte and then you can read the row from the Postgres shell. You will see that there are more bytes than expected.

The bug was very difficult to find because the code looked fine, but the automatic conversion from byte -1 to int -1 produces unexpected results.

For example:

```java
byte b = -1;
int octal = b & 0x000000ff;
            
System.out.println(StringUtils.leftPad(Integer.toOctalString(b), 3, '0'));
System.out.println(StringUtils.leftPad(Integer.toOctalString(octal), 3, '0'));
```

lukaseder:
Aha, I get it now. Yes, nice catch.

lukaseder:
The issue only arises when inlining bind values

lukaseder:
We'll merge this issue to versions:

- 3.5.2 (#3925)
- 3.4.5 (#3926)
- 3.3.5 (#3927)

Issue #2664, opened 2013-08-01
Add flag to Loader API to enable batch execution


lukaseder:
Currently, the `Loader` API just inserts row by row, which is far from optimal for large loads. There should be a flag to enable batch execution.

In some contexts, there may be conflict between batch execution and other flags, such as "on duplicate key" flags and error handling. This needs to be carefully reviewed

----
See also this thread:
- https://groups.google.com/d/msg/jooq-user/BElK0u4dOsk/g3LeUoirHgAJ

PR #3975, opened 2015-01-22
Add support for bulk insertion in the Loader API

detlef-brendle:
Implementing a bulk processing has shown, that our data processing is a lot faster.
For a h2 there comparison is  90sec vs.	20sec
and for a mssql 43min vs. 5min.

Especially when communicating to a remote instance, sending as much data as possible makes a huge different.

Please check if you can implement something like I did and bring it to a productive state.
Currently it is just a POC.

Thanks.
Detlef

### DIFF
--- a/jOOQ/src/main/java/org/jooq/LoaderOptionsStep.java
+++ b/jOOQ/src/main/java/org/jooq/LoaderOptionsStep.java
@@ -211,4 +211,17 @@ public interface LoaderOptionsStep<R extends TableRecord<R>> extends LoaderSourc
     @Support
     LoaderOptionsStep<R> commitNone();
 
+    /**
+     * Run insert in bulk mode.
+     * Specially useful for insert statements.
+     * In that case, a single statement can insert multiple records in one go.
+     * There might be a limit of markers (i.e. MSSQL has 2000). In such cases the
+     * inseret statement must be executed blockwise.
+     * @param markerLimit The limit of markers. Depends on the dialect.
+     * @return
+     */
+    @Support
+    LoaderOptionsStep<R> withBulkInsert(int markerLimit);
+
+
 }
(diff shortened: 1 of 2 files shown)


lukaseder:
Thank you very much for your contribution. We will review it shortly. Perhaps, this is already partially covered by another feature request on our roadmap: #2664. In general, batch processing can already have a tremendous effect on many RDBMS's performance:
http://blog.jooq.org/2014/01/16/what-you-didnt-know-about-jdbc-batch/

But you're right. Grouping several `INSERT` statements into one will also improve performance

detlef-brendle:
Hi Lukas,

thanks for making bulk operation work.
I tried it with the solution you implemented.
I set the batch property using batchAll() on the config and imported my data.
It works fine, but compared to my solution that needs 5 mins it took 12 mins. 
So maybe you could optimize it even further.

But at least it fastens the import quite a bit.

Thanks,
Detlef
PS : In case you need more info about our import I can help you.

lukaseder:
@detlef-brendle Thanks a lot for the feedbaack, this is greatly appreciated.

I didn't forget about your change request, but I wanted to implement the more complex change (batch) first to be sure that we'll have integration tests ready when we implement the additional bulk options. It should be possible to combine batch+bulk+commit flags.

All features will be added to jOOQ 3.6

> PS : In case you need more info about our import I can help you.

Sure. If you can show the table's DDL and describe your data set (number of rows, total size), that would be very useful

lukaseder:
All three performance tuning measures are now in place:

- The bulk statement size. This specifies how many rows will be inserted in a single bulk statement / multi-row INSERT statement.
- The batch statement size. This specifies how many bulk statements will be sent to the server as a single JDBC batch statement.
- The commit size. This specifies how many batch statements will be committed in a single transaction.

This way, users can fine-tune loading on a per-load basis, depending on their RDBMS and JDBC driver versions and their actual data.

PR #3976, opened 2015-01-23
Code generator does not apply binding to Routines

fuwjax:
copied binding support from generateUDT to generateRoutine

### DIFF
--- a/jOOQ-codegen/src/main/java/org/jooq/util/JavaGenerator.java
+++ b/jOOQ-codegen/src/main/java/org/jooq/util/JavaGenerator.java
@@ -2589,16 +2589,15 @@ public class JavaGenerator extends AbstractGenerator {
             final String paramName = parameter.getName();
             final String paramComment = StringUtils.defaultString(parameter.getComment());
             final String isDefaulted = parameter.isDefaulted() ? "true" : "false";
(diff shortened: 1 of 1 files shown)


lukaseder:
Thank you very much for your contribution. We'll review it shortly and keep you updated on this issue. 

In the meantime, for us to be able to merge your pull request and to continue dual-licensing jOOQ, we need you to agree on the transfer of rights of your contribution:

lukaseder:
Thanks for the signature. I'm sorry for the delay, we didn't get a notification. This should now be safe to merge.

lukaseder:
This is already correct in jOOQ 3.6

fuwjax:
Thanks Lukas, I discovered it was missing when I was trying to fix my issues with "timestamp with time zone". Turns out most of those issues were that a broken flyway migration wasn't failing the build correctly. Also, in the debug output for jOOQ a java.sql.Timestamp is printed in the local timezone, but on the db server, it is stored correctly. So, it turns out I didn't need this fix after all.

Just thought I'd point out that in addition to being a fantastic db connection library, jOOQ is excellent OSS. It's the only 3rd party dependency we use that works with the "import" scope in maven. I had it built from source and debugging a custom patch in minutes. Thank you for maintaining such high quality on the project.

PR #3988, opened 2015-01-26
Incorrect IntelliJ directory listings in .gitignore

brmeyer:
Trivial, I know, but this corrects .gitignore when the entire jOOQ project is imported into IntelliJ


lukaseder:
Yep you're right. It's been a while since anyone here used IntelliJ with jOOQ...

PR #3992, opened 2015-01-26
Add <T> DSL.coalesce(Field<T>, T)

brmeyer:
Additional convenience method for something like `select sum(col)`.  Without this method, I have to make assumptions about the return value (BigDecimal) and cast.  With the method, any Field/Value pairing will have the correctly-typed return.

### DIFF
--- a/jOOQ/src/main/java/org/jooq/impl/DSL.java
+++ b/jOOQ/src/main/java/org/jooq/impl/DSL.java
@@ -7418,6 +7418,17 @@ public class DSL {
     }
 
     /**
+     * Gets the Oracle-style <code>COALESCE(field, value)</code>
+     * function.
+     *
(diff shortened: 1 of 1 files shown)


lukaseder:
I must admit, I'm not sure anymore why we didn't include that method overload. It obviously seems useful. In the short run, you can use `DSL.nvl(Field<T>, T)` instead. It does the same thing.

In order to be able to merge your pull request and future pull requests, I will need to ask you to sign the following agreement that will allow us to dual-license your contribution and ship it also in the closed-source, commercial version of jOOQ:

If that is an impediment for you and/or for your employer(s), then we'll just implement these feature requests ourselves. Let me know what's easiest for you.

brmeyer:
Signed

lukaseder:
Thank you very much

Issue #3869, opened 2014-12-17
Include third-party dependencies via the Maven Shade Plugin


lukaseder:
Currently, we have a couple of dependencies that we internalised into jOOQ-local packages, e.g. commons-lang.

In principle it would be better to include the actual dependencies in order to automate upgrades etc. There is the Maven Shade Plugin that takes care of creating an "uber jar", which can also relocate the dependencies' packages in order to avoid the usual conflicts with transitive dependency hell:
http://maven.apache.org/plugins/maven-shade-plugin

thiloplanz:
How important is it for people to have jOOQ work without any extra jar file dependencies? 
The alternative would be to just declare the dependencies and have them be supplied by the user. 

Should not be too much trouble for anyone, no trouble for Maven users, makes for possibly slightly lighter artefacts (if things like commons-lang are already used by the application), OTOH also makes for potential version conflicts with unstable dependencies.

Personally, I don't like "uber jars" for anything except executable jar files (and there I love them, though recently more in the capsule variety).

Also note that jOOQ cannot be used without supplying at least one additional jar file anyway (the JDBC driver you want to use).

What are these dependencies? Are they currently stripped down to only the parts that jOOQ requires (that would mostly remove the slightly lighter objection above)?

lukaseder:
> How important is it for people to have jOOQ work without any extra jar file dependencies?

It is quite important to a couple of our customers.

> if things like commons-lang are already used by the application

We only use `StringUtils` from there. The overhead is negligible.

> OTOH also makes for potential version conflicts with unstable dependencies.

Yes. See some competitor products who have dependency issues throughout the dependency chain up into Spring Data, etc.

> What are these dependencies?

- JSON Simple (parts)
- CSV Parser (parts)
- Apache Commons (1 class)
- jOOU
- jOOR

They're all currently manually included. Note that we really only need parts of those dependencies, which is why we don't want to create a full dependency.

Perhaps we might create two distributions. One with included dependencies, and one that governs everything via Maven.

sytolk:
It's a very bad practice to include dependencies in jooq package. Тhis may cause conflicts in some projects that use original ones with different version like dependency. Its need to remove it from jooq and include all dependency in maven pom.xml 
For not maven users it`s have Intellij IDEA/Eclipce artifact creation with included dependencies in jooq.jar or maven plugins to do this.. but it`s a stupid I can`t think more about these users..

lukaseder:
[...]

PR #4219, opened 2015-04-20
Implements an instance toString() method on JSONObject.

alokmenghrajani:
http://www.jooq.org/javadoc/latest/ is not very clear about JSONObject
taking a map as a parameter in the constructor but calling toString on
the resulting object won't return a JSON string.

I feel things can be improved by either adding something to the documentation
or by having a toString() method which calls toJSONString. This commit
implements the latter.

### DIFF
--- a/jOOQ/src/main/java/org/jooq/tools/json/JSONObject.java
+++ b/jOOQ/src/main/java/org/jooq/tools/json/JSONObject.java
@@ -123,6 +123,10 @@ public class JSONObject extends HashMap{
         return sb.toString();
     }
 
+    @Override public String toString() {
+        return toJSONString(this);
+    }
+
     /**
      * Escape quotes, \, /, \r, \n, \b, \f, \t and other control characters
      * (U+0000 through U+001F). It's the same as JSONValue.escape() only for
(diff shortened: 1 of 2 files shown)


lukaseder:
Hmm, interesting. Thanks for the contribution!

But instead of improving our version of the library, we might just upgrade to the latest version of JSONSimple, which already includes this change:
https://github.com/fangyidong/json-simple/blob/tag_release_1_1_1/src/main/java/org/json/simple/JSONObject.java

(discussions about embedding dependencies can be found here, for instance: https://github.com/jOOQ/jOOQ/issues/3869)

lukaseder:
Upgrading JSONSimple broke our integration tests, so we cannot do that. It might be a good time to think about replacing the dependency entirely by the standard jsonp in jOOQ 4.0 (#4220).

I'll merge your pull request as it adds immediate value without any risks. The merge can be done without signing our transfer of rights agreement, as this third-party section of the code is ASL 2.0 only licensed.

alokmenghrajani:
:+1: 
Do you think JSONArray should also get the same change?

lukaseder:
Yes, good idea. I didn't even check that

lukaseder:
Done

PR #4397, opened 2015-06-25
Update Convert.java

deodeveloper:
boolean value "t" or "f" returned by Vertica and it is missing in converter, hence source code generation is faulty which always assigns true to all Table entity fields


deodeveloper:
duplicate

lukaseder:
Duplicate of #4398

PR #4398, opened 2015-06-25
Update Convert.java

deodeveloper:
boolean value "t" or "f" returned by Vertica and it is missing in converter, hence source code generation is faulty which always assigns true to all Table entity fields


lukaseder:
Thank you very much for your pull request and sorry for the delay. We'll be looking into this shortly.

lukaseder:
This certainly makes sense. Other databases (such as PostgreSQL) also return "t" and "f" string literals for booleans. I've also included capital letters for consistency

PR #4407, opened 2015-07-14
Moved jOOQ Spring Boot to dedicated configuration class.

jgon:


lukaseder:
Thanks for your contribution. Could you maybe add a little description describing the motivation of this change? How is it better than the previous version?

jgon:
While using the jOOQ Spring Boot sample in an application I moved the configuration into a dedicated  class for clarity purposes.

jgon:
Lukas, thank you for your reply.

Not sure if this is the best way/place to ask about why the build with my pull request failed in the first place.

I noticed that it was failing in the jOOQ module while my changes affected the examples only.

Is there anything I did wrong on my side?

Kind regards,

Jack

lukaseder:
> While using the jOOQ Spring Boot sample in an application I moved the configuration into a dedicated class for clarity purposes.

Alright, fair enough. We can merge this.

> I noticed that it was failing in the jOOQ module while my changes affected the examples only.
>
> Is there anything I did wrong on my side?

I don't think you did. These failing build messages in PR's are new. We were getting E-Mails behind the scenes because the out-of-the-box travis installation regrettably still uses Java 1.8.0_31, whose compiler has tons of bugs related to generics and type inference... I guess we'll turn off public travis for the time being...

PR #4423, opened 2015-07-22
Relax type constraint of DSL.shl() and DSL.shr()

kxbmap:
Operands of `shl` and `shr` don't need to be same type.


lukaseder:
Thank you for your suggested change. In order to be able to dual-license your contribution to both the jOOQ Open Source Edition and the commercial distributions, we would need you to sign the following transfer of rights agreement:

Please, let me know if this is OK with you, and if you have any specific questions regarding dual-licensing.

kxbmap:
I signed to it.

lukaseder:
Thank you very much, and thanks again for this useful contribution

lukaseder:
Marking this as an "incompatible change" as `Field` method signatures have been changed incompatibly ([but along what's acceptable for jOOQ minor releases according to our compatibility guidelines](http://www.jooq.org/doc/latest/manual/getting-started/semantic-versioning/))

PR #4424, opened 2015-07-22
Use ArrayDeque as a stack instead of java.util.Stack

kxbmap:
`ArrayDeque` is faster than `Stack`.
All of method calls of `Stack` are synchronized, but in these codes, it looks like not necessary.


lukaseder:
Nice catch! Thanks for this improvement. You're right, I suspect `Stack` was used out of old habit. `ArrayDeque` is an equivalent replacement for it.

PR #4430, opened 2015-07-27
Fix routine parameter code generation for types with a binding

lightoze:


lukaseder:
Thank you for your suggested change. In order to be able to dual-license your contribution to both the jOOQ Open Source Edition and the commercial distributions, we would need you to sign the following transfer of rights agreement:

Please, let me know if this is OK with you, and if you have any specific questions regarding dual-licensing.

lightoze:
I've accepted the agreement. Where is the mentioned “Help” or “About” file located? :)

lukaseder:
Thank you very much. I'll review the change ASAP and let you know about my findings.

> Where is the mentioned “Help” or “About” file located? :)

Excellent question. There is a contributor listing in the parent pom.xml, and we usually add/maintain `@author` tags in the relevant classes, but we should definitely also maintain a complete "about" file. I've created an issue to fix this: #4431. Thanks for pointing this out to us!

lightoze:
OK, hope to see it merged soon - the bug leads to uncompilable generated code for me.

lukaseder:
Interesting, how can the compilation error be reproduced? E.g. what DDL and code generator configuration lead to it?

lightoze:
I have a function with JSONB parameter which has a custom binding. Generated code is
```
Parameter<JsonElement> OBJ = createParameter("obj", org.jooq.impl.DefaultDataType.getDefaultDataType("jsonb"), false);
```
So `Parameter<Object>` cannot be assigned to `Parameter<JsonElement>`.

lukaseder:
Ok, thanks for the additional info

PR #4459, opened 2015-08-05
Result.formatHTML() doesn't escape HTML content, which might produce XSS vulnerabilities

artbristol:
Fix HTML injection

Used commons-lang3's `StringEscapeUtils` to escape field names and values in `ResultImpl.formatHTML`. Separated into another class to allow unit testing.

### DIFF
--- a/jOOQ-codegen-maven/pom.xml
+++ b/jOOQ-codegen-maven/pom.xml
@@ -6,7 +6,7 @@
     <parent>
         <groupId>org.jooq</groupId>
         <artifactId>jooq-parent</artifactId>
-        <version>3.6.2</version>
+        <version>3.6.3-SNAPSHOT</version>
     </parent>
     
     <groupId>org.jooq</groupId>
--- a/jOOQ-codegen/pom.xml
+++ b/jOOQ-codegen/pom.xml
@@ -7,7 +7,7 @@
     <parent>
         <groupId>org.jooq</groupId>
         <artifactId>jooq-parent</artifactId>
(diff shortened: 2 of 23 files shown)


lukaseder:
Thanks for the suggested fix. I understand that it might indeed be better to escape all HTML content to prevent XSS and other issues. However, this fix is a breaking change for some users that may have depended on the existing "functionality" - e.g. where `<strong>` and other markup was desired. We thus won't merge this fix back to pre-3.7 branches.

Note that the additional dependency will have to be resolved - we strive to have absolutely no unnecessary dependencies to other libraries from jOOQ (see http://www.jooq.org/doc/latest/manual/getting-started/dependencies). Also, I think from the implemented escaping:

```java
    public static final CharSequenceTranslator ESCAPE_HTML4 = 
        new AggregateTranslator(
            new LookupTranslator(EntityArrays.BASIC_ESCAPE()),
            new LookupTranslator(EntityArrays.ISO8859_1_ESCAPE()),
            new LookupTranslator(EntityArrays.HTML40_EXTENDED_ESCAPE())
        );
```

Only the first one (`BASIC_ESCAPE`) should really be applied in case users treat resulting HTML as XML.

I'll merge this and re-adapt according to the above. Thanks again for your contribution!

lukaseder:
(reverting again, because the fix won't be applied to the 3.6.0 branch)

lukaseder:
... in fact, there's already an `ResultImpl.escapeXML()` method that we can reuse...

Issue #3883, opened 2014-12-23
Redesign the code generation API


lukaseder:
The code generation "API" has grown organically over the past releases of jOOQ 2.x and 3.x. Users desiring to override some behaviour rely heavily on its internals.

This may seem acceptable as we're talking about "only" code generation, but it is still hard to maintain user-defined behaviour in a stable way across jOOQ minor releases. The risk of breaking things between minor releases is substantial.

We should definitely re-write the entire API in a much more object-oriented and extensible fashion in jOOQ 4.0.

---

Feedback from the user group (https://groups.google.com/forum/#!topic/jooq-user/aZ0CJwUfdkM):
### 1. Target languages and forms
- Java and Scala (current support)
- Ceylon, Kotlin, Clojure
- Interesting thoughts by customer: JavaScript (and TypeScript, ScalaJS, Ceylon) for client-side records, in case database active records should be made available also on the client (with custom sync protocol)
### 2. Templating language

Alternatives are
- Java (as it is today)
- Xtend
- Scala
- "Pure" templating languages like mustache, velocity
### 3. Style
- Little interest in customisation
- Apply Oracle's JDK styleguide
### 6. Custom code
- Repositories / Services / DAOs, possibly with custom (C)DI annotations
- Custom predicate builders in DAOs
- Data type management (see e.g. https://github.com/jOOQ/jOOQ/issues/10231#issuecomment-635848819)

---

See also:
- https://groups.google.com/forum/#!topic/jooq-user/aZ0CJwUfdkM
- #3882 
- #4384
- various other issues

napperley:
Will Kotlin support be part of the redesign? If that is the case then it would be very helpful to have Kotlin code generated that is **Kotlinic** (idiomatic Kotlin/the Kotlin way), and follows the official [Kotlin Coding Conventions](https://kotlinlang.org/docs/reference/coding-conventions.html).

FYI there is a library for generating Kotlin code called [Kotlin Poet](https://github.com/square/kotlinpoet) which is maintained by Square Inc.

lukaseder:
> Will Kotlin support be part of the redesign?

@napperley It says right there: *"1. Target languages and forms / Java and Scala (current support) / Ceylon, Kotlin, Clojure"*

> FYI there is a library for generating Kotlin code called Kotlin Poet which is maintained by Square Inc.

[...]

PR #4462, opened 2015-08-09
Tests for java code generator

abashev:
We are using java generator for generating DAO with Spring annotations and before version 3.5.something everything was fine - it was easy to update generator with new version and maintain it as our own class. But now we not able to migrate our changes because maintain our code and to have updated generator almost impossible. So I've decided to refactor generator to make it more supportable and also be able to generate classes with Spring annotations.
And before the start I need to add a few tests to be sure that I'm not breaking anything. This PR contains only test code and don't affect generator itself.


lukaseder:
Thank you for your contribution.

Unfortunately, I am inclined to reject it for these reasons:

1. Unit testing the source code generator on a string comparison basis won't scale for us. The amount of features it currently contains is too high, we won't be able to maintain such a fine-grained test suite.
2. Trivial changes (e.g. whitespace adaptations) are perfectly OK, even between patch releases. This kind of test will break all the time.
3. While I certainly appreciate the intent, we won't be covering even 10% in a foreseeable time, at least in publicly available test suites. Having such low coverage in a test suite might make people think that jOOQ's code generator is very poorly tested, which it is not. We have extensive, non-open source test suites that we run on all databases to ensure the source code generator produces compilable code.

I'm aware of the pain points that might arise from upgrading between jOOQ minor releases, when users have heavily adapted the source code generator, but I'm afraid that we currently cannot offer reasonable solutions here, short of offering you to look into possible feature requests to get your features working. What kind of extensions have you implemented that broke between 3.5 and 3.6?

abashev:
Lukas, it is great that you have tests for code generator, so I leave these changes only for my own.

Spring is our base framework and we are doing minor modifications to make DAO more spring-aware
1. We add annotation @Repostiry for all DAOs'
2. Removing default constructor and add @Autowire for constructor with configuration
3. Add two methods for creating new record objects

And right now with Scala things it is really painful so I've decided to implement mediator pattern for generating DAO classes - describe DAO generation as chain of different calls - generateAnnotations, generateConstructors and so on, and push in different implementations - ScalaStrategy, JavaStrategy, SpringStrategy. What do you think?

lukaseder:
> Lukas, it is great that you have tests for code generator, so I leave these changes only for my own.

Yep, don't worry about it :) we'll adapt any unforeseen breaking change, if any.

Interesting ideas about Spring - we should definitely add some improvements like the ones you suggested. I'm wondering if `DAO` should be replaced by a type that extends spring data's `CrudRepository`? We could generate repositories rather than `DAO`s...

In any case, would you mind sharing an example generated DAO of yours in code form, just to be sure I get this right?

> And right now with Scala things it is really painful so I've decided to implement mediator pattern for generating DAO classes - describe DAO generation as chain of different calls - generateAnnotations, generateConstructors and so on, and push in different implementations - ScalaStrategy, JavaStrategy, SpringStrategy. What do you think?

I'm actually not sure what you mean here. You mean, it's painful that the same generator produces all the code?

abashev:
CrudRepository is specific for spring-data project but we are using direct jooq access and I don't think it is possible to replace runtime spring-date generator with jooq internal or don't it make any sence. @Repository annotation is just a marker for spring beans + converter for DataAccessExceptions

there is our code - https://gist.github.com/abashev/06a1fad8a07cf8d1402b

> I'm actually not sure what you mean here. You mean, it's painful that the same generator produces all the code?

It is hard to maintain our changes with so complex code - before we just commented lines and update version but right now we need to find exact place, understand is it for scala or java, and what is it doing, and what is more important ref() calls have 'protected' scope.

lukaseder:
Thank you for your feedback. I see your point. Yes, supporting `@Repository` and `@Autowired` annotations in this way certainly makes sense. We aren't opposed to creating dependencies on popular third-party software in generated code, as that code is entirely in your control. I've created an issue to track this idea: https://github.com/jOOQ/jOOQ/issues/4470.

You're right, `ref()` calls (and many others) may be `protected` or even package-private. The code generation API unfortunately has never been designed for extension. We'll certainly aim to remedy that in the next major release (https://github.com/jOOQ/jOOQ/issues/3883), but right now, we can only look into improving things little by little...

abashev:
Don't mind if I'll improve code generator a little? 😎

lukaseder:
Sure! Feel free to make suggestions. But I may reject :)

abashev:
Lukas, I really tried but code generator has so many internal references that really hard to brake without rewriting Java generator from the scratch, for example getJavaObject was used inside DAO generation and it is first candidate for moving out but it has dependency on `strategy` and `is scala flag`. So it takes too much time and too many changes in your product :(

lukaseder:
Alexey, thank you very much for the feedback. I must admit, I'm not surprised. The current code generator really isn't designed to be extended arbitrarily. Such design is rather difficult to achieve, and it must be done from the beginning. Our generator just grew into what it is today.

Again, we're hoping to re-design it completely in jOOQ 4.0 (https://github.com/jOOQ/jOOQ/issues/3883), but until then, I suspect we cannot really make any improvements.

PR #4505, opened 2015-08-28
Add Maven plugin skip property

lightoze:


ooxi:
Could you please elaborate your proposed change? Note that you can always skip the execution of an inherited plugin [by defining a non existing phase](https://thomaswabner.wordpress.com/2010/02/16/howto-disable-inherited-maven-plugin/)

lightoze:
Yes, but when you want to switch plugin execution on/off for different profiles and different modules of multi-module project, and when you don't want to change any other parameters - having skip property is much more convenient.

lukaseder:
I agree that a ad-hoc skipping flag can be useful. I personally use the `-Dmaven.test.skip=true` flag all the time. I wonder what would be the most idiomatic way to do this. Just adding a `<skip/>` element to the plugin seems a bit weird, as you might as well just comment the plugin declaration with `<!-- .. -->`. How do other Maven plugins do this? Also note, that jOOQ's code generator isn't really a Maven plugin. The plugin is just convenience for the standalone code generator...

ooxi:
Other plugins I use have a `skip` boolean property

lightoze:
With `<skip>` element you will be able to bind it e.g. to `jooq.skip` property in your parent POM in use that property all around (e.g. `-Djooq.skip=true`). Many plugins do this out of the box, so that default value for `<skip>` is an environment property.

lukaseder:
OK, I'm convinced. Thanks again for your contribution!

0x6e6562:
This is a really cool feature - is there any roadmap to back port this to 3.6.x (for people still stuck on Java 7)?

lukaseder:
@0x6e6562: There wasn't. But there is now: https://github.com/jOOQ/jOOQ/issues/4701 :) I suspect that this is a trivial enough feature for it to be backported to a patch release.

Note that the commercial jOOQ 3.7 editions also still support Java 7:
http://www.jooq.org/download/versions

0x6e6562:
I saw that the commercial version had a JDK7 download option, but the issue I ran into was with the maven generator plugin. There may have been a proper solution for this, but I needed an expedient solution and dropping back down to 3.6 seemed to be quickest solution. Maybe I didn't install the commercial version of the maven plugin properly.

PR #4562, opened 2015-09-16
Do not escape dot in table names during code generation, but use capital case after it

begray:
In current version of jOOQ code generation table names like `user.users`, `user.achievements` are transformed to `User_2eusers`, `User_2eachievements` in class names (e.g. for tables, pojos and daos). 

Included patch changes this behaviour to: 
```user.users -> UserUsers```
```user.achievements -> UserAchievements```

i.e. dot is not included into the output, but instead next word is capitalized, producting nice pascal cased names.

### DIFF
--- a/jOOQ-codegen/src/main/java/org/jooq/util/GenerationUtil.java
+++ b/jOOQ-codegen/src/main/java/org/jooq/util/GenerationUtil.java
@@ -178,19 +178,29 @@ class GenerationUtil {
             return "_";
         }
 
+        boolean uppercaseNext = false;
         for (int i = 0; i < literal.length(); i++) {
             char c = literal.charAt(i);
 
             if (!Character.isJavaIdentifierPart(c)) {
-                sb.append(escape(c));
+                if ('.' != c)
+                    sb.append(escape(c));
             }
             else if (i == 0 && !Character.isJavaIdentifierStart(literal.charAt(0))) {
                 sb.append("_");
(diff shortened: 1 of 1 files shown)


lukaseder:
This is an interesting idea, thank you very much for your suggestion and for your contribution.

We could think of a couple of "special characters" that behave like the underscore when moving from `under_score` naming schemes to `PascalCase`.

Before we can proceed with merging your contribution, we would need you to sign our transfer of rights agreement, which allows us to continue dual-licensing all of jOOQ's code:

Thank you very much for your understanding

begray:
I've signed the agreement.

lukaseder:
Thank you very much. I've noticed that the implementation is probably not at the correct spot. The conversion from `under_score` to `PascalCase` must take place in a `GeneratorStrategy`. This way, users can override the behviour as needed. Implementations in `GenerationUtil` can no longer be overridden, which might be undesireable.

While I agree with escaping the period to an underscore, I think we should leave the `convertToIdentifier()` method as it is...

lukaseder:
I have moved your suggested implementation from `GenerationUtil` to `DefaultGeneratorStrategy`. The new implementation has the same desired effect:

- `user.users` -> `UserUsers`
- `user.achievements` -> `UserAchievements`

Thanks again for your report and contribution!

begray:
Great job Lukas, thanks :)

PR #4619, opened 2015-10-08
Let SQLTable constructor pass SQL string to AbstractTable parent constructor

sjachim:
…was as how SQLField does it taking value from delegate's toString(). 
It's not perfect, but at least it fixes the bug ("table" is not a good name) and it's consistent.


lukaseder:
Thank you very much for your suggestion. That looks reasonable on a high level. Not sure if there might be some reason why it hadn't been done like in `SQLField` - I'll investigate.

In the meantime, may I ask you to sign our transfer of rights agreement:

We need this on every contribution, even if it's a small one.

lukaseder:
Thanks for signing. The integration test suite didn't mind the change too much and it makes total sense. Thanks again for your contribution!

PR #4742, opened 2015-11-16
Code Quality fix - Printf-style format strings should not lead to unexpected behavior at runtime

javedkansi:
This pull request is focused on resolving occurrences of Sonar rule squid:S2275 - “Printf-style format strings should not lead to unexpected behaviour at runtime”. You can find more information about the issue here: https://dev.eclipse.org/sonar/rules/show/squid:S2275?layout=true

Please let me know if you have any questions.

Javed.

### DIFF
--- a/jOOQ/src/main/java/org/jooq/impl/AbstractRecord.java
+++ b/jOOQ/src/main/java/org/jooq/impl/AbstractRecord.java
@@ -1062,14 +1062,14 @@ abstract class AbstractRecord extends AbstractStore implements Record {
             throw new NullPointerException();
         }
         if (size() != that.size()) {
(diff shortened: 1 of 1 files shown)


ooxi:
I'm not quite sure this is the desired behaviour. `"\n"` evaluates to `"U+000A"` while `"%n"` could evaluate to everything (depending on the `line.separator` system property).

Could you explain why new lines lead to unexpected behaviour at runtime?

lukaseder:
Thank you for your suggestion.

While static code analysis tools like Sonar are capable of finding *some* issues, those issues are often very trivial. In this case, we might be superficially "fixing" line separators in a single occurrence of a `String.format()` call (and I'm not even sure if for the better), but what about all the other line separators scattered across the code base that aren't part of `String.format()` calls?

We could fix those, too, but I really feel that I'd rather spend my time on more important tasks, or real bugs...

PR #4743, opened 2015-11-16
Code Quality fix - Lazy initialization of "static" fields should be "synchronized

javedkansi:
This pull request is focused on resolving occurrences of Sonar rule squid:S2444 - “Lazy initialization of "static" fields should be "synchronized”. You can find more information about the issue here: https://dev.eclipse.org/sonar/rules/show/squid:S2444?layout=true

Please let me know if you have any questions.

Javed.


lukaseder:
Thanks for your suggestions. jOOQ-meta is mainly designed to be used with jOOQ-codegen, which is single-threaded.

It's true that some issues might be fixed superficially by adding `synchronized` keywords on methods, but that would be the wrong approach in my opinion, as more important issues might still arise from a fundamentally single-threaded design.

PR #4943, opened 2016-01-17
Generator extending JavaGenerator can override applied filters.

jakubkulhan:
Adding `JavaGenerator.AvoidAmbiguousClassesFilter` to `Database` is hardcoded in `JavaGenerator.generate()` method. This prohibits generator strategy from intentionally generating the same names. I propose to move adding filter to overridable method.

For example, I have 2 tables `transactions` & `transactions_history` both having `state` enum column with the same definition - the columns represent the same thing. Custom generator strategy intentionally generates the same class name `TransactionState` for both columns.

### DIFF
--- a/jOOQ-codegen/src/main/java/org/jooq/util/JavaGenerator.java
+++ b/jOOQ-codegen/src/main/java/org/jooq/util/JavaGenerator.java
@@ -199,7 +199,7 @@ public class JavaGenerator extends AbstractGenerator {
         this.catalogVersions = new LinkedHashMap<CatalogDefinition, String>();
 
         this.database = db;
-        this.database.addFilter(new AvoidAmbiguousClassesFilter());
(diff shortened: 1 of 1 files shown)


lukaseder:
Thank you very much for your suggestion. There are some interesting thoughts here.

I very much understand your use-case. However, I think that the approach of further opening up semi-internal API for extension via inheritance will not be the right way to go. We'll think about this...

jakubkulhan:
Ok, I see. I've managed to work around this issue by creating adapter for generator. Hopefully, there will be a clean way in future releases. Thank you for great tool&library! :)

lukaseder:
Great to hear that you've found a workaround. Will keep you posted on progress.

lukaseder:

I've created a feature request to keep track of this: https://github.com/jOOQ/jOOQ/issues/10286

Issue #2786, opened 2013-10-17
Correctly handle NULLS FIRST | NULLS LAST in SEEK method


lukaseder:
In #2657, the SEEK method was supported in jOOQ. The current implementation only works for `NOT NULL` columns. Support for `NULL` columns and `NULLS FIRST, NULLS LAST` is more complex and will be added later.

---

More details about the SEEK method can be seen here:
http://use-the-index-luke.com/sql/partial-results/fetch-next-page

See also:
https://groups.google.com/forum/#!topic/jooq-user/kABZWPQScSU

atollk:
Just came to this issue from a search engine. Considering that this has been dormant for 8 years, it's probably safe to assume that this won't be implemented in the foreseeable future? Would this issue be considered small enough to accept public PRs?

lukaseder:
Thanks for your message, @atollk. The issue being old doesn't have any meaning regarding its significance or status on the future roadmap. There are many flaws and open issues, and no strict reason why one flaw should require a more urgent fix than another.

Though the fact that there are 4 upvotes by now surely does change things. Let me reconsider this for jOOQ 3.17.

Regarding a PR, chances of it being correct or optimal are slim because you don't have access to integration tests (which aren't open source). If you're not licensed, then you can also not see the commercial code, which might require a different fix from that of the jOOQ Open Source Edition.

Though, if you're willing to spend time on analysing potential solutions, and make implementation suggestions, that might still be helpful (no promises). I can't say whether this is a "small enough" fix yet. It does seem to produce a few caveats, maybe even expose other flaws in the current `SEEK` implementation. Up to you whether you want to spend the time.

What would be helpful in any case is:

- A few test cases, scenarios, showing what kinds of queries you're trying to write
- A few workarounds showing how you implemented the query without `SEEK` (which is just syntax sugar, you can always work around `SEEK` limitations by writing the predicates manually)

atollk:
Sorry for the late answer; I had to get the clear for publishing this code. 

I tried to simplify it somewhat and remove our special cases, so this should be the general use case implementation:
```kotllin
import org.jooq.Condition
import org.jooq.Field
import org.jooq.Record
import org.jooq.SelectConditionStep
import org.jooq.SelectLimitStep
import org.jooq.SelectWhereStep
import org.jooq.SortField
import org.jooq.SortOrder
import org.jooq.impl.DSL
import org.jooq.impl.QOM

/**
 * A query ready to be paginated via a single attribute.
[...]

PR #5024, opened 2016-02-02
Wrong implementation of internal SortFieldList.nulls()

rvantonder:
The subexpressions between `||` are equal. Looks like this should be `getNullsLast()`.


lukaseder:
Thank you very much for the suggested fix and for signing the transfer of rights agreement.

You're right, this looks like a copy paste mistake in jOOQ's internals. Luckily, this code is not yet really used as it was implemented for #2786, which hasn't been implemented yet:
https://github.com/jOOQ/jOOQ/blob/5cb70bc7f473d1d09342201e3fd6288d50c2b164/jOOQ/src/main/java/org/jooq/impl/SelectQueryImpl.java#L1685

Nontheless, it's good to fix this "sleeping bug"

PR #5111, opened 2016-02-25
Added support for Vertabelo XSD version 2.3

nablaone:
Hi,

We've made changes to our XML schema. The new version (v2.3) will be deployed on production next week. I hope so. Version v2.2 will become legacy and will be still supported.

This patch contains support for both version. So it is backward compatible.


lukaseder:
Thank you very much for your contribution and update.

Unfortunately, recent events have made us review our integration and collaboration, and we've decided we will no longer be shipping the extension with our future distributions, or hosting it in our GitHub repository.

The reasons for this decision are:

1. We haven't seen much traction being generated from the integration
2. We have noticed that we're competitors for the same market: http://mobile-orm.vertabelo.com, and can thus no longer work for Vertabelo / e-point directly.

This integration, however, is Open Source and ASL 2.0 licensed, so you are free and very welcome to take it over and host it elsewhere. We'll leave the current integration on GitHub until we release jOOQ 3.8.

Thanks for your understanding.

PR #4939, opened 2016-01-15
SQLFeatureNotSupportedException thrown when initalizing MetaDataFieldProvider

terencez127:
Some JDBC drivers such as Teradata and Cassandra don't implement `ResultSetMetaData.getSchemaName` and/or `ResultSetMetaData.getTableName` methods, these methods are called in constructor of `MetadataFieldProvider` and would throw `SQLFeatureNotSupportedException`.

PR #5212, opened 2016-04-21
MetaDataFieldProvider should not catch SQLFeatureNotSupportedException from ResultSetMetaData calls

terencez127:
With Amazon Hive JDBC driver, `ResultSetMetaData.getSchemaName()` throws general `SQLException` rather than `SQLFeatureNotSupportedException`.

This is a followup of [pull request 4939](https://github.com/jOOQ/jOOQ/pull/4939).

### DIFF
--- a/jOOQ/src/main/java/org/jooq/impl/MetaDataFieldProvider.java
+++ b/jOOQ/src/main/java/org/jooq/impl/MetaDataFieldProvider.java
@@ -46,7 +46,6 @@ import static org.jooq.impl.DSL.name;
 import java.io.Serializable;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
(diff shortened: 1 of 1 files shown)


lukaseder:
Thank you very much for the suggested fix. Indeed, nothing in the JDBC spec explicitly refers to a `SQLFeatureNotSupportedException`:
https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSetMetaData.html#getTableName-int-

Only a generic `SQLException` is mentioned. Will merge immediately.

lukaseder:
Thanks again for this additional fix. I'll merge the fix to older releases as soon as possible.

terencez127:
Ah, didn't really look at the JDBC spec doc in the first place, good to know. Thank you.

Issue #4388, opened 2015-06-15
Compilation errors when applying <customType/> for PostgreSQL array types


lukaseder:
Reproduce like this:

```sql
CREATE TABLE t_exotic_types (
  ID                     INT  NOT NULL,
  UU_WRAPPER_ARRAY       UUID[],
  ...
```

And then

```xml
<customType>
    <name>org.jooq.test.all.bindings.UUIDWrapperArrayBinding</name>
    <type>org.jooq.test.all.types.UUIDWrapper[]</type>
    <binding>org.jooq.test.all.bindings.UUIDWrapperArrayBinding</binding>
</customType>

...

<forcedType>
    <name>org.jooq.test.all.bindings.UUIDWrapperArrayBinding</name>
    <expression>(?i:(.*?\.)T_EXOTIC_TYPES\.UU_WRAPPER_ARRAY)</expression>
</forcedType>
```

The generated output creates compilation errors:

```java
	/**
	 * The column <code>public.t_exotic_types.uu_wrapper_array</code>.
	 */
	public final TableField<TExoticTypesRecord, UUIDWrapper[][]> UU_WRAPPER_ARRAY = createField("uu_wrapper_array", org.jooq.impl.DefaultDataType.getDefaultDataType("org.jooq.test.all.types.UUIDWrapper[]").getArrayDataType(), this, "", new UUIDWrapperArrayBinding());
```

----
See also:
http://stackoverflow.com/q/30751083/521799

lukaseder:
The problem is related to the fact that PostgreSQL doesn't nicely report array types through the `information_schema` views. For instance, `column.data_type` simply contains `"ARRAY"`, and there's a `udt_name` column, which contains the base type with an underscore prefix.

A better way to fetch types and base types would be the following:

```sql
select a.attname, t.typname, t2.typname as basetype
from pg_catalog.pg_attribute a
join pg_catalog.pg_type t
on a.atttypid = t.oid
left join pg_catalog.pg_type t2
on t.typelem = t2.oid
```

The current design of jOOQ-meta doesn't allow for easily processing these array types. More bugs are probably out there.

lukaseder:
This change needs to introduce new internal (but protected, possibly extended by clients) API, which is why we cannot merge the fix to older releases

PR #5229, opened 2016-04-29
Allow forced type definitions for PostgreSQL arrays

lightoze:
This will be a very useful addition to #4388. Not the best design, but other options I have in mind would require too big changes (e.g. extending DataTypeDefinition with a method to return full database-specific type representation).

### DIFF
--- a/jOOQ-meta/src/main/java/org/jooq/util/AbstractDatabase.java
+++ b/jOOQ-meta/src/main/java/org/jooq/util/AbstractDatabase.java
@@ -1087,7 +1087,9 @@ public abstract class AbstractDatabase implements Database {
                      && (     definedType.getScale() != 0
                      ||   !p.matcher(definedType.getType() + "(" + definedType.getPrecision() + ")").matches())
                      && ( !p.matcher(definedType.getType() + "(" + definedType.getPrecision() + "," + definedType.getScale() + ")").matches() )
-                     && ( !p.matcher(definedType.getType() + "(" + definedType.getPrecision() + ", " + definedType.getScale() + ")").matches() )) {
+                     && ( !p.matcher(definedType.getType() + "(" + definedType.getPrecision() + ", " + definedType.getScale() + ")").matches() )
+                     && ( !isArrayType(definedType.getType())
+                     ||   !p.matcher(definedType.getType() + "[" + definedType.getUserType() + "]").matches() )) {
                     continue forcedTypeLoop;
                 }
             }


lukaseder:
Thank you very much for your suggestion. Is this fix still needed in jOOQ 3.8 (where #4388 is implemented)?

lightoze:
Yes, having it in 3.8 is very welcome.

lukaseder:
I meant is it still *needed* as in, does the problem persist now that #4388 is fixed in 3.8?

lukaseder:
... perhaps your issue is not the same as #4388... I'm trying to pinpoint the problem you're having here.

lightoze:
My issue is that I'd like to use #4388, but I'd like to configure binding for e.g. `uuid[]` type instead of specific table/column name.

lukaseder:
There must be a misunderstanding here, I'm sorry for the confusion. From your commit, I unfortunately cannot see exactly what you're trying to fix here. Given that #4388 is fixed 8and you referred to it), I'm not sure if your fix is really needed.

So, in order to move forward with this issue, could you please:

1. Show a case where something is broken with jOOQ 3.8, i.e. some steps to reproduce the issue you're having (in code: DDL statement, Your customer converter/binding, code generation configuration, expected behaviour / actual behaviour etc.)
2. Explain why your suggested fix would solve that problem

Thank you very much.

lightoze:
With #4388 fixed you can write:
```xml
<customType>
    <name>org.jooq.test.all.bindings.UUIDWrapperArrayBinding</name>
    <type>org.jooq.test.all.types.UUIDWrapper[]</type>
    <binding>org.jooq.test.all.bindings.UUIDWrapperArrayBinding</binding>
</customType>
```

But you cannot then use it (at least with PostgreSQL) like that:
```xml
<forcedType>
    <name>org.jooq.test.all.bindings.UUIDWrapperArrayBinding</name>
    <types>ARRAY\[UUID\]</types>
</forcedType>
```
Intention of this pull request is to make it possible to specify array type in `<types>` match. Is use case more clear now?

lukaseder:
Oh, thank you very much for clarifying! I get it now - yes that should be possible. But it doesn't work for me in the way you suggested as the `definedType.getUserType()` value reflects what's reported by PostgreSQL's `information_schema`. In the case of a `uuid[]` type, this is, unfortunately, `_uuid`.

Also, I'd like the regex to match the usual array syntax:

```xml
<types>UUID\[\]</types>
```

I'll think of a fix for this. Thanks for the suggestion!

lukaseder:
Hmm, I think we should re-organise 1-2 things in jOOQ-meta's internals in order to get a hold of these array types. The current implementation is not very satisfactory

lightoze:
Yes, I agree :)
This PR is just a quick and dirty fix.

mkurz:
@lukaseder also closed because of the `master` -> `main` rename?

lukaseder:
Yes, the PR got closed by github due to a recent rename of the main branch. I'm going through these right now.

lukaseder:
Anyway, I didn't agree with the implementation suggestion here. I've created an issue to keep track of this: https://github.com/jOOQ/jOOQ/issues/10285

Depending on user feedback, this issue might be prioritised in the future.

PR #5331, opened 2016-06-03
Accept column default expressions using sequences as identity columns in H2

oflege:
The H2TableDefinition.java currently only detects IDENTITY columns based on a non-null SEQUENCE_NAME in the column's INFORMATION_SCHEMA. 
In the database I have to deal with, there are composite primary keys defined like this:

```sql
CREATE TABLE foo (
  id        BIGINT       NOT NULL               DEFAULT nextval('uid_seq'),
  tenant_id BIGINT       NOT NULL,
...
  CONSTRAINT pk_user PRIMARY KEY (id, tenant_id))
```

for both Postgres and H2. In Postgres, inserts work as expected (due to the 'returning' clause for the keys), but nor for H2: After `insert into "PUBLIC"."FOO" ("TENANT_ID", ...) values (1, ...)`, no generated key for the id column is returned. In fact, the INFORMATION_SCHEMA does not contain a SEQUENCE_NAME
This patch resolves that, as it adds an alternative condition to detect an identity column (that is also used in PostgresTableDefinition).

### DIFF
--- a/jOOQ-meta/src/main/java/org/jooq/util/h2/H2TableDefinition.java
+++ b/jOOQ-meta/src/main/java/org/jooq/util/h2/H2TableDefinition.java
@@ -41,6 +41,7 @@
 package org.jooq.util.h2;
 
 import static org.jooq.util.h2.information_schema.tables.Columns.COLUMNS;
+import static org.jooq.tools.StringUtils.defaultString;
 
 import java.sql.SQLException;
 import java.util.ArrayList;
@@ -103,7 +104,8 @@ public class H2TableDefinition extends AbstractTableDefinition {
                 record.get(Columns.COLUMN_NAME),
                 record.get(Columns.ORDINAL_POSITION),
                 type,
-                null != record.get(Columns.SEQUENCE_NAME),
+                null != record.get(Columns.SEQUENCE_NAME)
+		    || defaultString(record.get(COLUMNS.COLUMN_DEFAULT)).startsWith("NEXTVAL"),
                 record.get(Columns.REMARKS));
 
             result.add(column);


lukaseder:
Thank you very much for your suggested pull request. Before being able to merge your suggestion into our dual-licensed product, may I ask you to read and sign our transfer of rights agreement? Details can be found here:

Nice find: The fact that the suggested code is also implemented in PostgreSQL. I have the following criticism that comes to mind (also for the current PostgreSQL implementation):

- It might be more sound to trim and lower-case the default expression prior to checking for `startsWith("nextval")`...
- I wonder if the presence of a `nextval` default expression is precise enough for a column to be considered an identity column. I wonder if it makes any sense for a table to default two columns to sequences...?

oflege:
- agree that toLowerCase() would make sense
- currently, h2 supports only one generated key column (http://www.h2database.com/javadoc/org/h2/jdbc/JdbcStatement.html#getGeneratedKeys), so most 'sensible' schemas would either use identity or a single nextval(...) expression for their primary key. But of course, there could be nextval expressions for other columns, too. So it might make sense to either throw a yet-unsupported-exception if multiple identity columns are detected or somehow limit nextval-identity-columns to those primary key columns

lukaseder:
Thanks for the feedback. Will merge and adapt accordingly:

- Add case insensitivity
- Log a warning on duplicate nextval expressions, ignoring the non-primary key one

lukaseder:
Thanks again for your contribution!

lukaseder:
Relabeling this as defect. Will merge to older releases as well

PR #5341, opened 2016-06-08
Correct signature of Record.map()

Ostralyan:
It should be possible to map using mappers of specific types of sub-record

Unfortunately Java does not support type covariance in this way by default

So probably <? extends Record> is the correct type of mapper

### DIFF
--- a/jOOQ/src/main/java/org/jooq/impl/AbstractRecord.java
+++ b/jOOQ/src/main/java/org/jooq/impl/AbstractRecord.java
@@ -761,7 +761,7 @@ abstract class AbstractRecord extends AbstractStore implements Record {
     }
 
     @Override
-    public final <E> E map(RecordMapper<Record, E> mapper) {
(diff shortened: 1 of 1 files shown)


lukaseder:

Hmm, if anything were to be fixed at all with that type signature, it would have to be `<? super Record>` due to the rules of PECS. But I doubt the current type hierarchy will support this. We've had this discussion several times previously on the mailing list. I'll look into this later with some more feedback.

lukaseder:
I'm closing this as won't fix. As it stands now, the suggested change is incorrect. Yes, the current signature of the `Record.map(RecordMapper)` method is unsatisfactory, but I don't see any means of improving it short of adding a recursive generic type on `Record<R extends Record<R>>` which would be asking for serious trouble ([details here](https://blog.jooq.org/2013/06/28/the-dangers-of-correlating-subtype-polymorphism-with-generic-polymorphism/))

PR #5343, opened 2016-06-09
Add createTableIfNotExists example

cluelessjoe:


lukaseder:
I've just noticed this - sorry, it has somehow escaped my attention in the past months. I agree this section of the manual could be improved. Rather than merging this PR, I prefer a thorough overhaul of the DDL section in the manual: #7553

PR #5350, opened 2016-06-17
treat column default expressions using RANDOM_UUID() as identity colu…

oflege:
…mns in H2

The fix in #5331 is incomplete as columns defaulting to RANDOM_UUID() should also be treated as identity columns

### DIFF
--- a/jOOQ-meta/src/main/java/org/jooq/util/h2/H2TableDefinition.java
+++ b/jOOQ-meta/src/main/java/org/jooq/util/h2/H2TableDefinition.java
@@ -109,7 +109,9 @@ public class H2TableDefinition extends AbstractTableDefinition {
                 null != record.get(Columns.SEQUENCE_NAME)
 
                 // [#5331] DEFAULT nextval('sequence') (PostgreSQL style)
-             || defaultString(record.get(Columns.COLUMN_DEFAULT)).trim().toLowerCase().startsWith("nextval"),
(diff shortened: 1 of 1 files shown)


lukaseder:
Hmm, thank you for your suggestion. I must say, that is very specific and I don't entirely agree. What if we wanted to assign random UUIDs to ... random columns? I'm not sure if this really qualifies as a strong indicator for an identity column. Perhaps, I'm getting more inclined towards supporting a "synthetic identity column" pattern, similar to the existing "synthetic primary key column" pattern, as can be seen here:

http://www.jooq.org/doc/latest/manual/code-generation/codegen-advanced

```xml
  <!-- A regular expression matching all columns that participate in "synthetic" primary keys,
       which should be placed on generated UpdatableRecords, to be used with

        - UpdatableRecord.store()
        - UpdatableRecord.update()
        - UpdatableRecord.delete()
        - UpdatableRecord.refresh()

       Synthetic primary keys will override existing primary keys. -->
  <syntheticPrimaryKeys>SCHEMA\.TABLE\.COLUMN(1|2)</syntheticPrimaryKeys>
```

That would resolve this problem more thoroughly. What do you think?

oflege:
I totaly agree, a syntheticIdentities re would be the most flexible way to solve the problem of having arbitrary identity columns

lukaseder:
Thanks for the feedback. I'll resolve this one as "won't fix", then, in favour of implementing #5360. Thanks for your continued feedback!

PR #5373, opened 2016-06-30
Add <syntheticIdentities> regular expression to code generator configuration

oflege:
…r configuration

Add feature for all databases/dialects that may create ColumnDefinitions with isIdentity = true

### DIFF
--- a/jOOQ-meta/src/main/java/org/jooq/util/AbstractDatabase.java
+++ b/jOOQ-meta/src/main/java/org/jooq/util/AbstractDatabase.java
@@ -52,6 +52,7 @@ import java.sql.Connection;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Collections;
+import static java.util.Collections.singletonList;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
@@ -120,6 +121,7 @@ public abstract class AbstractDatabase implements Database {
     private String[]                                                         recordVersionFields;
     private String[]                                                         recordTimestampFields;
     private String[]                                                         syntheticPrimaryKeys;
+    private String[]                                                         syntheticIdentities;
     private String[]                                                         overridePrimaryKeys;
     private boolean                                                          supportsUnsignedTypes;
     private boolean                                                          ignoreProcedureReturnValues;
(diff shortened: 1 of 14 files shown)


lukaseder:
Thank you very much for your suggested implementation. I will merge and comment on it afterwards

jOOQ-meta/src/main/java/org/jooq/util/AbstractDatabase.java
```diff
@@ -689,6 +691,21 @@ public void setSyntheticPrimaryKeys(String[] syntheticPrimaryKeys) {
...
+    }
+
+    @Override
+    public final boolean isSyntheticIdentity(ColumnDefinition columnDefinition) {
+        if (syntheticIdentities == null) {
+            return false;
+        }
+        return !filterExcludeInclude(singletonList(columnDefinition), null, syntheticIdentities, filters).isEmpty();
+    }
+
```
lukaseder: Hmm, I prefer to have this logic similar to the existing getters / setters, and not perform any filtering. I will adapt this and move the filter logic out of the getter...

jOOQ-meta/src/main/java/org/jooq/util/cubrid/CUBRIDTableDefinition.java
```diff
@@ -110,7 +110,7 @@ public CUBRIDTableDefinition(SchemaDefinition schema, String name, String commen
 			    null
 		    );
 
-			result.add(column);
+			result.add(applySyntheticIdentities(column));
 		}
```
lukaseder: This is easily forgotten, and I think it is thus applied at the wrong place. A column should always be produced with a synthetic identity already in place. Will try to find a better implementation

lukaseder:
Here's my suggestion:

- https://github.com/jOOQ/jOOQ/commit/2cc51949ea57c91e20a5329f972ae321e054f318
- https://github.com/jOOQ/jOOQ/commit/e91da53ccffe3b648c059a81d075d0406de927a1

It implements the logic in a single place, when initialising a `DefaultColumnDefinition`, whose `isIdentity()` flag needs to be adapted accordingly.

An integration test case succeeds with the current implementation.

Thanks again for your initial suggestion. This is a great feature to have in the code generator!

PR #5402, opened 2016-07-11
Space between type name and "unsigned"

subdir:
Seems like it may be missing unintentionally.

### DIFF
--- a/jOOQ-meta/src/main/java/org/jooq/util/mysql/MySQLTableDefinition.java
+++ b/jOOQ-meta/src/main/java/org/jooq/util/mysql/MySQLTableDefinition.java
@@ -96,7 +96,7 @@ public class MySQLTableDefinition extends AbstractTableDefinition {
 		    if (getDatabase().supportsUnsignedTypes()) {
     		    if (asList("tinyint", "smallint", "mediumint", "int", "bigint").contains(dataType.toLowerCase())) {
     	            if (record.get(Columns.COLUMN_TYPE).toLowerCase().contains("unsigned")) {
-    	                dataType += "unsigned";
+    	                dataType += " unsigned";
     	            }
     		    }
 		    }


lukaseder:
Thanks for the hint and suggestion.

This does look weird, although, it might be a workaround for a problem from the very early days of jOOQ. I'm going to be careful with merging this pull request... Did you run into an actual issue?

subdir:
> Did you run into an actual issue?

No, it just makes me think like I should write
```java
ForcedType.withName("bigint").withTypes("intunsigned");
```
instead of
```java
ForcedType.withName("bigint").withTypes("int unsigned");
```

And whatever I choose, it feels a little fragile.

lukaseder:
Thanks for the additional feedback. Yes, it is fragile. So are data type names as provided by the individual database vendors.

jOOQ follows a very lenient policy when it comes to matching data types. E.g. whitespace is ignored in order to make things like `number(1)` and `number ( 1 )` the same thing. But I'm sure there's potential for improving these things.

If you want to stay on the safe side, put a regular expression there, which matches different renderings of your type...

subdir:
Thanks for you answer.

> If you want to stay on the safe side, put a regular expression there, which matches different renderings of your type...

In some situations excessive regular expressions could be as dangerous as fixed type strings.

Not sure about other DBEs but the way you parse data types for MySQL is pretty strong and parsing results are deterministic so I would like to count on it. But may be it's a bad idea to count on implemetation details.

lukaseder:
> Not sure about other DBEs but the way you parse data types for MySQL is pretty strong and parsing results are deterministic so I would like to count on it.

Well, I wish the MySQL dictionary views were more reliable. They're not so easy to work with at times...

> But may be it's a bad idea to count on implemetation details.

Yes... It depends :)

lukaseder:
I'm rejecting this for now. Since there isn't any concrete issue, I'm afraid that accepting the suggested change might introduce a regression.

PR #5415, opened 2016-07-14
Recommend jOOQ Gradle plugin

oehme:
Instead of telling users to copy-paste a snippet of Groovy code into their build,
we now redirect them to a ready-to-use plugin. This provides several benefits:

- the user's build script is shorter
- the configuration syntax is cleaner
- it is easier to upgrade (users just update the plugin version instead of copy-paste-modifying a new snippet)
- the jOOQ team does not need to maintain documentation/examples
- the jOOQ team does not need to troubleshoot problems about generating with Gradle, users can instead open issues for the plugin

### DIFF
--- a/jOOQ-manual/src/main/resources/org/jooq/web/manual-3.8.xml
+++ b/jOOQ-manual/src/main/resources/org/jooq/web/manual-3.8.xml
@@ -15273,76 +15273,7 @@ public class Book {
                         <content><html>
                             <h3>Run generation with Gradle</h3>
                             <p>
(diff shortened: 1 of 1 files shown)


lukaseder:
;-)

Not what I expected, but OK. Let's give this a shot and see what happens. I suspect that with the suggestion at hand, I don't need to harass you with our transfer of rights agreement, this should be safe to merge as is

lukaseder:
- Updated all manuals on GitHub with the latest version
- Online manual will be updated by tomorrow

oehme:
I give you all rights to these two sentences ;)

lukaseder:
I will reward your generosity, should we ever meet in person.

PR #5421, opened 2016-07-18
Generate Scala-Style Getter/Setter for Record classes

er1c:
I think this is safe to just add-in as a default output format when using the scala code generator... I also added as a separate commit cleaning up the `name : Type` to `name: Type`

Reference for `_=`: http://docs.scala-lang.org/style/naming-conventions.html#accessorsmutators
Reference for `name: Type`: http://docs.scala-lang.org/style/types#annotations & http://docs.scala-lang.org/style/declarations#methods

### DIFF
--- a/jOOQ-codegen/src/main/java/org/jooq/util/JavaGenerator.java
+++ b/jOOQ-codegen/src/main/java/org/jooq/util/JavaGenerator.java
@@ -702,7 +702,7 @@ public class JavaGenerator extends AbstractGenerator {
         }
 
         if (scala)
-            out.tab(2).println("val %s : %s[%s, %s] = %s.createIdentity(%s, %s)",
+            out.tab(2).println("val %s: %s[%s, %s] = %s.createIdentity(%s, %s)",
                     getStrategy().getJavaIdentifier(identity),
                     Identity.class,
                     out.ref(getStrategy().getFullJavaClassName(identity.getTable(), Mode.RECORD)),
@@ -738,7 +738,7 @@ public class JavaGenerator extends AbstractGenerator {
         }
 
         if (scala)
-            out.tab(2).println("val %s : %s[%s] = %s.createUniqueKey(%s, \"%s\", [[%s]])",
+            out.tab(2).println("val %s: %s[%s] = %s.createUniqueKey(%s, \"%s\", [[%s]])",
                     getStrategy().getJavaIdentifier(uniqueKey),
                     UniqueKey.class,
                     out.ref(getStrategy().getFullJavaClassName(uniqueKey.getTable(), Mode.RECORD)),
@@ -774,7 +774,7 @@ public class JavaGenerator extends AbstractGenerator {
         }
 
         if (scala)
-        	out.tab(2).println("val %s : %s[%s, %s] = %s.createForeignKey(%s, %s, \"%s\", [[%s]])",
+        	out.tab(2).println("val %s: %s[%s, %s] = %s.createForeignKey(%s, %s, \"%s\", [[%s]])",
                     getStrategy().getJavaIdentifier(foreignKey),
                     ForeignKey.class,
                     out.ref(getStrategy().getFullJavaClassName(foreignKey.getKeyTable(), Mode.RECORD)),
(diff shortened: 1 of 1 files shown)


lukaseder:
Thank you very much for your suggestion. Interesting bit about the style. Just checked the [Scala style guide](http://docs.scala-lang.org/style/naming-conventions.html#values_variable_and_methods), looks like that's "best practice" or "least surprise". Interesting though. Why is it `name: Type` but `name = value`? Anyway I've registered a separate issue for this: https://github.com/jOOQ/jOOQ/issues/5422. This issue here will deal only with the main improvement.

May I ask you to also review and sign our transfer of rights agreement? We need this to be able to dual-license jOOQ including your contribution:

er1c:
I just signed the contribution agreement.  As far as the style goes, `val name: Type = value` is generally thought of as a strong convention for readability, even though the compiler's type inference would let you do `val name = value` and still be perfectly valid code

lukaseder:
Thanks a lot for signing, and thanks again for your contributions. Will merge immediately.

>is generally thought of as a strong convention for readability

Yes, I'm aware of that. I also prefer it, especially in generated code that users didn't write themselves. I was more wondering (bikeshedding) the whitespace itself. Before `:` there is none, but before `=` there is one... :)

lukaseder:
Hmm, I've done some additional introspection, and I have to reject both changes as they stand now.

1. The getter/setter logic is nice, but it generates additional getters / setters, in addition to jOOQ's pre-existing getters / setters. This should better be solved with naming strategies. I'm fine with specifying a different default strategy for Scala to produce the desired output, but there shouldn't be 2 setters and 2 getters in each record. I understand that you originally ran into this issue here: https://twitter.com/ericpeters/status/755088676959293441, which is why you implemented the pull request as is. But I prefer to fix the original issue here: https://github.com/jOOQ/jOOQ/issues/5424. Note, there is also a regression in the suggested implementation, when a setter is a reserved word, like `class`, which is currently quoted `` `class`_= ``.
2. The improved style causes regressions when names are disambiguated by jOOQ's code generator by appending an underscore, in case of which the colon becomes part of the method name, such as `def getClass_: Integer`

I'm reverting these changes for now.

er1c:
As far as 1. goes, that's fine, I was thinking about trying to change it somewhere here: https://github.com/jOOQ/jOOQ/blob/master/jOOQ-codegen/src/main/java/org/jooq/util/GenerationUtil.java#L184 - but the the hangup there was trying to figure out "am I in java mode or scala mode"
As far as 2. goes, that is really weird: https://gist.github.com/er1c/e227c204ef410ac5a97c383e4611ae39 I didn't realized a colan was possible there, that is definitely a fine point of lexical ambiguity I haven't run into before

lukaseder:
> that is definitely a fine point of lexical ambiguity I haven't run into before

We've seen it all in the jOOQ code generator :)

er1c:
@lukaseder Is there a regression test I missed?  I tried to run a mvn test on everything before creating the PR...

lukaseder:
You've missed tons :) But not your fault. The integration tests aren't open source...

er1c:
I might want to revisit this later...still thinking about it.
I might revisit the other dual-setter/duel-getter PR later...not super happy how this is coming together...

With a dual setter I can do:
```
      r.catalogId = catalogId
      r.username = username
      email.foreach{ r.setEmail }
      firstName.foreach{ r.setFirstName }
      lastName.foreach{ r.setLastName }
      company.foreach{ r.setCompany }
      address1.foreach{ r.setAddress1 }
      address2.foreach{ r.setAddress2 }
```

Converting it to strictly just be scala vs java:
```
      r.catalogId = catalogId
      r.username = username
      email.foreach{ r.email_= }
      firstName.foreach{ r.firstName_= }
      lastName.foreach{ r.lastName_= }
      company.foreach{ r.company_= }
      address1.foreach{ r.address1_= }
      address2.foreach{ r.address2_= }
```

It does the job, but wasn't as pretty as the dual-setter convenience methods.

lukaseder:
Hmm, I see. But I don't think a dual-setter/dual-getter is a reasonable default. You can, of course, implement a custom code section:

http://www.jooq.org/doc/latest/manual/code-generation/codegen-custom-code

PR #5429, opened 2016-07-19
Fix partitioning of window function

michael-simons:
The sum of the earnings was calculated wrong, each store in the result had the sum over both stores, which is not what the graph suggests to me. The graph shows the cumulative earnings, whereas the query must compute earnings per store.


michael-simons:
Ok, sorry to bother you. Too late, to hot… whatever. My bad :)

lukaseder:
OK, no worries :)

PR #5431, opened 2016-07-19
ScalaGenerator should follow Scala Style Guide in generated code

er1c:
Don't merge this yet, I'm still playing with it, but curious if your integration tests say anything...

### DIFF
--- a/jOOQ-codegen/src/main/java/org/jooq/util/GenerationUtil.java
+++ b/jOOQ-codegen/src/main/java/org/jooq/util/GenerationUtil.java
@@ -152,8 +152,91 @@ class GenerationUtil {
         "var",
         "while",
         "with",
-        "yield"
+        "yield"/*,
+        "_",
+        ":",
+        "=",
+        "=>",
+        "<-",
+        "<:",
+        "<%",
+        ">:",
+        "#",
+        "@"*/
     )));
+
+    private static Set<Character> SCALA_WHITESPACE = unmodifiableSet(new HashSet<Character>(asList(
+        (char)0x0020,
+        (char)0x0009,
+        (char)0x000D,
+        (char)0x000A
+    )));
+
+    private static Set<Character> SCALA_PARENTHESES = unmodifiableSet(new HashSet<Character>(asList(
+        '(',
+        ')',
+        '[',
+        ']',
+        '{',
+        '}'
+    )));
+
+    private static Set<Character> SCALA_DELIMITER = unmodifiableSet(new HashSet<Character>(asList(
+        '`',
+        '\'',
+        '"',
+        '.',
+        ';',
+        ','
+    )));
+
+    /**
+     * Take a character and determine if it's a valid "Scala Letter"
+     * http://www.scala-lang.org/files/archive/spec/2.11/01-lexical-syntax.html
+     *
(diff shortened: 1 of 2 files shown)


lukaseder:
Thanks. Will check this out by tomorrow, hopefully.

jOOQ-codegen/src/main/java/org/jooq/util/GenerationUtil.java
```diff
@@ -175,26 +258,38 @@ public static String convertToIdentifier(String literal, Language language) {
         StringBuilder sb = new StringBuilder();
 
         if ("".equals(literal))
-            return "_";
+            if (language == SCALA)
+                return "`_`";
```
lukaseder: Good point

lukaseder:
Yep, I agree with the status quo. I'll merge and run this through the integration tests, and will give you feedback soon.

lukaseder:
Hmm, that particular trailing `_` isn't added in `convertToIdentifier()` yet... It happens later in `GeneratorStrategyWrapper.disambiguateMethod()`. For instance, `getPrimaryKey` is a valid identifier per se, but it is not a valid method in the `Record` type hierarchy, as it would "accidentally" override a super type method.

I'll fix accordingly

lukaseder:
OK, never mind. There are still many edge cases, e.g. when passing such values to constructors, etc.

I really don't have time to fix all this madness given that we don't really gain anything useful at all :) I can live with the "bad style", so I'll just reject this issue.

lukaseder:
By the way: Thanks anyway for your contributions! It's very good to know about these caveats in the future

lukaseder:
By the way 2: I will recover the useful changes from this commit here https://github.com/jOOQ/jOOQ/pull/5431/commits/d0d0526cec336f3edfbea815455d0aea1025efc5. Will do that later during the day

er1c:
What was wrong with passing those values to a constructor?  Are those just bad tests? :)  I'd be curious what an example looked like, if nothing else for my own curiosity.

lukaseder:
If I remember correctly, there's an incident when an identifier is named like the top level package of the object it is contained in. Really wonky, I know, but this is a thing in Java and probably also in Scala. Java code below:

```java
package org.something.something;

class Something {
  SomeType org; // conflict with package here

}
```

Naming a member like a package in Java is asking for big trouble. This is why the member is automatically renamed to `org_`. Which again causes the same issues in many places.

jOOQ-codegen/src/main/java/org/jooq/util/GenerationUtil.java
```diff
@@ -175,26 +258,38 @@ public static String convertToIdentifier(String literal, Language language) {
...
 
-        return sb.toString();
+        String ret = sb.toString();
+
+        // In scala, _ should not be at the end of the identifier incase you have def foo_: Foo, so escape to def `foo_`: Foo
+        if (language == SCALA && ret.charAt(ret.length() - 1) == '_') {
+            return "`" + ret + "`";
+        } else {
+            return ret;
+        }
```
lukaseder: Agh. This also causes regressions. OK, I'm not going to evaluate why they happened :) I guess this whole escaping Scala identifier stuff is just a huge can of worms. Great to have learned this lesson. In the future, I'll always opt for less idiomatic style in favour of going out of the way of these Scala language caveats

PR #5432, opened 2016-07-20
Fix the type-erasure issue for scala vararg helpers in org.jooq.scala.Conversions using scala.predef.DummyImplicit

er1c:
See thread: http://stackoverflow.com/questions/34745066/dummyimplicits-is-this-used-and-how for how/why the DummyImplicit(s) are used to get around typeerasure issues

### DIFF
--- a/jOOQ-scala/src/main/scala/org/jooq/scala/Conversions.scala
+++ b/jOOQ-scala/src/main/scala/org/jooq/scala/Conversions.scala
@@ -134,16 +134,15 @@ object Conversions {
     def fetchOneOption[R <: Record]          (query : ResultQuery[R])                  : Option[R]      = Option(ctx.fetchOne(query))
     def fetchOneOption                       (rs : ResultSet)                          : Option[Record] = Option(ctx.fetchOne(rs))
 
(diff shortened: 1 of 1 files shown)


lukaseder:
Thank you very much for your suggestion. Not being too fluent with these Scala edge cases, would you mind providing some example client code that will be fixed by this suggestion?

er1c:
I was fixing your comments, either delete the comments or implement the patch, I'm a bit ambivalent until I run into needing them :)

lukaseder:
> I was fixing your comments, either delete the comments or implement the patch, I'm a bit ambivalent until I run into needing them :)

I'm sorry, I'm not sure if I understand...?

er1c:
@lukaseder See https://github.com/jOOQ/jOOQ/blob/master/jOOQ-scala/src/main/scala/org/jooq/scala/Conversions.scala#L137 - consider this PR a "fix" for those comments.   The "DummyImplicit" was (effectively) created exactly for this type-erasure problem.

lukaseder:
Oh, I see. Thank you very much! That's funky, though o_O

lukaseder:
Yeah, I should have taken more time to read through the page you originally linked. It seems obvious now

PR #5433, opened 2016-07-20
Fix Scala Conversions fetchOneOption consistent typing patterns

er1c:
This was probably just a copy/paste oversight, making the type parameters consistent with others


er1c:
/recompile - found a 2nd line/types to fix

PR #5469, opened 2016-07-25
Applied skeletal implementation to interface migration automated refactoring

khatchad:
This is a semantics-preserving automated refactoring that migrates existing method implementations in classes to the corresponding implemented interface methods as `default` methods. The tool **does not** add new code; it only rearranges *existing* code.

- We are evaluating a research prototype automated refactoring Eclipse plug-in called [Migrate Skeletal Implementation to Interface](https://github.com/khatchad/Migrate-Skeletal-Implementation-to-Interface-Refactoring). We have applied the tool to your project in the hopes of receiving feedback.
- The approach is very conservative. Specifically, it only performs a migration if the target interface is unambiguous. That may mean that not all changes that can be made were made. Please feel free to continue the refactoring manually if you wish.
- We only migrated methods declared in abstract classes. We believe such methods to be likely suitable default methods in corresponding interfaces.
- The source code should be semantically equivalent to the original.

Thank you for your help in this evaluation! Any feedback you can provide would be very helpful. In particular, we are interested if *each* of the proposed changes are helpful or not.

### DIFF
--- a/jOOQ-codegen/src/main/java/org/jooq/util/AbstractGenerator.java
+++ b/jOOQ-codegen/src/main/java/org/jooq/util/AbstractGenerator.java
@@ -472,18 +472,6 @@ abstract class AbstractGenerator implements Generator {
     }
 
     @Override
-    @Deprecated
-    public String fullyQualifiedTypes() {
-        return generateFullyQualifiedTypes();
-    }
-
-    @Override
-    @Deprecated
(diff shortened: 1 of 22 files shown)


lukaseder:
No thanks...

khatchad:
OK, thank you for the consideration.

Issue #5491, opened 2016-08-12
Provide maven-deploy script to install artifacts on remote repository


brunoborges:
It would be extremely useful if the release binary had a shell script to help deploy the artifacts into a remote Maven repository.

```
sh maven-deploy.sh <url, Required> [repositoryId, Optional]
```

Ideally, you should have one shell script only, with more parameters, such as:

```
Usage: maven-deploy.sh [options]
   -l: deploys artifacts into your local Maven repository
   -r: deploys artifacts into a remote Maven repository
   -u: the url of the remote Maven repository
   -i: the repository id in your settings.xml if remote repository requires authentication
```

lukaseder:
Thank you very much for your suggestion. Yes, that would be very helpful, indeed!

lukaseder:
This will be the usage:

```
Usage: maven-deploy.bat <options>

Options:
 -h, --help                      Display this help and exit
 -i, --repository   (Optional)   The repository id as specified in your settings.xml
 -u, --url          (Mandatory)  The repository URL
```

lukaseder:
This will be backported to 3.10.2 (#6810) and 3.9.7 (#6811)

PR #5492, opened 2016-08-12
Create maven-deploy.sh

brunoborges:
Script to deploy artifacts to a remove Maven repository. Fixes issue #5491

### DIFF
--- a/jOOQ-release/release/template/maven-deploy.sh
+++ b/jOOQ-release/release/template/maven-deploy.sh
@@ -1,0 +1,45 @@
+#!/bin/sh
+VERSION=3.9.0-SNAPSHOT
+
+# Lame but works like a charm... :-)
+# Maven will fail if these parameters are not properly provided anyway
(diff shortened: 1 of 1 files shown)


ooxi:
The binary artefacts of the open source version are available from Maven Central. For the commercial distribution I would nevertheless agree that having a (password protected) Maven Repository provided by Data Geekery would be beneficial.

brunoborges:
Please consider doing a more ellaborated shell script as described in #5491

lukaseder:
@ooxi: I agree that this would be beneficial to some users, but the setup and maintenance and support costs on our side would be quite high. At the same time, many organisations cannot connect to arbitrary Maven Repositories on the web for a variety of reasons, including firewall policies (whereas Maven Central is usually whitelisted). So for those organisations, it's usually better to proceed manually in the way @brunoborges has suggested:

1. Download jOOQ from the jOOQ website
2. Import / Deploy jOOQ to the local Maven Repository

Issue #3062, opened 2014-02-17
Add support for conversion into Collection types


lukaseder:
Currently, jOOQ's default data type conversion can convert from and to array types. It would be useful to be able to convert also to `Collection` types, although the collection element type information might not be available due to generic type erasure.

---

Related issue: #3060

kag0:
Just ran into situation today where this is needed. If you're open to PRs and can point me in the right direction, I can probably put in this feature.

lukaseder:
Thank you very much for your suggestion, @kag0! Yes, we're very open to PRs. Our only restriction is that we need you to sign this transfer of rights agreement in order for us to be able to dual license your contribution:
https://github.com/jOOQ/jOOQ/blob/master/CONTRIBUTING.md

The implementation should go in `org.jooq.tools.Convert`. (unless you meant a different conversion into collections?)

kag0:
Sounds good.   
I'll put it in `Convert`, is there anything in there for getting type of the elements for the target of the conversion?

lukaseder:
Hmm, I think we should definitely talk about what you're trying to do first. Can you show an example of API usage that you'd like to achieve after your change?

kag0:
So there are two levels of functionality.
#### simple case
The simple case is easy, and can but put in `Convert` with no problem. It just converts the array to a collection with no conversion of the elements.
```java
class myClass {
  int id;
  List<String> strings;
  ... getters, setters
}
```
```sql
CREATE TABLE my_table (
  id serial,
  strings varchar[]
)
```
```java
MyClass object = myTableRecord.into(MyClass.class);
```
#### advanced case
In the advanced case, it would be nice to also convert the elements, to say, enums. However, to do this we need the erased type of element stored in `enums`. 
```java
class myClass {
  int id;
  Set<MyEnum> enums;
  ... getters, setters
}
```
```sql
CREATE TABLE my_table (
  id serial,
  enums varchar[]
)
```
```java
MyClass object = myTableRecord.into(MyClass.class);
```

So in summary: with my change, I'd like to achieve the simple use case certainly. If generic type information is available somehow through jooq, then I'd like to handle the advanced case as well; but if not then I think addressing the advanced case is best left to a separate issue/feature.

lukaseder:
OK, perfect. I see the idea here. As a general rule of thumb, we can always live with some gray area as the `into(Class)` method is not expected to be completely type safe. It does a lot of mapping by convention, which may fail some times.

Having said so:

[...]

PR #5568, opened 2016-09-24
#3062 Work on adding ability to convert between collections and arrays.

kag0:
So this is not ready for merge, has some outstanding issues. 
- tests - I made `ConvertAll` package private so I could slap some unit tests in there. What is the correct place for tests?
- API - as you can see from the test, when converting from a collection you have to use `ConvertAll` directly, since adding another `convert(Collection<?> collection, Class<? extends T[]> type)` method would conflict, and changing the existing one might not be what you want to do.

### DIFF
--- a/jOOQ/src/main/java/org/jooq/tools/Convert.java
+++ b/jOOQ/src/main/java/org/jooq/tools/Convert.java
@@ -430,7 +430,7 @@ public final class Convert {
     /**
      * The converter to convert them all.
      */
-    private static class ConvertAll<U> implements Converter<Object, U> {
+    static class ConvertAll<U> implements Converter<Object, U> {
 
         /**
          * Generated UID
@@ -494,16 +494,31 @@ public final class Convert {
                     return convert(Arrays.toString((byte[]) from), toClass);
                 }
                 else if (fromClass.isArray()) {
+                    Object[] fromArray = (Object[]) from;
+
+                    if (Collection.class.isAssignableFrom(toClass) &&
+                            toClass.isAssignableFrom(ArrayList.class)) {
+                        return (U) new ArrayList(Arrays.asList(fromArray));
+                    }
+                    else if (Collection.class.isAssignableFrom(toClass) &&
+                            toClass.isAssignableFrom(LinkedHashSet.class)) {
+                        return (U) new LinkedHashSet(Arrays.asList(fromArray));
+                    }
 
                     // [#3443] Conversion from Object[] to JDBC Array
                     if (toClass == java.sql.Array.class) {
-                        return (U) new MockArray(null, (Object[]) from, fromClass);
+                        return (U) new MockArray(null, fromArray, fromClass);
                     }
                     else {
-                        return (U) convertArray((Object[]) from, toClass);
+                        return (U) convertArray(fromArray, toClass);
                     }
                 }
 
+                else if (toClass.isArray()
+                        && Collection.class.isAssignableFrom(fromClass)){
+                    Collection f = (Collection) from;
+                    return (U) f.toArray();
+                }
 
                 else if (toClass == Optional.class) {
                     return (U) Optional.of(from);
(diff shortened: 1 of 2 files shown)


jOOQ/src/main/java/org/jooq/tools/Convert.java
```diff
@@ -494,16 +494,31 @@ else if (fromClass == byte[].class) {
...
+                    Object[] fromArray = (Object[]) from;
+
+                    if (Collection.class.isAssignableFrom(toClass) &&
+                            toClass.isAssignableFrom(ArrayList.class)) {
+                        return (U) new ArrayList(Arrays.asList(fromArray));
+                    }
+                    else if (Collection.class.isAssignableFrom(toClass) &&
+                            toClass.isAssignableFrom(LinkedHashSet.class)) {
+                        return (U) new LinkedHashSet(Arrays.asList(fromArray));
+                    }
```
lukaseder: I like this implementation approach. Very cunning. Could you perhaps add a comment with an issue reference to the first `if`

```java
// [#3062] Default collections if no specific collection type was requested

File: jOOQ/src/main/java/org/jooq/tools/Convert.java [hunk 2]
```diff
@@ -430,7 +430,7 @@ private Convert() {}
     /**
      * The converter to convert them all.
      */
-    private static class ConvertAll<U> implements Converter<Object, U> {
+    static class ConvertAll<U> implements Converter<Object, U> {
```
lukaseder: I'd prefer not to open this up to package privacy. I understand you did this to access the class from tests, but this class shouldn't be accessible from tests. Tests should only call `Convert.xxx()` methods.

lukaseder:
> tests - I made ConvertAll package private so I could slap some unit tests in there. What is the correct place for tests?

Our tests are not open source. Any place you put them is fine, but do note, I'll remove them again from the public repository after merging.

> API - as you can see from the test, when converting from a collection you have to use ConvertAll directly, since adding another convert(Collection<?> collection, Class<? extends T[]> type) method would conflict, and changing the existing one might not be what you want to do.

There's `convertArray()` for special cases. I'm happy with adding `convertCollection()` in a similar fashion

kag0:
@lukaseder addressed those issues, let me know how it looks.

lukaseder:
I'm currently postponing this. I'm not 100% sure what the side-effects of such an added feature are for it to be included in 3.9. Will look again into this later.

Sorry for all the delays.

jOOQ/src/main/java/org/jooq/tools/Convert.java
```diff
@@ -517,7 +522,18 @@ else if (Collection.class.isAssignableFrom(toClass) &&
...
+                    Object[] list = f.stream()
+                            .map(e -> {
+                                if (!componentType.isAssignableFrom(e.getClass()))
+                                    return convert(e, componentType);
+                                return e;
+                            }).toArray();
+                    System.arraycopy(list, 0, dest, 0, dest.length);
+
+                    return (U) dest;
                 }
```
lukaseder: We can't generally make use of the `Stream` API in jOOQ as we still support Java 6. I will fix this code to retrofit it to our expectations.

lukaseder:
Thanks again for your pull request. I've reviewed it again now. There's another issue that I've commented on, but I'll fix that on my side. Will merge now and commit against this issue.

jOOQ/src/main/java/org/jooq/tools/Convert.java
lukaseder: I'm aware that Java 8 logic is scattered across this code (including references to `Optional`), but you don't see the full picture, unfortunately. The commercial distribution has special markers that delimit code with a Java 8 dependency and strips it from the Java 6 distribution. The Open Source edition doesn't have those markers.

lukaseder: Another issue (although unlikely), the new code block might conflict with the existing block handling `byte[]` target types. Will switch order of branch...

PR #5583, opened 2016-10-07
For generated code, change file header style from JavaDoc to regular comment

Stephan202:
Our codebase contains classes generated using `mvn jooq-codegen:generate`. When we compile with `-Werror -Xdoclint:reference` the following compilation failure is reported for each generated file:
```
/path/to/SomeGeneratedFile.java:2: warning: documentation comment not expected here
package some.package.name;
^
```

The reason is that the generated files have a `/** */`-style header, while (according to DocLint) that should be `/* */`. This PR changes the Java code generator to produce DocLint-compatibe code.

NB: This PR does not update the generated classes found under `jOOQ-meta` and `jOOQ-examples`.


Stephan202:
Note that the string "`This class is generated by jOOQ`" is also the default class JavaDoc text, as well as the `comment` attribute to the `@Generated` annotation (if enabled). So perhaps this file header can be dropped altogether. Or perhaps the word "`class`" should be replaced with "`file`".

Stephan202:
The build failure appears to be unrelated to this change; I see that the `master` branch build itself fails as well.

lukaseder:
Thank you very much for your suggestions.

- You're right. That file header should be a comment, not a javadoc
- Good point about jOOQ's internals. Will fix this separately: https://github.com/jOOQ/jOOQ/issues/5584
- I'm fine with your suggestion of changing the header from "this class" to "this file". We'll leave the header right now, as the `@Generated` annotation can be disabled. I'd like to keep some notion of mentioning that the file is generated, by default.
- Indeed, the build currently fails in the open source edition (since a couple of days). This is in our backlog to fix.

In order to be able to merge your PR, I will need you to read our contribution guidelines and sign our transfer of rights agreement, which allows us to dual-license your contribution, if that is OK with you:

Thanks again very much for your suggestions

Stephan202:
I will read and sign the CLA in a minute.

> I'm fine with your suggestion of changing the header from "this class" to "this file". 

So it's okay to do that in the context of this same PR? (Asking because some projects are stricter than others when it comes to the "one change per ticket/PR" policy.)

lukaseder:
> So it's okay to do that in the context of this same PR? (Asking because some projects are stricter than others when it comes to the "one change per ticket/PR" policy.)

Ideally, one ticket/PR would be great, of course, as this would benefit the release notes. But as long as it's one commit per change, I'm fine with it.

Thanks for asking!

Stephan202:
No shortage of PR numbers, so I'll file a separate PR for that in a bit, then ;).

(I agree that separate PRs are a good practice: keeps the discussion localized.)

Stephan202:
> I just realised, your contribution is just one line. We don't require a transfer of rights signature for these small things

Noticed that update only after I signed the CLA ;). Maybe something more substantial will come up in the future.

lukaseder:
> Maybe something more substantial will come up in the future.

Excellent. Looking forward! :)

PR #5585, opened 2016-10-07
Tweak generated Java file header

Stephan202:
As discussed in #5583. For context, comment explaining the rationale:
> Note that the string "`This class is generated by jOOQ`" is also the default class JavaDoc text, as well as the `comment` attribute to the `@Generated` annotation (if enabled). So perhaps this file header can be dropped altogether. Or perhaps the word "`class`" should be replaced with "`file`".


lukaseder:
Thanks a lot!

Issue #5333, opened 2016-06-06
DefaultBinding#pgNewArray does not handle arrays of records


BrentDouglas:
jOOQ 3.7.2 professional
PG 9.4.5

Assuming we have a schema in PG like this:

``` sql
create type foo as (
  id int,
  name text,
  otherThing text
);

create type bar as (
  id int,
  foos foo[]
);
```

Selecting some data as a `bar` might result in a record looking like:
`(1,"{""(2,\\""First\\"",)"",""(3,\\""Second\\"",\\""This one is set\\"")""}")`

The column `foos` will not be mapped correctly by `DefaultBinding` as the parse logic does not consider 
embedded records and tokenises the array values using `,` which will split the first record after `2` (in the above example).

A workaround (and probably more robust general solution) is to call `row_to_json` and the unmap it afterwards but it would be nice to avoid the encoding overhead.

lukaseder:
Thank you very much for reporting this.

May I know how exactly you fetch the `bar` type? Is it from a table or a stored procedure? I.e. what's the jOOQ API usage you're running?

BrentDouglas:
Sorry, I should have included this in the original report. The sql I have looks roughly like this (using the type defs above):

```sql
create table t_foo (
  id serial primary key,
  name text not null,
  other text
);
create table t_bar (
  id serial primary key,
  foo_ids int[]
);

create or replace view v_bar as
  select
      b.id,
      row(
          b.id,
          coalesce(f.data, '{}')
      )::bar as data
  from t_bar b
  left join lateral (
      select
          array_agg(
              row(
                  f.id,
                  f.name,
                  f.other
              )::foo
          )::foo[] as data
      from t_foo f
      where f.id = any(b.foo_ids)
  ) f on true;
```

And the jOOQ code is:
```java
    public List<BarRecord> findBars() {
        return db.select(V_BAR.DATA)
                .from(T_BAR)
                .leftJoin(V_BAR)
                .on(V_BAR.ID.eq(T_BAR.ID))
                .fetchInto(BarRecord.class);
    }
```

In the above code `BarRecord` is a Pojo generated by the code-gen plugin.

I created a POC here (https://github.com/BrentDouglas/jooq-5333)

lukaseder:
Oh, very interesting query! Thanks a lot for the details. I will look into this soon

BrentDouglas:
[...]

PR #5586, opened 2016-10-10
[#5333] Fix PG record parsing

BrentDouglas:
This commit extends the PG tokenization of records to handle the cases
of nested objects and arrays. It also adds support for java.time classes
in UDTs.

### DIFF
--- a/jOOQ/src/main/java/org/jooq/DataType.java
+++ b/jOOQ/src/main/java/org/jooq/DataType.java
@@ -100,6 +100,11 @@ public interface DataType<T> extends Serializable {
      */
     DataType<T[]> getArrayDataType();
 
+    /**
+     * Retrieve the data type for single value of this data type.
+     */
+    DataType<?> getElementType();
+
 
 
 
--- a/jOOQ/src/main/java/org/jooq/impl/ArrayDataType.java
+++ b/jOOQ/src/main/java/org/jooq/impl/ArrayDataType.java
@@ -75,6 +75,11 @@ final class ArrayDataType<T> extends DefaultDataType<T[]> {
         return getArrayType(configuration, castTypeName);
     }
 
+    @Override
+    public final DataType<?> getElementType() {
+        return this.elementType;
+    }
+
     private static String getArrayType(Configuration configuration, String dataType) {
         switch (configuration.family()) {
             case HSQLDB:
--- a/jOOQ/src/main/java/org/jooq/impl/DefaultBinding.java
+++ b/jOOQ/src/main/java/org/jooq/impl/DefaultBinding.java
@@ -1355,6 +1355,12 @@ public class DefaultBinding<T, U> implements Binding<T, U> {
     @SuppressWarnings("unchecked")
     @Override
     public void get(BindingGetResultSetContext<U> ctx) throws SQLException {
+        _get(ctx, type, converter);
+    }
+
+
+    @SuppressWarnings("unchecked")
+    private static <T,U> void _get(BindingGetResultSetContext<U> ctx, final Class<T> type, final Converter<T, U> converter) throws SQLException {
         T result = null;
 
         if (type == Blob.class) {
@@ -1546,19 +1552,19 @@ public class DefaultBinding<T, U> implements Binding<T, U> {
     }
 
 
-    private final LocalDate localDate(Date date) {
+    private static LocalDate localDate(Date date) {
         return date == null ? null : date.toLocalDate();
     }
 
-    private final LocalTime localTime(Time time) {
+    private static LocalTime localTime(Time time) {
         return time == null ? null : time.toLocalTime();
     }
 
-    private final LocalDateTime localDateTime(Timestamp timestamp) {
+    private static LocalDateTime localDateTime(Timestamp timestamp) {
         return timestamp == null ? null : timestamp.toLocalDateTime();
     }
 
-    private final OffsetTime offsetTime(String string) {
+    private static OffsetTime offsetTime(String string) {
         if (string == null)
             return null;
 
@@ -1569,7 +1575,7 @@ public class DefaultBinding<T, U> implements Binding<T, U> {
         return OffsetTime.parse(string);
     }
 
-    private final OffsetDateTime offsetDateTime(String string) {
+    private static OffsetDateTime offsetDateTime(String string) {
         if (string == null)
             return null;
 
@@ -2096,13 +2102,20 @@ public class DefaultBinding<T, U> implements Binding<T, U> {
     // interfaces. Instead, a string representation of a UDT has to be parsed
     // -------------------------------------------------------------------------
 
-    private static final <T> T pgFromString(Class<T> type, String string) {
-        return pgFromString(Converters.identity(type), string);
-    }
+    // PGobject parsing state machine
+    private static final int    PG_OBJECT_BEFORE_VALUE       = 1;
+    private static final int    PG_OBJECT_QUOTED_VALUE       = 2;
+    private static final int    PG_OBJECT_UNQUOTED_VALUE     = 3;
+    private static final int    PG_OBJECT_AFTER_VALUE        = 4;
+
+    // quotation stack markers
+    private static final int UNQUOTED = 0;
+    private static final int QUOTED = 1;
 
     @SuppressWarnings("unchecked")
-    private static final <T> T pgFromString(Converter<?, T> converter, String string) {
-        Class<T> type = converter.toType();
+    private static <T> T pgFromString(final Field<?> field, String string, ParseContext ctx) {
+        final Converter converter = field.getConverter();
+        Class<?> type = converter.toType();
 
         if (string == null) {
             return null;
@@ -2155,6 +2168,21 @@ public class DefaultBinding<T, U> implements Binding<T, U> {
         else if (type == Timestamp.class) {
             return (T) Timestamp.valueOf(string);
         }
+        else if (type == LocalTime.class) {
+            return (T) LocalTime.parse(string);
+        }
+        else if (type == LocalDate.class) {
+            return (T) LocalDate.parse(string);
+        }
+        else if (type == LocalDateTime.class) {
+            return (T) LocalDateTime.parse(string);
+        }
+        else if (type == OffsetTime.class) {
+            return (T) offsetTime(string);
+        }
+        else if (type == OffsetDateTime.class) {
+            return (T) offsetDateTime(string);
+        }
         else if (type == UByte.class) {
             return (T) UByte.valueOf(string);
         }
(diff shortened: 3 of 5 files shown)


BrentDouglas:
I would have liked to add some tests for this but I couldn't really see where they are meant to go. It works against a reasonably complicated UDT I have in a project though, e.g. this response maps correctly into the UDT for both splitting and quoting:

`(1,1gfds,"1hgfd",,"(1,Zxcv)","(20,""Fdsa & gfs"")","(8,""Bcx & gfds"")",,"{""(2,Bar,\\""{\\""\\""(3,Baz,)\\""\\""}\\"")"",""(1,Foo,\\""{\\""\\""(1,Bar,)\\""\\"",\\""\\""(2,Baz,)\\""\\"",\\""\\""(4,Aasf,)\\""\\""}\\"")"",""(3,Asdf,\\""{\\""\\""(5,Fds,)\\""\\""}\\"")""}",)`

lukaseder:
Thank you very much for your suggestion. That's greatly appreciated!

> I would have liked to add some tests for this but I couldn't really see where they are meant to go

Don't worry about it much. We haven't open sourced our integration tests for a variety of reasons, so we don't require tests in third-party contributions. But some example code to help reproduce the issue would be great, nonetheless. Especially, since you seem to have fixed two issues in one go.

What would be a `CREATE TYPE` definition for your UDT serialisation? And how do you fetch it? From a table?

BrentDouglas:
Hi Lucas, 

I extended the original test case I had that should display these issues. The quoting thing was something that I don't think you really going to be able to see without debugging through the patched version. The original version of my patch didn't deal with it and it worked until I put some data in that got returned at the third quote level (`\\""`). I think the current code will fail earlier due to issues with splitting the nested object on their commas.

https://github.com/BrentDouglas/jooq-5333

There are a couple of things in the patch that I'm not sure are necessary any more but I wanted to confirm with you before removing them. The object parser looks for `null` when parsing them however from what I have seen PG sends an empty value (e.g. `(1,,3)`) for null. I've only tested this on 9.5 though. The other was the handling of `\\` in the original code. I've not seen that either and wasn't sure if that was the way that the old code path handled the third quote level or if that value is allowed elsewhere in the response.

BrentDouglas:
The response in the test case now looks like this:

```
(1,"{""(1,First,)"",""(2,\\""Quoted & , asdf\\"",)""}","{""(1,\\""{\\""\\""(1,First,)\\""\\"",\\""\\""(2,\\\\\\\\\\""\\""Quoted & , asdf\\\\\\\\\\""\\"",)\\""\\""}\\"")"",""(2,\\""{\\""\\""(1,First,)\\""\\"",\\""\\""(2,\\\\\\\\\\""\\""Quoted & , asdf\\\\\\\\\\""\\"",)\\""\\""}\\"")""}")
```

So I guess that answers my question about handling `\\`

lukaseder:
This issue looks similar: https://github.com/jOOQ/jOOQ/issues/5613. Will check if that's the same problem

robertotru:
@BrentDouglas did you test the changes in the pull request with a `json` field inside a UDT that is inside an array in another UDT?
It seems that strings in json fields are now read in a value as a `String` where all the `"` are removed (maybe is an unescaping issue). Thus, when the custom JSON parser is called by jOOQ it fails due to a wrong JSON syntax. 

For instance, in a PostgresSQL 9.5 database I have a `json` field with a row containing the value `{"command":1,"message":"this is a test message"}` but in  `DefaultBinding.pgFromString` it is now passed as the Java String with value `{command:1,message:this is a test message}`.

For completeness, I also report the stacktrace

```
com.google.gson.JsonSyntaxException: com.google.gson.stream.MalformedJsonException: Unterminated object at line 1 column 26 path $.message
	at com.google.gson.Gson.fromJson(Gson.java:902)
	at com.google.gson.Gson.fromJson(Gson.java:852)
	at com.google.gson.Gson.fromJson(Gson.java:801)
	at com.google.gson.Gson.fromJson(Gson.java:773)
	at com.mytest.dbms.jooq.databinding.PostgresJSONGsonBinding$1.from(PostgresJSONGsonBinding.java:32)
	at com.mytest.dbms.jooq.databinding.PostgresJSONGsonBinding$1.from(PostgresJSONGsonBinding.java:28)
	at org.jooq.impl.DefaultBinding.pgFromString(DefaultBinding.java:2203)
	at org.jooq.impl.DefaultBinding.pgSetValue(DefaultBinding.java:2336)
	at org.jooq.impl.DefaultBinding$2.operate(DefaultBinding.java:2235)
	at org.jooq.impl.RecordDelegate.operate(RecordDelegate.java:128)
	at org.jooq.impl.DefaultBinding.pgNewRecord(DefaultBinding.java:2227)
	at org.jooq.impl.DefaultBinding.pgFromString(DefaultBinding.java:2196)
	at org.jooq.impl.DefaultBinding.pgSetValue(DefaultBinding.java:2336)
	at org.jooq.impl.DefaultBinding$2.operate(DefaultBinding.java:2235)
	at org.jooq.impl.RecordDelegate.operate(RecordDelegate.java:128)
	at org.jooq.impl.DefaultBinding.pgNewRecord(DefaultBinding.java:2227)
	at org.jooq.impl.DefaultBinding.get(DefaultBinding.java:1521)
	at org.jooq.impl.DefaultBinding.pgGetArray(DefaultBinding.java:2274)
	at org.jooq.impl.DefaultBinding.get(DefaultBinding.java:1499)
	at org.jooq.impl.CursorImpl$CursorIterator$CursorRecordInitialiser.setValue(CursorImpl.java:1653)
	at org.jooq.impl.CursorImpl$CursorIterator$CursorRecordInitialiser.operate(CursorImpl.java:1624)
	at org.jooq.impl.CursorImpl$CursorIterator$CursorRecordInitialiser.operate(CursorImpl.java:1589)
	at org.jooq.impl.RecordDelegate.operate(RecordDelegate.java:128)
	at org.jooq.impl.CursorImpl$CursorIterator.fetchOne(CursorImpl.java:1553)
	at org.jooq.impl.CursorImpl$CursorIterator.hasNext(CursorImpl.java:1520)
	at org.jooq.impl.CursorImpl.fetch(CursorImpl.java:342)
	at org.jooq.impl.CursorImpl.fetchOne(CursorImpl.java:300)
	at org.jooq.impl.Tools.fetchOne(Tools.java:1481)
	at org.jooq.impl.AbstractResultQuery.fetchOne(AbstractResultQuery.java:517)
	at org.jooq.impl.AbstractResultQuery.fetchOneInto(AbstractResultQuery.java:540)
	at org.jooq.impl.AbstractResultQuery.fetchOptionalInto(AbstractResultQuery.java:633)
	at org.jooq.impl.SelectImpl.fetchOptionalInto(SelectImpl.java:2982)
	at com.mytest.MyJooqRepository.udtSelect(MyJooqRepository.java:29)
Caused by: com.google.gson.stream.MalformedJsonException: Unterminated object at line 1 column 26 path $.message
	at com.google.gson.stream.JsonReader.syntaxError(JsonReader.java:1559)
	at com.google.gson.stream.JsonReader.doPeek(JsonReader.java:491)
	at com.google.gson.stream.JsonReader.hasNext(JsonReader.java:414)
	at com.google.gson.internal.bind.TypeAdapters$29.read(TypeAdapters.java:738)
	at com.google.gson.internal.bind.TypeAdapters$29.read(TypeAdapters.java:714)
	at com.google.gson.internal.bind.TypeAdapters$35$1.read(TypeAdapters.java:910)
	at com.google.gson.Gson.fromJson(Gson.java:887)
	... 170 common frames omitted
```

BrentDouglas:
No I didn't. I'll check it out though, thanks.

Brent

lukaseder:
I have fixed #5613 and I will reject this PR as a duplicate. The PR substantially changes the parsing mechanism, which introduces quite a bit of risk of regressions. Perhaps, I'm missing 1-2 ideas from this PR, but as far as this particular fix is concerned, I'm not sure if all the changes are really needed. E.g. the improved JSR-310 support should not be piggy-backed to this fix.

lukaseder:
In any case: Thanks a lot for the detailed issue report and for trying to provide a PR, @BrentDouglas. This is greatly appreciated!

BrentDouglas:
Just FYI, the commit for #5613 does not fix the issues I am seeing. After looking at the issue posted by @robertotru I can see that this MR is flawed though I have not had any time since to look into it.

lukaseder:
Thanks for following up, @BrentDouglas. Well, I can take over of course. If you can provide an isolated test case (like shown here: https://github.com/jOOQ/jOOQ/issues/5613#issuecomment-255401791), it should be easy to continue for me...

BrentDouglas:
Essentially the core of the issue is that quoting is not handled correctly. I looked into it a bit more today are afaict this is how PG quotes things:

- The characters `"` and `\` require quoting.
- Arrays quote things by adding a `\` character to every `"` and `\` within it's content, i.e. `"` and `\` become `\"` and `\\` respectively
- Objects quote things by doubling every quotable character within it's content, i.e. `"` and `\` become `""` and `\\` respectively
- Arrays quote nested objects & objects quote nested arrays if there is quotable content anywhere inside
- This happens every time they are nested.

Some examples:

```
test=# select array[row()];
 array
-------
 {()}
(1 row)

test=# select array[row('')];
   array
------------
 {"(\"\")"}
(1 row)

test=# select array[row('\"')];
       array
--------------------
 {"(\"\\\\\"\"\")"}
(1 row)

test=# select row(array[9]);
  row
-------
 ({9})
(1 row)

test=# select row(array['']);
    row
------------
 ("{""""}")
(1 row)

test=# select row(array['\"']);
        row
--------------------
 ("{""\\\\\\""""}")
(1 row)

test=# select array[array[row('\"')]];
        array
----------------------
 {{"(\"\\\\\"\"\")"}}
(1 row)

test=# select array[row(array[row('\"')])];
                                 array
------------------------------------------------------------------------
 {"(\"{\"\"(\\\\\"\"\\\\\\\\\\\\\\\\\\\\\"\"\\\\\"\"\\\\\"\")\"\"}\")"}
(1 row)
```

This MR code tried to handle it by stripping one layer of quoting when it reads the values, however it has a couple of issues in that 1) I hadn't realised that the above is how PG quoted things so the code isn't written for that and 2) It doesn't deal with json, which gets mixed up in the array handling code and totally destroys it's state handling for quotes.

What I was looking into today (I'll push the changes but they aren't even close to working) was parsing recursively (e.g. so pgNewObject calls pgNewArray when it hits one) so that we can count the stack frames and knock off `(2 ^ depth) - 1` quote characters.

The test code I am using has this schema:

```sql
create type foo as (
  id int,
  name text,
  json json
);

create type bar as (
  id int,
  foos foo[]
);

create type baz as (
  id int,
  foos foo[],
  bars bar[]
);

create table t_foo (
  id serial primary key,
  name text not null,
  json json
);
create table t_bar (
  id serial primary key,
  foo_ids int[]
);
create table t_baz (
  id serial primary key,
  foo_ids int[],
  foo_jsons json[],
  bar_ids int[]
);

create or replace view v_bar as
  select
      b.id,
      row(
          b.id,
          coalesce(f.data, '{}')
      )::bar as data
  from t_bar b
  left join lateral (
      select
          array_agg(
              row(
                  f.id,
                  f.name,
                  f.json
              )::foo
          )::foo[] as data
      from t_foo f
      where f.id = any(b.foo_ids)
  ) f on true;

create or replace view v_baz as
  select
      b.id,
      row(
          b.id,
          coalesce(f.data, '{}'),
          coalesce(f.json, '{}'),
          coalesce(r.data, '{}')
      )::baz as data
  from t_baz b
  left join lateral (
      select
          f.json,
          array_agg(
              row(
                  f.id,
                  f.name,
                  f.json
              )::foo
          )::foo[] as data
      from t_foo f
      where f.id = any(b.foo_ids)
  ) f on true
  left join lateral (
      select
          array_agg(
                  v.data
          )::bar[] as data
      from t_bar f
      left join v_bar v on f.id=v.id
      where f.id = any(b.bar_ids)
  ) r on true;
```

and I'm calling it with this:

```java

    @Test
    public void testJOOQ5333() throws Exception {
        try {
            db.batch(
                    db.insertInto(T_FOO)
                            .set(T_FOO.ID, 1)
                            .set(T_FOO.NAME, "\"First"),
                    db.insertInto(T_FOO)
                            .set(T_FOO.ID, 2)
                            .set(T_FOO.NAME, "\"Quoted & , asdf")
                            .set(T_FOO.JSON, "{\"foo\":9,\"bar\":[\"a\",\"b\"],\"c\":{\"a\":[{\"b\":5},{\"b\":5}]}}"),
                    db.insertInto(T_FOO)
                            .set(T_FOO.ID, 3)
                            .set(T_FOO.NAME, "adf")
                            .set(T_FOO.JSON,
                                    "[{\"foo\":9,\"bar\":[\"a\",\"b\"],\"c\":{\"a\":[{\"b\":5},{\"b\":5}]}},{\"foo\":9,\"bar\":[\"a\",\"b\"],\"c\":{\"a\":[{\"b\":5},{\"b\":5}]}}]"),
                    db.insertInto(T_FOO)
                            .set(T_FOO.ID, 4)
                            .set(T_FOO.NAME, "asdgf"),
                    db.insertInto(T_BAR)
                            .set(T_BAR.ID, 1)
                            .set(T_BAR.FOO_IDS, new Integer[]{1, 2}),
                    db.insertInto(T_BAR)
                            .set(T_BAR.ID, 2)
                            .set(T_BAR.FOO_IDS, new Integer[]{3, 4}),
                    db.insertInto(T_BAZ)
                            .set(T_BAZ.ID, 1)
                            .set(T_BAZ.FOO_IDS, new Integer[]{1, 2, 3, 4})
                            .set(T_BAZ.BAR_IDS, new Integer[]{1, 2})
            ).execute();

            final List<BazRecord> ret = db
                    .select(V_BAZ.DATA)
                    .from(T_BAZ)
                    .leftJoin(V_BAZ)
                    .on(V_BAZ.ID.eq(T_BAZ.ID))
                    .fetchInto(BazRecord.class);

            assertEquals(1, ret.size());
        } catch (final DataAccessException e) {
            if (e.getCause() instanceof SQLException) {
                e.addSuppressed(((SQLException) e.getCause()).getNextException());
            }
            throw e;
        }
    }
```

I haven't got around to it yet but I am going to add some `bytea` stuff in there as they start with `\x` and which looks like it's probably easy to break when unquoting things.

If that still doesn't explain the issues let me know and I'll try and explain more clearly.

BrentDouglas:
Ah, ignore those changes I pushed for the moment, they are very wip anyway while I was trying to work out the difference between PG's actual quoting and the parser's expected quoting. I must have hit reformat with my work config.

The only relevant parts are the changes at the bottom of `DefaultBindings`, the idea of which is to change the parser to work recursively rather than returning lists so that we never have to write the quotes to the return value, just the content.

lukaseder:
Thank you very much for all your work on this, @BrentDouglas. I'm sorry I wasn't faster in reviewing and merging these things - the recent months have been quite busy with trainings and preparations for jOOQ 3.9. I'm hoping that this can be merged after the release, soon

BrentDouglas:
@lukaseder I think this can be closed. My original intention with this was to change away from encoding types as json  on the server to returning the raw results, ideally lowering the overhead of encoding json away from the db server. After seeing the crazy quoting scheme used I no longer have any desire to do this. I don't really have time to work on this at the moment so I will close it. I think you can close the associated ticket as well. I think if anyone runs into this again its probably easiest to recommend calling row_to_json and then deserializing it in a RecordMapper. Thanks for your patience with this.

lukaseder:
OK, thanks for the feedback :)

Issue #5639, opened 2016-11-04
MockFileDatabase ignores erroneous rows specification


lukaseder:
It's possible to do things like this with the `MockFileDatabase`:

```
select "TABLE2"."ID2", "TABLE2"."NAME2" from "TABLE2"
> +---+-----+
> |ID2|NAME2|
> +---+-----+
> |1  |X    |
> |2  |Y    |
> +---+-----+
@ rows: 1000
```

This should probably throw an exception

PR #5676, opened 2016-11-19
[#5639] Fix MockFileDatabase ignores erroneous rows specification

samy-deghou:
Erroneous row specification in file-based MockDataProvider  does not pass anymore and an Exception is thrown instead


lukaseder:
Thank you very much for your suggested fix. I will look into your suggestions ASAP. In the meantime, may I ask you to review our contribution guidelines here:

... and sign our transfer of rights agreement, which is required for us to be able to dual-license your contribution to be included also in the commercial distribution?

lukaseder:
Thank you for signing and thanks again for your suggested contribution. I'll implement a different version based on your suggestion.

Here's my feedback on your suggestion:

- There are a couple of intermediate commits, which make it a bit hard to see what has really changed. E.g. at some point, you've added exception declarations to the throws clause and then removed them again. I'd prefer if there was only a single commit with only one change
- I think this exception is not important enough to be placed in its own class. I'll replace it by a more generic `MockFileDatabaseException` that can be recycled for other problems that arise from ill-configured `MockFileDatabase` files.

I'm closing this issue and commit the adapted version directly on #5639. Thanks again for your help!

PR #5687, opened 2016-11-30
Fix for mock of delete

codeturner:
Mocks for delete always returns 0.  Instead, it should get the return value from MockStatement.getUpdateCount(). To accomplish this, MockStatement.execute0 should return false for non-queries, such as delete, to indicate that a result set is not available, and should be obtained via call to getUpdateCount(). Also, changes need to be made to MockStatement.getUpdateCount() to return value from data as an integer for non-queries.

### DIFF
--- a/jOOQ/src/main/java/org/jooq/tools/jdbc/MockStatement.java
+++ b/jOOQ/src/main/java/org/jooq/tools/jdbc/MockStatement.java
@@ -188,7 +188,8 @@ public class MockStatement extends JDBC41Statement implements CallableStatement 
         );
 
         result = data.execute(context);
-        return result != null && result.length > 0 && result[resultIndex].data != null;
+
+        return localSql.startsWith("select");
     }
 
     private static final int[] unbox(List<Integer> list) {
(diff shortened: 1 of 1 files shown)


jOOQ/src/main/java/org/jooq/tools/jdbc/MockStatement.java
```diff
@@ -188,7 +188,8 @@ private boolean execute0(
         );
 
         result = data.execute(context);
-        return result != null && result.length > 0 && result[resultIndex].data != null;
+
+        return localSql.startsWith("select");
```
lukaseder: There are many ways in which this will not be robust enough, among which
- there could be a parenthesis (or several) leading the  `select` keyword
- the keyword might not be in lower case
- there might be a comment prepended to the keyword
- there might be a `with` clause prepended to `select`

lukaseder:

This integration test works for me:

```java
    @Test
    public void testInsert() {
        DSLContext e = DSL.using(new MockConnection(new MockDataProvider() {

            @Override
            public MockResult[] execute(MockExecuteContext ctx) throws SQLException {
                return new MockResult[] { new MockResult(15, null) }; // Update count here
            }
        }), SQLDialect.POSTGRES);

        assertEquals(15,
            e.insertInto(TABLE1, FIELD_ID1)
             .values(1)
             .execute());
    }
```

How are you currently using this API?

codeturner:
Sorry, my mistake.  When I went back to get a simple unit test to validate, I noticed that I had an error in my custom implementation of MockDataProvider.  It was creating a MockResult that was incorrect.

I'll close this pull request.

Thanks for your time.  This is a great library.  Keep up the great work.

lukaseder:
No worries, thank you for the feedback!

Issue #5699, opened 2016-12-08
Add support for PostgreSQL ON CONFLICT .. DO NOTHING


fleroy84:
Hello,

Postgresql 9.5 allows one to specify one field for instruction "on conflict … do ignore".

In class "InsertQueryImpl" in the 3.9.0-SNAPSHOT version, I can see that in case of an "on conflict … do update" case, fields can be passed as parameters to build the SQL query but not for the "on conflict … do ignore" case. 

In my project, I have a table with a PK and a column with a unique constraint.
When I use the instruction  "on conflict … do ignore", both constraints are checked because jOOQ does not specify any field. As a result, Postgresql 9.5 checks every field with a unique constraint.
I want to specify the field on which the constraint will be verified.

I would like, if possible, something like : 

getDslContext()
      .insertInto([myTable])
      .set([myRecord])
      .onConflict([myField])
      .doIgnore())

Thank you.

lukaseder:
Oh, interesting. I think we've overlooked that clause when `ON CONFLICT` was implemented. That's certainly worth adding, especially since we already have a synthetic `onDuplicateKeyIgnore()` clause.

Thanks for the suggestion.

safaci2000:
+1 on this feature. I thought it was already implemented. It's a must have for me as well.

PR #5702, opened 2016-12-09
[#5699] Specify field for instruction "on conflict … do ignore" 

fleroy84:
Suggestion of implementation for ticket #5699

### DIFF
--- a/jOOQ/src/main/java/org/jooq/impl/InsertImpl.java
+++ b/jOOQ/src/main/java/org/jooq/impl/InsertImpl.java
@@ -123,8 +123,7 @@ class InsertImpl<R extends Record, T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11,
     InsertValuesStepN<R>,
     InsertSetStep<R>,
     InsertSetMoreStep<R>,
-    InsertOnDuplicateSetMoreStep<R>,
-    InsertOnConflictDoUpdateStep<R>,
+    InsertOnDuplicateSetMoreStep<R>, InsertOnConflictDoUpdateStep<R>,
     InsertResultStep<R> {
 
     /**
@@ -580,6 +579,13 @@ class InsertImpl<R extends Record, T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11,
 
     @Override
     public final InsertImpl doUpdate() {
+        onDuplicateKeyUpdate = true;
+        return this;
+    }
+
+    @Override
+    public final InsertImpl doNothing() {
+        getDelegate().onDuplicateKeyIgnore(true);
         return this;
     }
 
@@ -590,7 +596,6 @@ class InsertImpl<R extends Record, T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11,
 
     @Override
     public final InsertImpl onConflict(Collection<? extends Field<?>> keys) {
-        onDuplicateKeyUpdate = true;
         getDelegate().onConflict(keys);
         return this;
     }
--- a/jOOQ/src/main/java/org/jooq/impl/InsertQueryImpl.java
+++ b/jOOQ/src/main/java/org/jooq/impl/InsertQueryImpl.java
@@ -54,10 +54,10 @@ import static org.jooq.impl.DSL.select;
 import static org.jooq.impl.DSL.selectFrom;
 import static org.jooq.impl.DSL.selectOne;
 import static org.jooq.impl.DSL.table;
+import static org.jooq.impl.Tools.DataKey.DATA_INSERT_SELECT_WITHOUT_INSERT_COLUMN_LIST;
 import static org.jooq.impl.Tools.EMPTY_FIELD;
 import static org.jooq.impl.Tools.aliasedFields;
 import static org.jooq.impl.Tools.fieldNames;
-import static org.jooq.impl.Tools.DataKey.DATA_INSERT_SELECT_WITHOUT_INSERT_COLUMN_LIST;
 
 import java.util.Arrays;
 import java.util.Collection;
@@ -140,7 +140,6 @@ final class InsertQueryImpl<R extends Record> extends AbstractStoreQuery<R> impl
     public final void onDuplicateKeyIgnore(boolean flag) {
         this.onDuplicateKeyUpdate = false;
         this.onDuplicateKeyIgnore = flag;
-        this.onConflict = null;
     }
 
     @Override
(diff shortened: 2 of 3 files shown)


lukaseder:
Thank you very much for your suggestion. I'll review it shortly

jOOQ/src/main/java/org/jooq/InsertOnConflictDoActionStep.java
```diff
@@ -59,11 +59,17 @@
  *
  * @author Lukas Eder
  */
-public interface InsertOnConflictDoUpdateStep<R extends Record> {
+public interface InsertOnConflictDoActionStep<R extends Record> {
```
lukaseder: We cannot rename types due to backwards compatibility constraints. It's OK if the type name doesn't fully match all of its features...

jOOQ/src/main/java/org/jooq/InsertOnConflictDoActionStep.java
```diff
@@ -59,11 +59,17 @@
...
      */
     @Support({ POSTGRES_9_5 })
     InsertOnDuplicateSetStep<R> doUpdate();
+
+    /**
+     * Add the <code>DO IGNORE</code> clause.
+     */
+    @Support({ POSTGRES_9_5 })
+    InsertFinalStep<R> doIgnore();
 }
```
lukaseder: The actual syntax in PostgreSQL is actually `DO NOTHING`. I would prefer if we repeat that in our API as well

jOOQ/src/main/java/org/jooq/impl/InsertImpl.java
```diff
@@ -580,6 +580,13 @@ public final InsertImpl defaultValues() {
...
     @Override
     public final InsertImpl doUpdate() {
+        onDuplicateKeyUpdate = true;
+        return this;
+    }
+
+    @Override
+    public final InsertImpl doIgnore() {
+        getDelegate().onDuplicateKeyIgnore(true);
         return this;
```
lukaseder: I think that from a "principle of least surprise" point of view, we should now set `onDuplicateKeyUpdate` in both of these methods, and we should stop setting the flag in `onConflict()`

jOOQ/src/main/java/org/jooq/impl/InsertQueryImpl.java
```diff
@@ -78,12 +78,14 @@
...
 import org.jooq.exception.SQLDialectNotSupportedException;
+import org.jooq.tools.JooqLogger;
 
 /**
  * @author Lukas Eder
  */
 final class InsertQueryImpl<R extends Record> extends AbstractStoreQuery<R> implements InsertQuery<R> {
 
+    private static final JooqLogger LOGGER = JooqLogger.getLogger(InsertQueryImpl.class);
     private static final long        serialVersionUID = 4466005417945353842L;
```
lukaseder: Minor thing: Our logger is always called `log`

jOOQ/src/main/java/org/jooq/impl/InsertQueryImpl.java
```diff
@@ -278,10 +279,21 @@ else if (onDuplicateKeyIgnore) {
...
-                       .keyword("on conflict do nothing")
-                       .end(INSERT_ON_DUPLICATE_KEY_UPDATE);
+                    ctx.formatSeparator().start(INSERT_ON_DUPLICATE_KEY_UPDATE).keyword("on conflict");
+
+                    if (onConflict != null && onConflict.size() > 0) {
+                        if (onConflict.size() != 1) {
+                            LOGGER.info(
+                                    "ON DUPLICATE KEY IGNORE accept only one field for unique/exclusion constraint, " +
+                                            "let's take the first one on the list {}", onConflict.get(0).getName());
+                        }
```
lukaseder: Is this really correct? Can you point me to the docs where this is specified? I would imagine that we should be able to list the columns of a composite key here...

jOOQ/src/main/java/org/jooq/impl/InsertQueryImpl.java
```diff
@@ -278,10 +279,21 @@ else if (onDuplicateKeyIgnore) {
...
+                        ctx.sql(" (");
+                        boolean qualify = ctx.qualify();
+                        ctx.qualify(false).visit(onConflict.get(0)).qualify(qualify);
+                        ctx.sql(") ");
+                    }
+
+                    ctx.sql(" do nothing").end(INSERT_ON_DUPLICATE_KEY_UPDATE);
```
lukaseder: Keywords are passed to the `ctx` using `ctx.keyword()`, i.e. this should be:

```java
ctx.sql(' ').keyword("do nothing")...
```

jOOQ/src/main/java/org/jooq/impl/InsertImpl.java
lukaseder: I see, you've done the latter in a separate commit:
https://github.com/jOOQ/jOOQ/pull/5702/commits/8eb788082d57fd34b6cf82c3833e253c86f47ca7

jOOQ/src/main/java/org/jooq/impl/InsertQueryImpl.java
lukaseder: OK, that's been done in a separate commit: https://github.com/jOOQ/jOOQ/pull/5702/commits/b75cf702a8ee272fc7e62adf25c0cf463fb0e131

fleroy84:
You suggest to add a method “onConflictDoNothing()”. This method will be duplicate of “onDuplicateKeyIgnore()”, no ?
 Or may be you want this method to be coherent with postgresql instruction ?

```
getDslContext().insertInto[myTable])).set([myRecord]).onDuplicateKeyIgnore());

getDslContext().insertInto[myTable])).set([myRecord]).onConflictDoNothing());
```

This 2 lines of code, will do the same thing, right ?

jOOQ/src/main/java/org/jooq/impl/InsertQueryImpl.java
fleroy84: You're right, an error has been made in my test case ! I will fix it.

lukaseder:
> These 3 lines of code, will do the same thing, right ?

Yes, they'll do the same thing.

I'll merge your updated suggestion and will further comment / fix things later on as the 3.9 release is due this week. Thanks again for your contribution, this is greatly appreciated!

PR #5767, opened 2017-01-03
Add a comment to the manual's section about MatcherRules regarding unqualified / qualified identifier matching

deinspanjer:
Add note about the fact the expression is tested against the unqualified name first then the qualified name.
Might help prevent confusion similar to this thread:
https://groups.google.com/forum/#!topic/jooq-user/XwRzNf5Gtjs


deinspanjer:
That is great, I'm glad that this text could be of some use. I just did not want to make a bunch of suggestions on things to do and not offer any assistance.

lukaseder:
Hey, sure. Your assistance is greatly appreciated. Looking forward to more of it! :)

Issue #5858, opened 2017-02-09
Data loader unable to decode Base64 encoding binary types


codeturner:
I'm attempting to use the formatter to dump test data out into a csv file for each table in our schema.  I would then commit these files into our source repo so that other developers can load test data for development.  I then use loadInto to import the data.  This is by far the best solution I could come up with, since jOOQ already knows our schema, and it's easier to maintain the test data by editing csv files.  Plus, we can do a snapshot dump of the database to merge changes into the source repo.  Perfect!

This works great except when I bump into binary types.  The fix for issue [#3961] converts binary values to base64 encoded strings when formatting to XML, JSON, CSV, etc.  However, when trying to load back those values using loadInto, the converter doesn't know it needs to be decoded, and the data ends up corrupted.

I think base64 is a great way of representing a binary value, but can we fix the importer to recognize that the input is base64 encoded, and then automatically decode it?

I propose prepending a tag to the value, something like "BASE64:" so that when the converter sees that tag prepended to the value, it can decode it before use.  I think this is harmless since any user of the exported data will have to process the binary value before using it anyway.

Thoughts?

lukaseder:
Thank you very much for your report. The implicit specification of #3961 (and of other features) is for jOOQ's export functionality to be reversible on import. The fact that #3961 is not reversible is an oversight and should be fixed in all relevant places.

I'm not a fan of an arbitrary prefix, though. The decision to encode export data is based on the value's data type (i.e. `byte[]`). It would be more formally correct to make that decision based on the `Field.getType()`, which is mostly the same. In any case, we should definitely make that decision based on `Field.getType()` upon import (much like we parse numbers if the `Field.getType() == Integer.class`, for instance).

lukaseder:
Do note there is also a pending feature request that allows for specifying custom import/export formatters: https://github.com/jOOQ/jOOQ/issues/2741

codeturner:
Another option I thought about was using MySQL's way of specifying a binary using the hexidecimal literal format: X'4D7953514C'.  But this isn't a SQL standard, so that seemed as arbitrary to me as a prefix.

I was hesitant in making the assumption that all imported values to type byte[] would be expected to be in base64 format.  That's why I felt that the importing data type should have a way of indicating to the converter that it is encoded and needs to be decoded.  But if we can make that assumption to get it to work, I'd be beyond happy.

[...]

PR #5859, opened 2017-02-09
[#5858] prefix encoding of binary types with tag and decode in loadInto conversion

codeturner:
[#5858] prefix base64 encoding of binary types with 'BASE64:' and decode these values when encountering them in loadInto conversion


lukaseder:
Thank you very much for your suggested fix. As commented in issue #5858 (https://github.com/jOOQ/jOOQ/issues/5858#issuecomment-278887633), I do not agree with adding an arbitrary prefix to the encoded output.

I will thus reject this pull request. Discussion will be continued directly on #5858

Issue #5898, opened 2017-02-21
Field.contains() and Custom binding


lmarchau:
Hello, 
we define custom binding for CiText DataType (postgres):

`<forcedTypes>
  
                      <forcedType>  

                        <userType>com.example.CiText</userType>  

                        <binding>com.example.CiTextBinding</binding>  

                        <expression>  
                          schema.rule.name |  
                          schema.core.ref_object.label  

                        </expression>  

                      </forcedType>  

                    </forcedTypes>  `  

Using this custom binding on equals constraint (where clause) succeed. On the other hand, contains constraint ignore custom binding.

`Tools.escapeForLike(value, configuration)` in Contains.condition() method use varchar datatype on default.

Contains.condition() may use `Tools.escapeForLike(value, field, configuration)` and support custom binding define into field (lhs variable contains field configuration).

Thank you

Cheers !!

lukaseder:
Thank you very much for reporting. May I ask you to provide an example of your Java / jOOQ code where this fails? Might be the same issue as https://github.com/jOOQ/jOOQ/issues/4556, but just to be sure...

lmarchau:
Here an example of failing code: 

https://gist.github.com/lmarchau/8fe7b94b01d0fc85fabf006b28a22738 

Problem comes from https://github.com/jOOQ/jOOQ/blob/master/jOOQ/src/main/java/org/jooq/impl/Tools.java#L2210. This method casts object to String by force when it calls the val () method.

Cheers

lukaseder:
Thank you very much for the clarifications and the test case. I can now also reproduce this:

```java
class StringWrapper {
    final String string;
    StringWrapper(String string) {
        this.string = string;
    }
    @Override
    public String toString() {
        return "nope";
    }
}

public void testContainsWithConverter() throws Exception {
    Converter<String, StringWrapper> converter = Converter.ofNullable(
        String.class,
        StringWrapper.class,
        StringWrapper::new,
        s -> s.string
    );

    DataType<StringWrapper> type = SQLDataType.VARCHAR.asConvertedDataType(converter);
    B book =
    create().selectFrom(TBook())
            .where(field(TBook_TITLE().getName(), type).contains(new StringWrapper("Animal")))
            .fetchOne();

    assertEquals("Animal Farm", book.get(TBook_TITLE()));
}
```

Indeed, that's not strictly the same issue as #4556.

lukaseder:
The test I've added here no longer fails:

[...]

PR #5915, opened 2017-02-25
[#5898] Field.contains(T) must not call toString() on T value

lmarchau:
Suggestion for issue #5898 

I found same issue with Field.startsWith(T) and Field.endsWith(T) with same cause, so I include fix here.

### DIFF
--- a/jOOQ/src/main/java/org/jooq/impl/AbstractField.java
+++ b/jOOQ/src/main/java/org/jooq/impl/AbstractField.java
@@ -837,7 +837,7 @@ abstract class AbstractField<T> extends AbstractQueryPart implements Field<T> {
 
     @Override
     public final Condition startsWith(T value) {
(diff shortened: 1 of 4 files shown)


knutwannheden:
AFAICT this issue has already been solved in 3.12. Using a test as proposed [here](https://github.com/jOOQ/jOOQ/issues/5898#issuecomment-282289532) I was not able to reproduce the problem.

@lmarchau Would it be possible for you to confirm that the problem has been solved on jOOQ master also for your particular use case?

lmarchau:
Hi @knutwannheden ,

i'm not working on the project for more than 1 year and cannot access code now.

Sorry,

knutwannheden:
@lmarchau Thanks for the heads-up.

lukaseder:

PR #5938, opened 2017-03-02
Fix typo in `having` example java code

zellyn:
You can't do `count(*)` in Java.


lukaseder:
Thank you very much for reporting. Nice catch!

Issue #6063, opened 2017-04-12
Use eq / ne / etc in manual, rather than equal / notEqual / etc


lukaseder:
Users may run into the `equals()` /  `equal()` ambiguity when they learn jOOQ from the manual, as the manual makes use of the `equal()` method. We should use `eq()` instead.

Example:
https://www.jooq.org/doc/3.0/manual/sql-building/sql-statements/update-statement/#comment-3252197203

PR #6114, opened 2017-04-20
[#6063] change Field.equal() and Field.notEqual() to Field.eq() and Field.ne() in manual pages

AnkBurov:
Open issue is here: https://github.com/jOOQ/jOOQ/issues/6063


lukaseder:
Thank you very much for your help!

Issue #6022, opened 2017-03-25
Add Name.fields(Name, ...)


lukaseder:
CTE can currently only be constructed using `String` for names, not `Name`

PR #6136, opened 2017-04-23
[#6022] add Name.fields(Name...) method along with the generated ones

AnkBurov:
PR to support such queries:

```java
dsl.with(USER.getQualifiedName()
                .fields(USER.ID.getQualifiedName(),
                        USER.LOGIN.getQualifiedName(),
                        USER.CATEGORY_ID.getQualifiedName())
                .as(dsl.select(
                        USER.ID,
                        USER.LOGIN,
                        USER.CATEGORY_ID)
                        .from(USER)))
                .select()
                .from(USER.getQualifiedName())
                .fetchInto(User.class);
```

### DIFF
--- a/jOOQ/src/main/java/org/jooq/Name.java
+++ b/jOOQ/src/main/java/org/jooq/Name.java
@@ -137,6 +137,16 @@ public interface Name extends QueryPart {
     @Support({ FIREBIRD, H2, HSQLDB, POSTGRES })
     DerivedColumnList fields(String... fieldNames);
 
+    /**
+     * Add a list of fields to this name to make this name a
+     * {@link DerivedColumnList}.
+     * <p>
+     * The <code>DerivedColumnList</code> can then be used along with a
+     * subselect to form a {@link CommonTableExpression} to be used with
+     * <code>WITH</code> clauses.
+     */
+    @Support({FIREBIRD, H2, HSQLDB, POSTGRES})
+    DerivedColumnList fields(Name... fieldNames);
 
     /**
      * Add a list of fields to this name to make this name a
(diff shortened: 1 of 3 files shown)


lukaseder:
Thank you very much for your suggested fix. I will review it shortly. In the meantime, may I ask you to read our contribution guidelines, and in particular, sign our transfer of rights agreement, which allows us to dual-license our software:

jOOQ/src/main/java/org/jooq/Name.java
```diff
@@ -435,6 +445,270 @@
...
+     * Add a list of fields to this name to make this name a
+     * {@link DerivedColumnList}.
+     * <p>
+     * The <code>DerivedColumnList</code> can then be used along with a
+     * subselect to form a {@link CommonTableExpression} to be used with
+     * <code>WITH</code> clauses.
+     */
+    @Generated("This method was generated using jOOQ-tools")
+    @Support({ FIREBIRD, H2, HSQLDB, POSTGRES })
+    DerivedColumnList2 fields(Name fieldName1, Name fieldName2);
```
lukaseder: Oh, I'm sorry you went through all this trouble :) We have a code generator (not open source), which generates this degree-22 API madness for us...

jOOQ/src/main/java/org/jooq/impl/AbstractName.java
```diff
@@ -86,7 +86,13 @@ public final DerivedColumnListImpl fields(String... fieldNames) {
         return new DerivedColumnListImpl(last(), fieldNames);
     }
 
+    @Override
+    public final DerivedColumnListImpl fields(Name... fieldNames) {
+        if (getName().length != 1)
+            throw new IllegalStateException("Cannot create a DerivedColumnList from a qualified name : " + Arrays.asList(getName()));
 
+        return new DerivedColumnListImpl(last(), Tools.fieldNamesToStrings(fieldNames));
```
lukaseder: Hmm, I had rather refactor the `DerivedColumnListImpl` along with this change in one go, passing the `Name` references along such that the interesting information (the quoted flag) will not be removed...

lukaseder:
Thanks again for your suggestion. I have changed the implementation such that `DerivedColumnListImpl` now works with `Name` rather than `String`

jOOQ/src/main/java/org/jooq/Name.java
AnkBurov: I assumed this when I didn't find this tooling for 22 API code generation, but still currently here is just a template and single loop so it's nothing to worry about. :)

jOOQ/src/main/java/org/jooq/impl/AbstractName.java
AnkBurov: It's ok. If it's already fixed (whole DerivedColumnListImpl to Name and Name[] fields) then there is nothing to repair and this PR is not needed.

jOOQ/src/main/java/org/jooq/Name.java
lukaseder: Alright ;)

PR #6207, opened 2017-05-10
SQL Server compatability

rstub:
- add GO after DB creation
- fix datetime formating

### DIFF
--- a/jOOQ-examples/Sakila/sql-server-sakila-db/sql-server-sakila-schema.sql
+++ b/jOOQ-examples/Sakila/sql-server-sakila-db/sql-server-sakila-schema.sql
@@ -11,6 +11,7 @@ http://www.etl-tools.com
 */
 
 CREATE DATABASE sakila;
+GO
 USE sakila;
 
 --


lukaseder:
Thank you very much for your suggestion. From what I can remember, the existing version of the script works just fine (tried with SQL Server 2014). Would you mind explaining what problems your suggestion is solving, before I review the change?

rstub:
When I execute the Schema script from SSMS on SQL Server 2016 (Developer Edition) I get an error message that the sakila-DB does not exist from "USE sakila;" 

> Meldung 911, Ebene 16, Status 1, Zeile 14
> Die sakila-Datenbank ist nicht vorhanden. Stellen Sie sicher, dass der Name richtig eingegeben wurde.

After that all other statements fail as well and the schema is not created. Adding the GO between CREATE and USE solves this.

When I execute the data insert script after creating the schema I get *many* error message that converting the varchar to datetime resulted in a value outside the allowed range:

> Meldung 242, Ebene 16, Status 3, Zeile 48
> Bei der Konvertierung eines varchar-Datentyps in einen datetime-Datentyp liegt der Wert außerhalb des gültigen Bereichs.

Basically no data is inserted. After inserting the 'T's required by ISO 8601, the script works ok. BTW, datetime2 seems to be less picky:

    SELECT 
    	'fails' -- CAST('2006-02-15 05:02:19.000' AS datetime)
    ,	CAST('20060215 05:02:19.000' AS datetime)
    ,	CAST('2006-02-15T05:02:19.000' AS datetime)
    ,	CAST('2006-02-15 05:02:19.000' AS datetime2)
    ,	CAST('20060215 05:02:19.000' AS datetime2)
    ,	CAST('2006-02-15T05:02:19.000' AS datetime2)`

rstub:
BTW, the first error does not occur when I use `sqlcmd -i sql-server-sakila-insert-data.sql`. However, the second error prevails even when using `sqlcmd -i sql-server-sakila-insert-data.sql`.

lukaseder:
Oh, interesting, thank you very much for the explanations. Perhaps, something changed between versions 2014 and 2016... On my installation, this statement works perfectly:

```sql
SELECT 
	CAST('2006-02-15 05:02:19.000' AS datetime)
,	CAST('20060215 05:02:19.000' AS datetime)
,	CAST('2006-02-15T05:02:19.000' AS datetime)
,	CAST('2006-02-15 05:02:19.000' AS datetime2)
,	CAST('20060215 05:02:19.000' AS datetime2)
,	CAST('2006-02-15T05:02:19.000' AS datetime2)
```

Result:

image

The interesting thing is that formatted values also do not include the (in my opinion obnoxious) `T` character. After all, ISO 8601 allows for omitting it.

I'm surprised that such a change would be implemented in SQL Server, but well, better not expect too much leniency if it occasionally doesn't

Thanks again for your contribution and for your explanations!

Issue #5645, opened 2016-11-06
Add DSL.localDateAdd() localDateSub(), localDateDiff()


lukaseder:
These would work the same way as `DSL.dateAdd()`, `DSL.dateSub()`, `DSL.dateDiff()`, except that they operate on `LocalDate` instead of `java.sql.Date`

bitbrain:
Preparing a Pull Request right now. @lukaseder you are okay with that or is there already work on this?

lukaseder:
Sure, thank you very much for your offer, @bitbrain. May I just ask you to read this first and sign the transfer of rights agreement linked from the document:
https://github.com/jOOQ/jOOQ/blob/master/CONTRIBUTING.md

lukaseder:
Fixed this, partially taking over @bitbrain's PR #6215. There's still content missing, namely working with timestamps / local date time, which has been extracted into a new issue: https://github.com/jOOQ/jOOQ/issues/6220.

PR #6215, opened 2017-05-12
Add localDateAdd() localDateSub(), localDateDiff() support to DSL

bitbrain:
This PR aims to close issue #5645.

It introduces the following new methods in overloaded variants inside the DSL:
- localDateAdd
- localDateSub
- localDateDiff

### DIFF
--- a/jOOQ/src/main/java/org/jooq/impl/DSL.java
+++ b/jOOQ/src/main/java/org/jooq/impl/DSL.java
@@ -12691,6 +12691,46 @@ public class DSL {
     }
 
     /**
+     * Get the date difference in number of days.
+     *
+     * @see Field#sub(Field)
+     */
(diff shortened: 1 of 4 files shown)


lukaseder:

I'll merge this, as your suggested API is already as good as it gets. Some comments about the implementations:

1. I had rather we don't duplicate the implementation (duplicating the API is inevitable). I'd rather refactor the existing implementations to accept both `Field<java.sql.Date>` and `Field<java.time.LocalDate>`, for instance. Refactoring might be tricky, though, as you don't have access to the integration tests or the commercial code.
2. I see you've stripped out all the whitespace from the open-source edition and auto-formatted the code. This makes merging the implementation a bit harder. In this case, it's irrelevant, because I'll refactor the existing implementation after merging, but as a general hint: It's always good to avoid whitespace changes in pull requests

Thanks again for your suggestion!

PR #6255, opened 2017-05-19
Add better Enum to Enum conversion support in org.jooq.tools.Convert

triptec:
Hello!
I'm wondering if there's a reason we haven't got a simple enum to enum.
I might have missed it or something but if it doesn't exist it would be nice to have it.

### DIFF
--- a/jOOQ/src/main/java/org/jooq/tools/Convert.java
+++ b/jOOQ/src/main/java/org/jooq/tools/Convert.java
@@ -980,6 +980,17 @@ public final class Convert {
                     }
                 }
 
+                // Some users may find it useful to convert Enum
+                // literals to Enum values without a Converter
+                else if (java.lang.Enum.class.isAssignableFrom(fromClass) && java.lang.Enum.class.isAssignableFrom(toClass)) {
+                    try {
+                        return (U) java.lang.Enum.valueOf((Class) toClass, ((EnumType) from).getLiteral());
+                    }
+                    catch (IllegalArgumentException e) {
+                        return null;
+                    }
+                }
+
                 // [#1624] UUID data types can be read from Strings
                 else if ((fromClass == String.class) && toClass == UUID.class) {
                     try {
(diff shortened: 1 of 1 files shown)


lukaseder:
Thank you very much for your suggestion. It may not be as simple as this, though. In particular, your suggestion shows that there are essentially two types of enum in jOOQ:

- Ordinary enums `Enum.valueOf()`
- `EnumType`, where `getLiteral()` is the interesting value

There should be 6 different conversion options:

- String to `Enum` (via `name()`)
- Integer to `Enum` (via `ordinal()`)
- String to `EnumType` (via `getLiteral()`)
- `Enum` to String (via `name()`)
- `Enum` to Integer (via `ordinal()`)
- `EnumType` to String (via `getLiteral()`)

Also, may I ask you to read our guidelines, and in particular, sign our transfer of rights agreement, which allows us to dual-license your contribution:

triptec:
I signed the transfer of right and I also updated the pull request to check the "from" instance.

I agree about the other conversions, but I drew my conclusions from  https://github.com/jOOQ/jOOQ/blob/master/jOOQ/src/main/java/org/jooq/tools/Convert.java#L538-L545
```
                // All types can be converted into String
                else if (toClass == String.class) {
                    if (from instanceof EnumType) {
                        return (U) ((EnumType) from).getLiteral();
                    }

                    return (U) from.toString();
                }
```
and https://github.com/jOOQ/jOOQ/blob/master/jOOQ/src/main/java/org/jooq/tools/Convert.java#L972-L981
```                
                // [#1448] Some users may find it useful to convert string
                // literals to Enum values without a Converter
                else if ((fromClass == String.class) && java.lang.Enum.class.isAssignableFrom(toClass)) {
                    try {
                        return (U) java.lang.Enum.valueOf((Class) toClass, (String) from);
                    }
                    catch (IllegalArgumentException e) {
                        return null;
                    }
                }
```
and while I agree that it would be nice to have the 6 conversion options I'm only scratching my own itch and I think it's inline with those other parts of the code cited above. Further more I can't say I know how to implement what you are suggesting of the bat.

Not a native english speaker, might sound abrupt.

lukaseder:
OK, thanks for your feedback. No worries. I'll review this again and merge, soon. Quite possibly, I'll refactor things afterwards to make them more generic, but it's a good start!

lukaseder:
OK, I won't implement the ordinal conversion part (yet). That might not be the expected behaviour in some cases. Let's wait until someone requests it. As for the rest, that's done.

triptec:
Thanks

Issue #6266, opened 2017-05-23
JoinTable.join(...).onKey() does not work


RichardBradley:
I love the [join "onKey()" functionality](https://www.jooq.org/doc/3.10/manual/sql-building/sql-statements/select-statement/join-clause/#N43BF2), I've always missed that in SQL.

I have found that the "onKey()" cannot infer which FK to use when there are more than one join in the query. Is that just an omission? I can't see any reason why that wouldn't work.

For example, if I do:

```
create.select()
      .from(AUTHOR)
      .join(BOOK).onKey()
      .join(LANGUAGE).onKey()
      .fetch();
```

... then I get:

```
org.jooq.exception.DataAccessException: Key ambiguous between tables "author"
  join "book"
  on "book"."author_id" = "author"."id" and "language"
```

I think all that would be required to fix this would be to handle "JoinTables" as a special case in the [AbstractTable.getReferencesTo](https://github.com/jOOQ/jOOQ/blob/3605697b8d0c74de6b2fbafc676dfc78cd74d11c/jOOQ/src/main/java/org/jooq/impl/AbstractTable.java#L482) method.

(It might also be nice to add more info to the "Key ambiguous" exception, which doesn't currently list the FKs which were found.)

lukaseder:
This should work as expected, but the question is, is your path from `AUTHOR` to `LANGUAGE` really non-ambiguous? What's the DDL to produce these tables? (note, our manual examples have `BOOK.AUTHOR_ID` and `BOOK.CO_AUTHOR_ID` columns, which make this type of join ambiguous)

RichardBradley:
Hmm, maybe I have slightly mis-stated or misunderstood the problem, sorry.

I have a schema which has a non-ambiguous path along my multi-table join which is failing with the above exception. If I put a breakpoint at the "throw" in JoinTable.onKey, then I see that both "leftToRight" and "rightToLeft" are empty (when an FK exists), which I think matches the above diagnosis.

I'll see if I can prepare a failing test case. What is the best example project to start from? Should I just fork something from https://github.com/jOOQ/jOOQ/tree/master/jOOQ-examples or maybe look for some unit tests near the "onKey" feature in the main repos?

lukaseder:
If I have the DDL for those exact tables, I'll probably be able to reproduce...

RichardBradley:
> I'll see if I can prepare a failing test case. 

There aren't any tests? I didn't see that coming. 

[...]

PR #6271, opened 2017-05-24
Improve onKey() join error messages, when join expression is ambiguous

RichardBradley:
This improves the error messages when join onKey() fails, as discussed in #6266

I have signed https://www.jooq.org/legal/contributions as Richard Bradley

### DIFF
--- a/jOOQ/src/main/java/org/jooq/impl/JoinTable.java
+++ b/jOOQ/src/main/java/org/jooq/impl/JoinTable.java
@@ -49,7 +49,6 @@ import static org.jooq.Clause.TABLE_JOIN_OUTER_APPLY;
 import static org.jooq.Clause.TABLE_JOIN_OUTER_FULL;
 import static org.jooq.Clause.TABLE_JOIN_OUTER_LEFT;
 import static org.jooq.Clause.TABLE_JOIN_OUTER_RIGHT;
-import static org.jooq.Clause.TABLE_JOIN_PARTITION_BY;
 import static org.jooq.Clause.TABLE_JOIN_SEMI_LEFT;
 import static org.jooq.Clause.TABLE_JOIN_STRAIGHT;
 import static org.jooq.Clause.TABLE_JOIN_USING;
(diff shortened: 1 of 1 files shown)


jOOQ/src/main/java/org/jooq/impl/JoinTable.java
```diff
@@ -49,7 +49,6 @@
 import static org.jooq.Clause.TABLE_JOIN_OUTER_FULL;
 import static org.jooq.Clause.TABLE_JOIN_OUTER_LEFT;
 import static org.jooq.Clause.TABLE_JOIN_OUTER_RIGHT;
-import static org.jooq.Clause.TABLE_JOIN_PARTITION_BY;
```
lukaseder: These imports are not needed in the jOOQ Open Source edition, but in order to preserve the exact same line numbers with the commercial distributions, we don't clean up unneeded imports.

I'll put these back again after merging.

jOOQ/src/main/java/org/jooq/impl/JoinTable.java
```diff
@@ -562,7 +558,14 @@ else if (rightToLeft.size() == 1 && leftToRight.size() == 0) {
...
 
-        throw onKeyException();
+        if (rightToLeft.isEmpty() && leftToRight.isEmpty()) {
+            throw new DataAccessException("No matching Key found between tables " + lhs + " and " + rhs);
+        } else {
+            throw new DataAccessException(
+                    "`onKey()` was ambiguous between tables " + lhs + " and " + rhs + ".\n" +
+                            "Found possible keys: " + leftToRight + " and " + rightToLeft + ".\n" +
+                            "Please use .onKey(ForeignKey) instead.");
+        }
```
lukaseder: I see what you mean. I'll adapt this code in two ways, though:

1. I'll factor out the different messags back into that utility method, to increase reuse, because:
2. I suspect that ambiguity can also happen with the other ways of using `onKey()` in more complicated setups

PR #6287, opened 2017-05-26
Upgrade to Spring Boot 1.5.3 and some polish

michael-simons:
If you still need it, contains some more polishing in the pom and the tests.

### DIFF
--- a/jOOQ-examples/jOOQ-spring-boot-example/pom.xml
+++ b/jOOQ-examples/jOOQ-spring-boot-example/pom.xml
@@ -11,7 +11,7 @@
 
     <groupId>org.jooq</groupId>
     <artifactId>jooq-spring-boot-example</artifactId>
-    <version>${org.jooq.version}</version>
+    <version>${jooq.version}</version>
     <packaging>jar</packaging>
     <name>jOOQ Spring Boot Example</name>
 
@@ -26,7 +26,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <start-class>org.jooq.example.spring.Application</start-class>
-        <org.jooq.version>3.10.0-SNAPSHOT</org.jooq.version>
+        <jooq.version>3.10.0-SNAPSHOT</jooq.version>
         <java.version>1.8</java.version>
     </properties>
 
@@ -39,7 +39,6 @@
         <dependency>
             <groupId>org.jooq</groupId>
             <artifactId>jooq</artifactId>
-            <version>${org.jooq.version}</version>
         </dependency>
 
         <dependency>
@@ -127,7 +126,6 @@
             <plugin>
                 <groupId>org.jooq</groupId>
                 <artifactId>jooq-codegen-maven</artifactId>
-                <version>${org.jooq.version}</version>
 
                 <executions>
                     <execution>
--- a/jOOQ-examples/jOOQ-spring-boot-example/pom.xml
+++ b/jOOQ-examples/jOOQ-spring-boot-example/pom.xml
@@ -6,7 +6,7 @@
     <parent>
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-starter-parent</artifactId>
-        <version>1.3.3.RELEASE</version>
+        <version>1.5.3.RELEASE</version>
     </parent>
 
     <groupId>org.jooq</groupId>
--- a/jOOQ-examples/jOOQ-spring-boot-example/src/test/java/org/jooq/example/spring/QueryTest.java
+++ b/jOOQ-examples/jOOQ-spring-boot-example/src/test/java/org/jooq/example/spring/QueryTest.java
@@ -19,15 +19,15 @@ import org.jooq.example.db.h2.tables.records.BookRecord;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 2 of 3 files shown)


jOOQ-examples/jOOQ-spring-boot-example/pom.xml
```diff
@@ -11,7 +11,7 @@
 
     <groupId>org.jooq</groupId>
     <artifactId>jooq-spring-boot-example</artifactId>
-    <version>${org.jooq.version}</version>
+    <version>${jooq.version}</version>
```
lukaseder: Hmm, I see, this change may make sense in general. But currently, all the examples use this property naming scheme...

jOOQ-examples/jOOQ-spring-boot-example/pom.xml
```diff
@@ -39,7 +39,6 @@
         <dependency>
             <groupId>org.jooq</groupId>
             <artifactId>jooq</artifactId>
-            <version>${org.jooq.version}</version>
```
lukaseder: Same here. We could reference some external or default jOOQ version instead of the latest snapshot build from here on GitHub, but that would incur quite some maintenance work to keep up to date...

jOOQ-examples/jOOQ-spring-boot-example/src/test/java/org/jooq/example/spring/QueryTest.java
```diff
@@ -19,15 +19,15 @@
...
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.junit4.SpringRunner;
 
 /**
  * @author Lukas Eder
  * @author Thomas Darimont
  */
-@RunWith(SpringJUnit4ClassRunner.class)
-@SpringApplicationConfiguration(classes = Application.class)
+@RunWith(SpringRunner.class)
```
lukaseder: Interesting. How is this different?

michael-simons: The other runner is just icing on the cake atm, the SpringBootTest autodetects the "main" configuration class if it's in the same package, usually people omit if not necessary.

lukaseder:
Thanks for your suggestions. And sorry for the lack of coordination. I know I said I'd be jumping for beers, but then I quickly fixed this after Phil explained it in his tweet :) So, now we're left with two unrelated changes (which should be looked at in a broader conflict), and the one I've already done.

I'll close this then.

jOOQ-examples/jOOQ-spring-boot-example/src/test/java/org/jooq/example/spring/QueryTest.java
lukaseder: I see, thanks for the explanation.

jOOQ-examples/jOOQ-spring-boot-example/pom.xml
michael-simons: I think it's better using the "spring" property because that's mimics the stuff you get from start.spring.io which is the recommend way with the Spring Boot BOM. You could set jooq.version to the value of org.jooq.version… But i'd remove the explicit version in the dependencies… IMHO the usage of the Spring Boot BOM is highly recommend to get version of stuff that works well together.

michael-simons:
Fine with me… :) No offense taken.
But while fixing that I thought: Hey, there's no `@JooqTest` annotation analog to `@DataJpaTest` or `@JdbcTest`. Those annotation fires up only the part of the container you'll actually need to run the integration test against a datasource and that is generally a lot fast than firing up the whole application.

I'll see if I have time next week to contribute something.

jOOQ-examples/jOOQ-spring-boot-example/pom.xml
lukaseder: What if I fix a really really nasty bug in say jOOQ 3.17.348 but Spring Boot is still on 3.17.342 and when Spring Boot upgrades to 3.17.348, they inadvertedly also upgrade from Hibernate 4 to 6? :)

Dependencies.

OK, would you mind creating a separate issue for this?

michael-simons: Yeah, that could happen :) But that's also the reason they recommend overwriting the managed properties, if you need a special version. If jOOQ 3.17.349 has a new module which is included via the Bill of Materials, you won't have the correct version if you set org.jooq.version to whatever.
Yeah, dependencies are fun.

lukaseder: So essentially, once you're done having fun writing 1-2 hello worlds, you'll inevitably need a special version of everything... :)

Oh well.

wilkinsona: I agree with @michael-simons. If you override Boot's `${jooq.version}` then you'll get a consistent version for all of jOOQ as we manage the version of its main modules and its Maven plugin all with the same `jooq.version` property (as long as you're using `spring-boot-starter-parent` which you are here).

lukaseder: Thank you very much for your help. That must've been my misunderstanding then, or my lack of maven-fu. I can see it now. Will restore the commit.

lukaseder:
Original PR restored, including `SpringRunner` and version string

Issue #6350, opened 2017-06-19
aliasing in fields, TABLE.fieldStream().map(field -> field.as("category_"+field.getName())).toArray(Field[]::new)


kingbbode:
Hello.

I am using a DTO object that contains Jooq pojo.

```java
Class DTO {
 Long idx;
 JTableA atable;
 JTableB btable;
}
```

I use `fetchMap` to get the Map, and `ModelMapper` to DTO.

```java
 List<Map<String, Object>> record =
                dsl
                        .select(TABLE_A.IDX.as("Idx"))
                        .select(TABLE_A.fieldStream().map(field -> field.as("atable_"+field.getName())).toArray(Field[]::new))
                        .select(CATEGORY.fieldStream().map(field -> field.as("btable_"+field.getName())).toArray(Field[]::new))
                        .from(TABLE_A)
                        .join(TABLE_B).on(TABLE_B.IDX.eq(TABLE_A.TABLE_B_IDX))
                        .fetchMaps();
        List<DTO> contentResponses = modelMapper.map(record, new TypeToken<List<DTO>>() {
        }.getType());
```

It is cumbersome and unclear to associate an alias to the joins of two tables at this case.

```java
TABLE.fieldStream().map(field -> field.as("atable_"+field.getName())).toArray(Field[]::new)
```

So I think that the basic support of this feature can make JOOQ better.

For example,

```java
Field<?>[] fields(java.util.function.Function<? super Field<?>,? extends String> aliasFunction)
```
More simply,

```java
Field<?>[] fieldsWithPrefix(String prefix)
Field<?>[] fieldsWithPrefix(String prefix, String... fieldNames)
```

Thank you

Cheers !!

lukaseder:
Thank you very much for your suggestion. We'll look into this. Surely, there's room for improvement, but we shouldn't jump to adding too many of these utility functions. For instance, I don't think a `fieldsWithPrefix()` method is a good idea.

Note, you can already create a derived column list using functions when using `Table.as(String, Function)`

kingbbode:
In this situation, `Table.as(String, Function)` is inefficient.

```java
Table<JPostsRecord> posts = POSTS.as("POSTS", field -> "post_"+field.getName());
        Table<JCommentsRecord> comments = COMMENTS.as("COMMENTS", field -> "comments_"+field.getName());
        return dsl
                .select(posts.fields("post_ID", "post_TITLE"))
[...]

PR #6366, opened 2017-06-29
[#6350] Add Fields aliasFunction

kingbbode:
Suggestion for issue #6350 

Add Fields aliasFunction.

Thank you.

### DIFF
--- a/jOOQ/src/main/java/org/jooq/RecordType.java
+++ b/jOOQ/src/main/java/org/jooq/RecordType.java
@@ -34,6 +34,9 @@
  */
 package org.jooq;
 
+import java.util.function.BiFunction;
+import java.util.function.Function;
+
 /**
  * A record type for {@link Table}, {@link Cursor}, {@link Result} and other
  * objects.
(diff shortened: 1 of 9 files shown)


lukaseder:
Thank you very much for your suggestion. As a general rule of thumb, please wait with providing pull requests until we have reached consensus and agreement in what we really want to do. I'm really not convinced about this use-case as explained in #6350, yet. Here's some feedback:

1. I don't like the reuse of the method name `fields()`. The existing `fields()` method *extract* existing fields from a row type (or similar type) by name or index. Combining this operation with the aliasing operation seems counter intuitive to me, and won't scale in terms of API design (e.g. what if someone would like to see a feature where not the fields are aliased, but the tables).
2. Does this kind of functionality really make sense on the types `Row`, `Select`?
    - A `Row` is mostly used in predicates, e.g. to write things like `(A, B) = (1, 2)`. I don't see the value of aliasing individual fields in such a row.
    - Why alias fields that are contained in a `Select`? They should have already been aliased - I'm not sure about the purpose of re-aliasing them.

In any case, I'll reject this pull request, as the suggested functionality requires much more thinking first. We want to be careful not to add such convenience / utility API eagerly, especially if it is so easy to write a quick utility outside of jOOQ like you did yourself.

PR #6374, opened 2017-07-03
In jOOQ-spring-boot-example, use jooq starter instead of jdbc starter

rajadilipkolli:
Jooq starter includes jdbc starter as transitive dependency, so we can use this and jooq is obtained from jooq starter.
also upgrades to latest spring boot version


rajadilipkolli:
@lukaseder Can you help me why travis build failed. There is nothing I have changed in this pull request

lukaseder:
Thanks for your contribution. I'll look at it later this week.

The travis build failed before your pull request already, don't worry :)

lukaseder:
Thank you very much for your contribution. Indeed, this makes perfect sense.

PR #6399, opened 2017-07-11
Fix change in property key 

rajadilipkolli:
Corrected property name as per spring boot document https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#_customizing_jooq


lukaseder:
Thank you very much for your suggestion. Indeed, that dialect configuration was inherited from other spring examples, which may or may not read this property explicitly.

In this case, it makes more sense to follow the docs.

Issue #4627, opened 2015-10-11
Support Java 8 constructor parameter reflection in DefaultRecordMapper


rgoldberg:
Now that jOOQ 3.7+ requires Java 8, can you add a `RecordMapper` implementation that uses Java 8 constructor parameter reflection via `java.lang.reflect.Executable#getParameters()`?

----
See also:
https://groups.google.com/forum/#!topic/jooq-user/xQOloCcM2-4

lukaseder:
Thanks for your suggestion. Very interesting idea, indeed. This certainly makes sense!

eric239:
My plan is to modify `DefaultRecordMapper.ImmutablePOJOMapperWithConstructorProperties`, renaming it and making it work with either `ConstructorProperties` or `Parameter`. Makes sense?

lukaseder:
Thank you very much for your contribution. The implementation looks reasonable at first, but we must be careful about backwards incompatibilities. Right now, if someone relied on the mapping using the order of parameters, and suddenly, by accident, things are reordered because of typos in similar parameter names, then the result will be different. Example:

```sql
SELECT p2, p1 FROM table
```

Mapping this into:

```java
class MyPojo {
  MyPojo(int p1, int p2) { ... } 
}
```

We'll now get a regression, assuming that the naming was made this way on purpose.

The solution is to introduce a new `Settings.mapConstructorParameterNames` flag to opt in to this feature (we can turn this on by default in the next major release).

Also, I'll refactor your solution to make it Java 6 compatible (we still support that in the commercial distributions), and I might look into supporting also calling a constructor whose number of parameters doesn't match the number of fields, which might make a lot of sense, e.g. when doing:

```sql
SELECT id, first_name, last_name FROM customer
```

And then

```java
class Customer {
  // We don't care about the id
  Customer(String firstName, String lastName) { ... }
}
```

If there is only one constructor, then we can call that one in the presence of parameter names.

So, there are three remaining tasks:

- [ ] Add the new setting
- [ ] Make the code Java 6 compatible
- [ ] Support non-exact matches

I'll do the first two right away. Would you be interested in tackling the third?

eric239:
@lukaseder good points, I can take the third one. Should it (non-exact matches) remain "here" or become a separate issue?

lukaseder:
If by "here", you mean this issue, I'd prefer so, yes.

lukaseder:
[...]

PR #6796, opened 2017-11-08
[#4627] Support parameter info compiled into bytecode

eric239:


lukaseder:
Thank you very much for this suggested pull request. I mostly agree with it. Further discussion on the issue: https://github.com/jOOQ/jOOQ/issues/4627#issuecomment-342750456

Issue #6814, opened 2017-11-13
JDBCUtils.dialect(Connection) should recognise database version


lukaseder:
Currently, `JDBCUtils.dialect(Connection)` simply returns the family, not the dialect, which includes the database version. In the `Connection` overload, it should be possible to derive the version from the JDBC driver information.

----
This could be considered a somewhat breaking change, as clients might have relied on the fact that the family is returned, not the dialect

lukaseder:
Implemented the discussed changes. Thanks again for your contribution, @timur-sh !

timur-sh:
Thank you for your product too!

PR #6818, opened 2017-11-16
[#6814] overload the JDBCUtils.dialect method

timur-sh:
overload the JDBCUtils.dialect method to recognize and return an appropriate dialect

### DIFF
--- a/jOOQ/src/main/java/org/jooq/tools/jdbc/JDBCUtils.java
+++ b/jOOQ/src/main/java/org/jooq/tools/jdbc/JDBCUtils.java
@@ -40,6 +40,8 @@ import static org.jooq.SQLDialect.CUBRID;
 import static org.jooq.SQLDialect.DEFAULT;
 import static org.jooq.SQLDialect.DERBY;
 import static org.jooq.SQLDialect.FIREBIRD;
+import static org.jooq.SQLDialect.FIREBIRD_2_5;
+import static org.jooq.SQLDialect.FIREBIRD_3_0;
 import static org.jooq.SQLDialect.H2;
 // ...
 import static org.jooq.SQLDialect.HSQLDB;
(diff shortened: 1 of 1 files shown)


jOOQ/src/main/java/org/jooq/tools/jdbc/JDBCUtils.java
```diff
@@ -106,7 +114,9 @@ public static final SQLDialect dialect(Connection connection) {
 
 
                 String url = m.getURL();
-                result = dialect(url);
+                int majorVersion = m.getDatabaseMajorVersion();
+                int minorVersion = m.getDatabaseMinorVersion();
```
lukaseder: From experience with (and distrust of) JDBC drivers, I'll wrap these in a try-catch-ignore block :)

jOOQ/src/main/java/org/jooq/tools/jdbc/JDBCUtils.java
```diff
@@ -119,6 +129,66 @@ public static final SQLDialect dialect(Connection connection) {
...
+            case MYSQL:
+                return getMySqlDialect( majorVersion, minorVersion );
+            case FIREBIRD:
+                return getFirebirdDialect( majorVersion, minorVersion );
+        }
+
+        return dialect;
+    }
+
+    public static final SQLDialect getPostgreSqlDialect(int majorVersion, int minorVersion) {
```
lukaseder: I'd prefer these to be private until someone can show a use-case why they should be public...

jOOQ/src/main/java/org/jooq/tools/jdbc/JDBCUtils.java
```diff
@@ -119,6 +129,66 @@ public static final SQLDialect dialect(Connection connection) {
...
+            case 9:
+                if ( minorVersion >= 5 ) {
+                    return POSTGRES_9_5;
+                }
+                else if ( minorVersion == 4 ) {
+                    return POSTGRES_9_4;
+                }
+                else if ( minorVersion == 3 ) {
+                    return POSTGRES_9_3;
+                }
```
lukaseder: Hmm, another thing is that we probably want to return `POSTGRES_9_3` on all versions <= 9.3 as well, because `POSTGRES` (the dialect family) behaves like the latest version currently supported by jOOQ

jOOQ/src/main/java/org/jooq/tools/jdbc/JDBCUtils.java
timur-sh: Okay :) I'll take on mind this!

jOOQ/src/main/java/org/jooq/tools/jdbc/JDBCUtils.java
timur-sh: The same thing we should do for ```FIREBIRD``` doesn't it?

lukaseder: Sure, I already did: https://github.com/jOOQ/jOOQ/commit/236c65b26db95eeb4c8eaa218ba201991dc9ba7c

Issue #6838, opened 2017-11-22
Add org.jooq.User


lukaseder:
Create a new type `org.jooq.User` and allow for constructing it using `DSL.user(String)` and `DSL.user(Name)` overloads.

----
This is a prerequisite for #6812

PR #6853, opened 2017-11-23
[#6838] The User type for GRANT statement

timur-sh:


jOOQ/src/main/java/org/jooq/impl/UserImpl.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.jooq.User;
+
+/**
+ * A common implementation of the User type
+ *
+ * @author Timur Shaidullin
+ */
+public class UserImpl implements User {
+    private String name;
```
lukaseder: The internal representation of a `UserImpl` should be `Name`. The reason for this is the fact that `Name` encodes whether an identifier is quoted or not. In many databases, these are the same identifiers:

- `unquoted`
- `UNQUOTED`

But in most databases, these ones are different:

- `"quoted"` or `[quoted]` or `` `quoted` `` 
- `"QUOTED"` or `[QUOTED]` or `` `QUOTED` ``

jOOQ/src/main/java/org/jooq/impl/DSL.java
```diff
@@ -7852,6 +7853,18 @@ public static DropIndexOnStep dropIndexIfExists(Index index) {
...
+    // -------------------------------------------------------------------------
+    // XXX Access control
+    // -------------------------------------------------------------------------
+
+    public static User user(String name) {
+        return new UserImpl(name);
+    }
+
+    public static User user(Name name) {
+        return new UserImpl(name.toString());
```
lukaseder: This should be the main constructor. The one taking the `String` argument should be the convenience constructor, calling `user(name(name))`

jOOQ/src/main/java/org/jooq/impl/UserImpl.java
timur-sh: Sure, you're right.

jOOQ/src/main/java/org/jooq/User.java
```diff
@@ -9,5 +9,5 @@
     /**
      * The name of user
      */
-    public String getName();
+    public Name getName();
```
lukaseder: OK, I'll merge this but I'll change this back again. While the *internal* representation should be `Name`, the *external* one will be a `String` for API consistency reasons. Sorry... :)

lukaseder:
Thank you very much!

jOOQ/src/main/java/org/jooq/impl/UserImpl.java
```diff
@@ -8,14 +9,14 @@
  * @author Timur Shaidullin
  */
 public class UserImpl implements User {
```
lukaseder: All implementation classes need to be final and package-private (I'll fix this)

jOOQ/src/main/java/org/jooq/impl/UserImpl.java
```diff
@@ -8,14 +9,14 @@
...
-    private String name;
+    private Name name;
 
-    public UserImpl(String name) {
+    public UserImpl(Name name) {
```
lukaseder: Also, constructors are package-private, for internal implementations

jOOQ/src/main/java/org/jooq/User.java
timur-sh: Okay. Could you explain why  the external representation to be `String`, please.

jOOQ/src/main/java/org/jooq/impl/UserImpl.java
timur-sh: Thank you. I'll take this on mind in the future

jOOQ/src/main/java/org/jooq/User.java
lukaseder: For API consistency reasons with all the other API, e.g. `Table.getName()` or `Field.getName()`. Yes, it would be nice to return a `Name` reference here, but historically, we're returning strings...

timur-sh: I see, thanks!

PR #6872, opened 2017-11-28
[#4627] Support using first available constructor with **different** number of arguments as the last resort

eric239:


lukaseder:
Thanks for the suggestion. At first sight, this looks good. Will add more integration tests tomorrow to validate the implementation.

PR #6874, opened 2017-11-29
Added a common support of GRANT and REVOKE statements.

timur-sh:


lukaseder:
Thanks a lot for your suggestion. I will comment directly on the commits

jOOQ/src/main/java/org/jooq/Clause.java
```diff
@@ -48,6 +48,8 @@
 
     USER,
     ROLE,
+    PRIVILEGE,
+    GRANT,
```
lukaseder: The `GRANT` clause will need to be reviewed. `GRANT` should be the statement prefix, and then there will be clauses inside of it. Also it will need documentation. I'll take care of this after merging.

jOOQ/src/main/java/org/jooq/Grant.java
```diff
@@ -0,0 +1,53 @@
...
+ *
+ */
+package org.jooq;
+
+import java.util.Collection;
+
+/**
+ * @author Timur Shaidullin
+ */
+public interface Grant<R extends Record> extends QueryPart, Query {
```
lukaseder: Why is `Grant` generic? It doesn't really make sense to reference `<R>` as the grant statement doesn't necessarily work on record types.

jOOQ/src/main/java/org/jooq/Grant.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public Grant<R> grant(Privilege privilege);
+
+    public Grant<R> grant(Collection<? extends Privilege> privileges);
+
+    public Grant<R> on(Table<?> table);
+
+    public Grant<R> to(User user);
+
+    public Grant<R> to(Role role);
```
lukaseder: The design of this class is following a "builder pattern" / "fluent API", but not jOOQ's DSL API strategy. Please, take inspiration by the various other APIs in order to see how this should be done. Your version would allow bogus statements, like:

```java
grant(a).grant(b).on(t1).on(t2).to(a).grant(c).on(t3);
```

Here's a blog post showing how this API design principle works:
https://blog.jooq.org/2012/01/05/the-java-fluent-api-designer-crash-course/

jOOQ/src/main/java/org/jooq/impl/DSL.java
```diff
@@ -7858,6 +7860,22 @@ public static DropIndexOnStep dropIndexIfExists(Index index) {
     // XXX Access control
     // -------------------------------------------------------------------------
 
+    //todo add javadoc
+    public static Grant<Record> grant(String privilege) {
+        return using(new DefaultConfiguration()).grant(new PrivilegeImpl(name(privilege)));
```
lukaseder: For higher API cohesion, it is usually preferrable not to call this constructor here, but to reuse the static method that delegates to the constructor. This may lead to higher stack traces, but also to more quality and better code coverage by the integration tests.

jOOQ/src/main/java/org/jooq/impl/DSL.java
```diff
@@ -7858,6 +7860,22 @@ public static DropIndexOnStep dropIndexIfExists(Index index) {
...
+    }
+
+    //todo add javadoc
+    public static Grant<Record> grant(List<String> privileges) {
```
lukaseder: While the convenience method `grant(String privilege)` is fine, this method is not. If we have a `Privilege` type, then the collection overload must either:

- Accept the `Privilege` type through `Collection<? extends Privilege>`
- Accept any type through `Collection<?>`

I prefer the former. Note also the use of `Collection` rather than `List` in the API.

jOOQ/src/main/java/org/jooq/impl/DSL.java
```diff
@@ -7858,6 +7860,22 @@ public static DropIndexOnStep dropIndexIfExists(Index index) {
...
+        List<Privilege> collection = new ArrayList<>(privileges.size());
```
lukaseder: The diamond operator is not allowed as we still support Java 6

jOOQ/src/main/java/org/jooq/impl/DSL.java
```diff
@@ -7858,6 +7860,22 @@ public static DropIndexOnStep dropIndexIfExists(Index index) {
...
+
+        for (String privilege : privileges) {
+            collection.add(new PrivilegeImpl(name(privilege)));
```
lukaseder: Again, please don't call the constructor here, but the `DSL.privilege()` method

jOOQ/src/main/java/org/jooq/impl/DSL.java
lukaseder: ... note, it is correct in `DSLContext`

jOOQ/src/main/java/org/jooq/impl/GrantImpl.java
```diff
@@ -0,0 +1,142 @@
...
+    // ------------------------------------------------------------------------
+    // XXX: QueryPart API
+    // ------------------------------------------------------------------------
+
+    @Override
+    public void accept(Context<?> ctx) {
+        ctx.start(GRANT)
+            .visit(K_GRANT).sql(' ');
+
+        Privilege[] arrayOfPrivileges = privileges.toArray(new Privilege[privileges.size()]);
```
lukaseder: For consistency reasons, I'd prefer if the `toArray()` call would reference an empty `Privilege[]` that should be placed in `Tools`, just like `Tools.EMPTY_STRING`. (According to recent benchmarks, empty arrays outperform exactly sized arrays here).

jOOQ/src/main/java/org/jooq/impl/GrantImpl.java
```diff
@@ -0,0 +1,142 @@
...
+
+        for (int i = 0; i < arrayOfPrivileges.length; i++) {
+            ctx.visit(arrayOfPrivileges[i]);
+
+            if (i != arrayOfPrivileges.length - 1) {
+                ctx.sql(",");
```
lukaseder: Minor improvement: Use `','` (char) rather than `","` (String)

jOOQ/src/main/java/org/jooq/impl/GrantImpl.java
```diff
@@ -0,0 +1,142 @@
...
+        ctx.visit(K_ON).sql(' ')
+            .visit(table).sql(' ')
+            .visit(K_TO).sql(' ');
+
+        if (user != null) {
+            ctx.visit(user);
+        }
+
+        if (role != null) {
+            if (user != null) {
```
lukaseder: I prefer we make this mutually exclusive for now

jOOQ/src/main/java/org/jooq/impl/PrivilegeImpl.java
```diff
@@ -0,0 +1,82 @@
...
+final class PrivilegeImpl extends AbstractQueryPart implements Privilege {
+
+    /**
+     * Generated UID
+     */
+    private static final long     serialVersionUID = -3106268610481536038L;
+    private static final Clause[] CLAUSES          = { PRIVILEGE };
+    private Name                  name;
+
+    PrivilegeImpl(Name name) {
```
lukaseder: I think that the privilege shouldn't be modelled by a `Name` but by a `Keyword` instead. For instance, I doubt that `GRANT "SELECT" ...` is possible.

jOOQ/src/main/java/org/jooq/impl/PrivilegeImpl.java
```diff
@@ -0,0 +1,82 @@
...
+ * @author Timur Shaidullin
+ */
+final class PrivilegeImpl extends AbstractQueryPart implements Privilege {
+
+    /**
+     * Generated UID
+     */
+    private static final long     serialVersionUID = -3106268610481536038L;
+    private static final Clause[] CLAUSES          = { PRIVILEGE };
+    private Name                  name;
```
lukaseder: This should be final

lukaseder:
I see that you've already applied 1-2 fixes in your second commit. May I ask you to create a fresh pull request with only a single, "squashed" commit instead? It will be much easier to review.

jOOQ/src/main/java/org/jooq/impl/PrivilegeImpl.java
```diff
@@ -63,7 +63,10 @@
 
     @Override
     public void accept(Context<?> ctx) {
-        ctx.visit(name);
+        ctx
+            .quote(false)
```
lukaseder: This will not be required once `Name` is replaced by `Keyword`

jOOQ/src/main/java/org/jooq/impl/PrivilegeImpl.java
timur-sh: Yes, ` I doubt that GRANT "SELECT" ` is not possible. I change it

jOOQ/src/main/java/org/jooq/Grant.java
timur-sh: I fixed it in the second commit.

jOOQ/src/main/java/org/jooq/Clause.java
lukaseder: Yes

timur-sh:
I don't understand this: `Implement the command as a DSL, not a builder`. Could you get a bit more explanation?

jOOQ/src/main/java/org/jooq/Grant.java
timur-sh: Thank you for an article. It's the awesome way to build DSL API strategy

lukaseder:
> I don't understand this: Implement the command as a DSL, not a builder. Could you give a bit more explanation?

Look around you :) The entire jOOQ DSL API is built quite differently from what you suggested... For example, look at the various `Select*Step` types.

lukaseder:
This PR has been replaced by https://github.com/jOOQ/jOOQ/pull/6878

Issue #6812, opened 2017-11-09
Add support for GRANT and REVOKE statements


lukaseder:
Many databases have some way of granting (and revoking) object rights to users / roles. It would be useful to have these statements as well in our repertoire.

We'll probably need a new top level type `org.jooq.User`, and perhaps `org.jooq.Role`. It’s currently unclear how these two types should relate to each other. Once we have the types, we’ll need:

- [x] Ways to construct them using DSL.user() and DSL.role()
- [x] ~Perhaps, code generation support~ (separate issue: see #6902)
- [x] The actual API and its implementation
- [x] Parser support (#6485) and documentation in grammar

Additional improvements in separate issues:

- [x] Granting to and revoking from `PUBLIC`: #6911
- [x] `GRANT .. [ WITH GRANT OPTION ]`: #6913
- [x] `REVOKE [ GRANT OPTION FOR ] ..`: #6931

We should collect all the related features / syntaxes from different vendors – and especially the little quirks that each has to them. This subset of vendors should be a good starter:

- [x] ~DB2 LUW~ (We'll do this later)
- [x] Derby
- [x] H2
- [x] HSQLDB
- [x] MySQL / MariaDB
- [x] Oracle
- [x] PostgreSQL
- [x] SQL Server
- [x] ~SQLite~ (does not support grants)

timur-sh:
There is a common denominator to use GRANT statement for above database. Except is SQLite, it doesn't support any accession control

It is separated on two possibles:
First one is `GRANT PRIVILEGES`
`GRANT <privilege> ON <objectName> TO <grantees> [<PRIVILEGE OPTIONS>]`
`<privilege>`  = ACTION [, ...] - this is an action specific to particular database. You can find whole list below  
`<objectName>` - OBJECT this can be table, schema, view, etc. You can find whole list below  
`<PRIVILEGE OPTIONS>` -  PRIVILEGE_OPTIONS options that may be depend on certain database. You can find them below

The second one is `GRANT ROLES`
`GRANT roleName TO <grantees> [<ROLE_OPTIONS>]`
`<ROLE_OPTIONS>` -  ROLE_OPTIONS options that may be depend on certain database. You can find them below
`<grantees>` = user | role | PUBLIC 

> `<grantees>`  may be changed by specific database

**DB2 LUW**

**Derby** https://db.apache.org/derby/docs/10.5/ref/rrefsqljgrant.html
[...]

PR #6878, opened 2017-11-29
[#6812] Added GRANT statement and its implementation.

timur-sh:


lukaseder:
Thank you very much for your contribution. That looks workable, thus far. I will fix a few minor things that are easier to just fix rather than discuss. Will commit soon and comment on the fixes thereafter.

timur-sh:
Thank you!

lukaseder:
Here are some changes that I've done:

- `DSL.privilege(Keyword)` is package-private for now. I'm not 100% convinced that privilege : keyword is a 1:1 relationship. There are probably more complex privileges, which cannot be modelled by a keyword alone. However, privileges can always be expressed by strings
- Moved methods around in `DSLContext`
- Added overload `DSLContext.grant(Privilege...)` for API consistency reasons. The same with `revoke()`. I'm leaving `grant(Privilege)` (and `revoke()`) as I'm assuming there are dialects that allow for only granting/revoking a single privilege at a time.
- Changed `DSL.grant()` and `DSL.revoke()` signatures to be exactly the same as those in `DSLContext`
- Added `execute()` calls to Javadoc (people were often confused when that call was lacking, in the past)
- Changed interface names, e.g. `GrantOnStep` (`Step` is always the suffix)
- Removed unused interfaces `GrantFirstStep` and `RevokeFirstStep`
- Added `GrantFinalStep` and `RevokeFinalStep` types, which are there for historic reasons and for edge cases in backwards compatible API evolution
- RevokeFromStep shouldn't extend Query, otherwise the from() clause would be optional. Same with Grant
- RevokeOnStep shouldn't extend RevokeFromStep, otherwise the on() clause would be optional. Same with Grant
- Added a third overload for on(Table) methods: on(Name)
- Annotated plain SQL API with @PlainSQL and the relevant Javadoc disclaimer
- Added the @Support annotation throughout the DSL
- Made all impl methods final
- Fixed clauses

**TODOs**

- I'm still not happy with the fact that we currently have a String as the privilege. There should be something better for the basic privileges
- I've run a quick integration test on Oracle, which works (see code below). Will do more testing next week.

```java
    public void testGrantRevoke() throws Exception {
        Table<?> t = table(name(schema().getName(), "t"));
        Field<Integer> i = field(name("i"), SQLDataType.INTEGER);
        DSLContext norights = using(getConnectionNoRights());
        User user = user(getConnectionNoRights().getMetaData().getUserName());

        try {
            create().createTable(t).columns(i).execute();
            assertEquals(0, create().fetchCount(t));
            assertThrows(DataAccessException.class, () -> norights.fetchCount(t));

            create().grant(privilege("select")).on(t).to(user).execute();
            assertEquals(0, norights.fetchCount(t));

            create().revoke(privilege("select")).on(t).from(user).execute();
            assertThrows(DataAccessException.class, () -> norights.fetchCount(t));
        }
        finally {
            ignoreThrows(create().dropTable(t)::execute);
        }
    }
```

lukaseder:
Anyway, this works nicely so far. Well done! Further discussion / improvements will be made on the main ticket #6812

timur-sh:
I've reviewed your aforementioned enhancements. It makes the API more readable and better. Thank you. About the `Privilege` we should improve it, such as we see  #6812, there are a lot of variate privileges

Issue #6867, opened 2017-11-28
Add DSLContext.ddl(Table[]) and ddl(Collection<? extends Table<?>>)


lukaseder:
Currently, there's only a possibility to export DDL for a single table, or for an entire schema. Sometimes, we want to export several table's DDL, though.

PR #6885, opened 2017-12-05
[#6867] Added a possibility to export several DDL of tables

timur-sh:


lukaseder:
Thank you very much for your suggestion. I will look into this during the next days.

timur-sh:
Thank you!

lukaseder:
Looks good, thanks again. I will merge and make minor adaptations on my side.

jOOQ/src/main/java/org/jooq/DSLContext.java
```diff
@@ -8362,6 +8362,62 @@
      */
     Queries ddl(Table<?> table, DDLFlag... flags);
 
+    /**
+     * Generate the complete creation script for tables.
+     *
+     * @see #ddl(Table[], DDLFlag...)
+     */
+    Queries ddl(Table[] tables);
```
lukaseder: This will get a wildcard: `Table<?>` to avoid the raw type

jOOQ/src/main/java/org/jooq/impl/DDL.java
```diff
@@ -78,6 +78,10 @@
     }
 
     final Queries queries(Table<?> table) {
```
lukaseder: I guess, the single table method can be removed from the internal API to remove redundancy a little bit.

jOOQ/src/main/java/org/jooq/impl/DDL.java
```diff
@@ -96,11 +100,19 @@ final Queries queries(Table<?> table) {
...
-               .columns(table.fields())
-               .constraints(constraints)
-        );
+        return ctx.createTable(table)
+            .columns(table.fields())
+            .constraints(constraints);
+    }
+
+    final Queries queries(Table[] tables) {
+        List<Query> queries = new ArrayList<Query>();
```
lukaseder: Local variable can be a `Query[]`

jOOQ/src/main/java/org/jooq/DSLContext.java
lukaseder: Also, the version of this API that doesn't specify any `DDLFlag` can profit from varargs

jOOQ/src/main/java/org/jooq/impl/DefaultDSLContext.java
```diff
@@ -2817,6 +2818,26 @@ public Queries ddl(Table<?> table, DDLFlag... flags) {
         return new DDL(this, flags).queries(table);
     }
 
+    @Override
+    public Queries ddl(Table[] tables) {
```
lukaseder: In order to improve cohesion, the existing methods will call the new methods rather than passing responsibility directly to `DDL`

Issue #6736, opened 2017-10-24
Add support for PostgreSQL ON CONFLICT .. WHERE <index_predicate>


lukaseder:
We already support `ON CONFLICT .. DO UPDATE .. WHERE` as of #5637, but there's another possible `WHERE` clause that further specifies the `ON CONFLICT` clause:

> where conflict_target can be one of:
> 
>     ( { index_column_name | ( index_expression ) } 
>           [ COLLATE collation ] [ opclass ] [, ...] ) [ WHERE index_predicate ]

----
See also:
https://www.postgresql.org/docs/current/static/sql-insert.html

timur-sh:
There is missed there `ON CONSTRAINT constraint_name` clause. It provides useful functional allowing to specify constraint by name. That's why it should be added.

lukaseder:
Thanks for your message, @timur-sh. I've created a separate feature request for this: #6906

dzagorovsky:
@lukaseder do you have any plans to support PostgreSQL ```ON CONFLICT ...  WHERE ... DO UPDTAE```?
Maybe there is some non-plain sql workaround?

lukaseder:
Yes, the plan is this issue. No, there's no non-plain SQL workaround.

DaanVandenBosch:
As far as I know, the only workaround at the moment is reverting to dslContext.query("sql string"). Is this correct?

And @lukaseder, I'd like to thank you and your team for building and open sourcing such a wonderful tool! I'm so glad I've finally found a more powerful alternative to an ORM. Working with JOOQ code is so much more pleasant than a tangle of Spring Data, Hibernate Specifications, JPQL, HQL and the unavoidable raw SQL or dropping down to JDBC. I appreciate the Kotlin/Scala support too (or at least it plays nice with other languages).

lukaseder:
Thanks for your nice words, @DaanVandenBosch.

> As far as I know, the only workaround at the moment is reverting to dslContext.query("sql string"). Is this correct?

This is one workaround. Another is to patch the generated SQL with an `ExecuteListener`. A third is, of course, to patch the library itself :-)

DaanVandenBosch:
[...]

Issue #2908, opened 2013-12-20
Add support for COLLATIONS


lukaseder:
Some database(s) support a `COLLATE` clause to define the character set to be used for string comparison. For example:

``` sql
INNER JOIN T2 ON T1.COL1 = T2.COL1 COLLATE utf8_bin
```

It makes sense to support this natively in jOOQ:

- [x] Add a `org.jooq.Collation` type and a `DSL.collation(String)` constructor: #7120
- [x] Add `Field.collate(Collation)` and `Field.collate(String)`: #7116
- [x] Add `DataType.collate(Collation)`: #7117
- [x] Add `CREATE TABLE` and `ALTER TABLE` support for explicit column collations: #7123
- [x] ~Add `CREATE TABLE` and `ALTER TABLE` support for default table collation: #7118~
- [x] ~Add code generator support: #7121~: Moved to #2908
- [x] ~Add collation information to `InformationSchema` (jooq-meta.xsd): #7122~: Moved to #2908
- [x] ~Add `SET COLLATION` statement support: #7119~: Moved to #2908
- [x] Add parser support for all of the above: #6485 (operator precedence!)
- [x] Update manual

This feature has been seen to be supported by (check = jOOQ supports it for the database):

- [x] HSQLDB
- [x] MariaDB
- [x] MySQL
- [x] Oracle
- [x] PostgreSQL
- [x] SQL Server
- [x] SQLite

----
See also:
https://groups.google.com/d/msg/jooq-user/W_LKypJrI_k/k8zLhvwhhEoJ

schaffe:
+1

cbossi:
Hi Lukas,
is this still intended to be implemented? If yes, is there a planned date/release?

lukaseder:
> is this still intended to be implemented?

Everything that is not closed might be implemented at some point if there's enough traction

> If yes, is there a planned date/release?

Not yet. But you can easily work around the issue:

```java
// Beware of SQL injection risks!
static <T> Field<T> collate(Field<T> field, String collation) {
    return DSL.field("{0} collate {1}", field.getDataType(), field, DSL.sql(collation));}
```

cbossi:
Thanks, looks like a nifty workaround

lukaseder:
@cbossi: It'll always be your workaround of choice, once a feature is not (yet) available in jOOQ.

lukaseder:
OK, this will make it in jOOQ 3.11. See the updated issue description for a list of subtasks

PR #6910, opened 2017-12-13
[#6736] Add support for PostgreSQL ON CONFLICT conflict_target WHERE  ..

timur-sh:


timur-sh:
To semantic reason I've used a `Keyword` type to distinguish an operator class. I don't have a pleasure on it. It should be changed in the future  on more suitable type.

lukaseder:
Thanks for your suggestion. I will reject this PR in its entirety for the following reasons:

- The DSL API makes use of terminology that does not appear in SQL syntax, including:
  o `addOnConflict`. There's no `ADD` keyword
  o `whereIndex`. There's no `INDEX` keyword
- You hijacked feature request #6736 in order to add tons of additional features which we hadn't discussed first (some of which I prefer not to support):
  o The collation. We currently do not support collations or `collate` clauses, see #2908. There are many other places where collation support would be much more useful rather than in this arcane PostgreSQL-specific statement. Even if we did add support for the `collate` clause, it would have to be added using a `collate()` method, and a new `org.jooq.Collation` type (which we don't have right now)
  o The opclass. I really don't see the point of supporting this at this stage. Your (ab)using the `Keyword` type for this in the public API should be an indication that this addition was premature.
- The new public interface `OnConflictProvider` shouldn't be part of the public API. I doubt it should even be there at all.

If you want to try again, please restrict this particular change to add only the requested `WHERE` clause.

timur-sh:
Thank you for your explanation. I'd like to complete this feature. To do it there are a several milestones:

Some points need to be changed:
-  `InserImpl#onConflict` can take a 'OnConflict` type

Some points need to be deleted
- The `InsertOnConflictStep` interface 
- The `collation` implementations
- The `opclass` implementations
- All `addOnConflict` methods
- The `OnConflictProvider` interface

But, I'm not sure what to do with `whereIndex` and `addIndexConditions`, because, `InsertIml` implements `where` methods.

Could you give any advices also?

lukaseder:
> InserImpl#onConflict can take a 'OnConflict` type

I don't think that's the appropriate approach. Let's look at the grammar:

```
[ ON CONFLICT [ conflict_target ] conflict_action ]
where conflict_target can be one of:

    ( { index_column_name | ( index_expression ) } [ COLLATE collation ] [ opclass ] [, ...] ) [ WHERE index_predicate ]
    ON CONSTRAINT constraint_name
```

Now, the jOOQ way of translating this sort of grammar is to offer the following "overloads":

```
// index_column_name version
onConflict(Field<?>...)
onConflict(Collection<? extends Field<?>>)

// ON CONSTRAINT version
onConflictOnConstraint(Constraint)
```

> But, I'm not sure what to do with whereIndex and addIndexConditions, because, InsertIml implements where methods.

The fact that there is already a `where()` method in the implementation shouldn't leak into the API. There are similar cases in other implementations, where a `where()` method has several meanings. Have a look at `MergeImpl.where()`.

Since the current implementations of the DSL are mutable, we can easily "remember" what the last clause was, and then act accordingly. If the DSL were immutable, it would be even easier to implement.

Hope this helps

timur-sh:
> // index_column_name version
onConflict(Field<?>...)
onConflict(Collection<? extends Field<?>>)

As you know, this works as `index_expression` version also.

lukaseder:
> As you know, this works as index_expression version also.

Yes, that's probably not fully supported yet.

> When will be implemented collation and opclass clauses, it is be appropriate to use a type like OnConflict to group them.

No, why?

```java
create.insertInto(table, ...)
      ...
      .onConflict(fields)
      .collate(collation)
      .varcharPatternOps()
      ...
```

`COLLATE` is a keyword, so this translates to a method. The collation is an object reference, so this translates to a parameter. `opclass` is a placeholder for a set of possible keywords, which we can either hard-code in the API if the set is relatively small, or we add a synthetic keyword `OPCLASS`, which translates to a method `opclass(opclass)`.

There is really no need to create any wrapper around a set of parameters in the jOOQ DSL. Any such wrapper will be an impediment to learning the DSL API, which should map to the backing SQL syntax in the closest possible way as illustrated in this article that I've linked earlier:
https://blog.jooq.org/2012/01/05/the-java-fluent-api-designer-crash-course

Let's keep it simple (for the user) :-)

Hope this helps.

(And really, I'd love to see the use-case and at least 10 people asking for the feature before we'll add support for this arcane opclass thingy. We don't have to add support for every wacko vendor-specific feature (I really really wish PostgreSQL had supported SQL standard MERGE, rather than this weird `ON CONFLICT` nonsense))

Issue #6906, opened 2017-12-11
Add support for PostgreSQL ON CONFLICT .. ON CONSTRAINT ..


lukaseder:
We already support `ON CONFLICT .. ` , but there's another possible clause that further specifies the `ON CONFLICT` clause:

> where conflict_target can be one of:
> 
>     ( ... )
>     ON CONSTRAINT *constraint_name*

----
See also:
https://www.postgresql.org/docs/current/static/sql-insert.html

timur-sh:
@lukaseder I'd like to pick it up. I will add a new `onConflictOnConstraint(name)` method. Should the `name`' argument to be a new `Constraint` type, or the `Keyword` is appropriate for this ?

lukaseder:
We already have a type for this. In fact, we have two types and both should be supported by overload:

- `org.jooq.Constraint` (this is used in DDL statements)
- `org.jooq.UniqueKey` (this is generated by the code generator)

In general, `Keyword` is almost always the wrong choice of type for things that are names or references. I mean, it corresponds to keywords like `SELECT` or `AS`, etc., not to names.

lukaseder:
Thanks again for your contribution. This is now implemented for jOOQ 3.11

lukaseder:
Added parser support as well: https://github.com/jOOQ/jOOQ/commit/fe34d9a84c26701763c0b2099445eb5a5d89fdd0

PR #6919, opened 2017-12-14
[#6906] Added support for PostgreSQL ON CONFLICT .. ON CONSTRAINT ..

timur-sh:


lukaseder:
Thanks for your suggestion. I'll review this soon.

jOOQ/src/main/java/org/jooq/impl/InsertImpl.java
```diff
@@ -604,6 +609,27 @@ public final InsertImpl doNothing() {
...
+    @Override
+    public final InsertImpl onConflictOnConstraint(Name constraint) {
+        onConflictOnConstraint(constraint(constraint));
+        return this;
+    }
+
+    @Override
+    public final InsertImpl onConflictOnConstraint(UniqueKey<R> constraint) {
+        if (constraint.getName() == null)
+            throw new IllegalStateException("UniqueKey's name is not specified");
```
lukaseder: Good idea. I'll switch this to `StringUtils.isEmpty()` (as empty names are possible too, and are also illegal here). Also, `IllegalArgumentException` seems more appropriate in this particular case

lukaseder: ... I've noticed just now that the exception is also thrown from `InsertQueryImpl` (which makes sense, of course). So, to remove repetition of code, I will change the above `InsertImpl` logic to simply delegate to `InsertQueryImpl` and let the latter handle the plumbing / data type conversion.

jOOQ/src/main/java/org/jooq/impl/InsertQueryImpl.java
```diff
@@ -248,30 +273,41 @@ final void accept0(Context<?> ctx) {
...
-                       .start(INSERT_ON_DUPLICATE_KEY_UPDATE)
-                       .visit(K_ON_CONFLICT)
-                       .sql(" (");
+                       .start(INSERT_ON_DUPLICATE_KEY_UPDATE);
 
-                    if (onConflict != null && onConflict.size() > 0) {
-                        boolean qualify = ctx.qualify();
+                    if (onConstraint != null) {
+                        ctx.data(DATA_CONSTRAINT_REFERENCE, true);
+                        ctx.visit(K_ON_CONSTRAINT)
```
lukaseder: I suspect the `ON CONSTRAINT` keywords here should be prefixed by `ON CONFLICT`. The current implementation seems to produce a syntax error. I'll fix this when implementing the relevant integration tests.

lukaseder:
Cool, no regressions in the integration tests... Now adding integration tests for the new functionality

jOOQ/src/main/java/org/jooq/impl/InsertQueryImpl.java
```diff
@@ -335,18 +371,30 @@ else if (onDuplicateKeyIgnore) {
...
                     ctx.formatSeparator()
-                       .start(INSERT_ON_DUPLICATE_KEY_UPDATE)
-                       .visit(K_ON_CONFLICT)
-                       .sql(' ');
+                       .start(INSERT_ON_DUPLICATE_KEY_UPDATE);
 
-                    if (onConflict != null && onConflict.size() > 0) {
-                        boolean qualify = ctx.qualify();
+                    if (onConstraint != null) {
+                        ctx.data(DATA_CONSTRAINT_REFERENCE, true);
```
lukaseder: Nice one. Curious about this, how did you discover this flag?

timur-sh: I've investigated `ConstraintImpl#accept` method. As I see, it's perfectly appropriate in use cases, when it's need to get only `name` of `Constraint`'s object by `Context`

lukaseder: Yeah, it's an "acceptable hack" - a workaround if either the rendering API or the data model isn't sophisticated enough to communicate such a "special case" and the special case isn't worth an API change...

jOOQ/src/main/java/org/jooq/impl/InsertImpl.java
timur-sh: I agree, using `StringUtils.isEmpty()` in this case is more appropriate then checking on `null`

PR #6921, opened 2017-12-14
Optimization of {ULong,UInteger,UShort}.toBigInteger

zhurs:
Now, UNumber.toBigInteger implemented as new BigInteger(toString());
BigInteger is immutable, so we can just return this.value for ULong, and use BigInteger.valueOf for other types

### DIFF
--- a/jOOQ/src/main/java/org/jooq/types/UInteger.java
+++ b/jOOQ/src/main/java/org/jooq/types/UInteger.java
@@ -16,6 +16,7 @@
 package org.jooq.types;
 
 import java.io.ObjectStreamException;
+import java.math.BigInteger;
 
 /**
  * The <code>unsigned int</code> type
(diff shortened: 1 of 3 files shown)


lukaseder:
Thank you very much for your suggestion. In fact, the original jOOU library already changed the implementation of `ULong` to be more efficient (being based on a `long` rather than a `BigInteger`). I still agree with this improvement though:
https://github.com/jOOQ/jOOQ/pull/6921/commits/e30299beb087b75e90a8e5fae3ccf5da288056bb

Would you be interested in contributing this second change to jOOU instead? https://github.com/jOOQ/jOOU

I'm about to release a new version of jOOU, so this change might still make it.

lukaseder:
Marking this as a duplicate of https://github.com/jOOQ/jOOQ/issues/6922, as we're going to upgrade the jOOU dependency with this fix in it.

lukaseder:
I can actually merge this here as well to give you credit

lukaseder:
This contribution does not require signing the [transfer of rights agreement](https://github.com/jOOQ/jOOQ/blob/master/CONTRIBUTING.md), as it modifies jOOU, which is not dual licensed, but ASL 2.0 licensed only.

lukaseder:
For the record, I've already merged this change also into jOOU: https://github.com/jOOQ/jOOU/issues/14

Thanks again for your contribution!

zhurs:
Thank you!

Issue #6913, opened 2017-12-13
Add support for WITH GRANT OPTION clause to GRANT statement


lukaseder:
This is a minor addition to #6812

timur-sh:
I've completed my investigation what databases are support `WITH GRANT OPTIONS` clause. This is summarize in the list below:
1. **HSQLDB** http://hsqldb.org/doc/guide/accesscontrol-chapt.html 
2. **Firebird** https://firebirdsql.org/refdocs/langrefupd25-security-grant-revoke.html. This database supports `WITH {GRANT|ADMIN} OPTION`. For now we omit `ADMIN` clause, and will use rest of the clause `WITH GRANT OPTION`
3. **MySQL / MariaDB** https://dev.mysql.com/doc/refman/8.0/en/grant.html / https://mariadb.com/kb/en/library/grant/
4. **PostgreSQL** https://www.postgresql.org/docs/10/static/sql-grant.html 

Thus the [previous grammar form](https://github.com/jOOQ/jOOQ/issues/6812#issuecomment-346331783) will be extended to
```
grantStatement ::= GRANT <privilege> ON <table> TO { <user> | <role> } [ WITH GRANT OPTION ]
```

I plan to add a method `withGrantOption()`. 

Must be added a `GRANT_WITH_GRANT_OPTION` to `Clause` class?

lukaseder:
Thank you very much for your investigation. Yes, indeed, it's a standard feature:

![image](https://user-images.githubusercontent.com/734593/34149406-3e0eb09e-e4a5-11e7-9672-b6874260f145.png)

It's also supported by:

- DB2: https://www.ibm.com/support/knowledgecenter/en/SSEPEK_11.0.0/sqlref/src/tpc/db2z_sql_granttableorviewprivileges.html
- Oracle: https://docs.oracle.com/database/121/SQLRF/statements_9014.htm#SQLRF01603
- SQL Server: https://docs.microsoft.com/en-us/sql/t-sql/statements/grant-transact-sql

And probably others.

Indeed, there should be a `withGrantOption()` method on a new optional DSL "step" type that extends (and replaces references to) `GrantFinalStep`

lukaseder:
Note that for completeness' sake, we should also allow for revoking this option: #6931

timur-sh:
I could not add support for DB2, Oracle and other dialects, because them are not open source.

timur-sh:
Could you say, where can I get a SQL standard specification 2016?

[...]

PR #6934, opened 2017-12-19
[#6913] Added a support for GRANT .. WITH GRANT OPTION

timur-sh:


lukaseder:
Fantastic, thank you very much! I'll review and merge immediately!

jOOQ/src/main/java/org/jooq/GrantGrantedStep.java
```diff
@@ -0,0 +1,54 @@
...
+import static org.jooq.SQLDialect.MYSQL;
+import static org.jooq.SQLDialect.POSTGRES;
+
+/**
+ * The step of creation <code>GRANT</code> statement that assign user
+ * or role a grant option.
+ *
+ * @author Timur Shaidullin
+ */
+public interface GrantGrantedStep extends GrantFinalStep{
```
lukaseder: Usually, the naming scheme is indicating what's possible in the current step, not what has been done in the previous step(s) as that could be ambiguous. So, a good name would be `GrantWithGrantOptionStep` or `GrantGrantOptionStep`. I'll fix this after merging.

jOOQ/src/main/java/org/jooq/GrantGrantedStep.java
```diff
@@ -0,0 +1,54 @@
...
+    @Support({ POSTGRES, MYSQL, HSQLDB, FIREBIRD })
```
lukaseder: Just to comment, I'll remove the individual dialect references until we cover the functionality in our integration tests. While it is quite probable that this is correct, we want to *formally* support the feature by guaranteeing that it works...

timur-sh: Do you plan make an integration test open in the future? This step a bit decrease your workload, if users write integration test by themselves.

jOOQ/src/main/java/org/jooq/GrantGrantedStep.java
timur-sh: Name `GrantGrantOptionStep` is more suitable for this case, it lets to add `[ GRANTED BY <grantor> ]` in the future.

jOOQ/src/main/java/org/jooq/GrantGrantedStep.java
lukaseder: No, the integration tests are closed source and available to our paying customers only. We follow a similar dual-licensing business model as SQLite, which also does not open source all the tests.

jOOQ/src/main/java/org/jooq/GrantGrantedStep.java
lukaseder: Yeah, perhaps, although I'd like to wait for someone to actually request the `GRANTED BY <grantor>` feature before implementing that...

PR #6940, opened 2017-12-20
[#6913] Added parser support for WITH GRANT OPTION clause

timur-sh:


lukaseder:
Thanks a lot for your contribution!

jOOQ/src/main/java/org/jooq/impl/ParserImpl.java
```diff
@@ -1313,7 +1314,14 @@ private static final DDLQuery parseGrant(ParserContext ctx) {
 
         GrantOnStep s1 = privileges == null ? ctx.dsl.grant(privilege) : ctx.dsl.grant(privileges);
         GrantToStep s2 = s1.on(table);
-        return user == null ? s2.toPublic() : s2.to(user);
+        GrantWithGrantOptionStep s3 = user == null ? s2.toPublic() : s2.to(user);
+
+        if (peekKeyword(ctx, "WITH GRANT OPTION")) {
+            parseKeyword(ctx, "WITH GRANT OPTION");
```
lukaseder: Notice that there's a `parseKeywordIf()` method, that does these two things together.

jOOQ/src/main/java/org/jooq/impl/ParserImpl.java
```diff
@@ -1313,7 +1314,14 @@ private static final DDLQuery parseGrant(ParserContext ctx) {
...
+            s3.withGrantOption();
```
lukaseder: While this currently works (because the DSL API is mutable), it's generally a good idea to pretend the DSL API is already immutable (we'll change that in jOOQ 4.0), and thus assign the result of this to a new type.

We can't get this right all the time because it still works the way you wrote it, but it's a good thing to keep in mind.

jOOQ/src/main/java/org/jooq/impl/ParserImpl.java
timur-sh: I understand. Thank you!

jOOQ/src/main/java/org/jooq/impl/ParserImpl.java
timur-sh: Thank you for explanation, I understood it, but I cannot imagine how it will work when be an immutable. Could you give a link where I can take more information about DSL immutable feature?

lukaseder: Cannot seem to find the issue number... Just imagine it will be:

```java
GrantWithGrantOptionStep s3 = ...
assertFalse(s3 == s3.withGrantOption());
```

Because the implementation will not modify its properties but copy them into a new implementation instance.

Issue #6931, opened 2017-12-19
Add support for REVOKE GRANT OPTION FOR


lukaseder:
This is a minor addition to #6812, the inverse operation of #6913 (`GRANT .. WITH GRANT OPTION`).

The feature will be implemented with `DSL[Context].revokeGrantOptionFor(...)` methods. There's no additional `Step` type needed in the DSL API.

lukaseder:
So this would then also need parser support. If you want, you can also implement parser support in a single commit referencing both "[#6913] [#6931]" (funny how these two issues have almost the same issue number...)

timur-sh:
Yes, I'll pick up this one too.

PR #6941, opened 2017-12-20
[#6931] Added parser support for GRANT OPTION FOR clause

timur-sh:


lukaseder:
Looks perfect. Thanks very much!

Issue #6772, opened 2017-11-01
Add a <jpaVersion/> code generator configuration


lukaseder:
When we generate JPA annotations, we must be careful to respect the JPA version. The recent introduction of the `@Index` annotation that was generated with #6323 broke some assumptions by users who were restricting their generated code to use JPA 2.0 only (or less).

We need a new flag that allow for restricting the JPA version to be considered with `<jpaAnnotations/>`. If no such flags are specified, then the latest JPA version is assumed (just like the `SQLDialect` family / dialect duality).

The new flag is called `<jpaVersion>2.0</jpaVersion>`.

timur-sh:
@lukaseder I'll pick it up.

lukaseder:
Thanks again for your contribution! This is now in jOOQ 3.11

PR #6955, opened 2017-12-27
[#6772] Added a new flag `jpaVersion` to code generator.

timur-sh:


lukaseder:
Thank you very much for your suggestion. I'll review shortly.

jOOQ-meta/src/main/resources/xsd/jooq-codegen-3.11.0.xsd
```diff
@@ -1033,6 +1037,15 @@ This flag is ignored in the commercial Java 6 distribution of jOOQ 3.9+ ]]></jxb
     </all>
   </complexType>
 
+
+  <simpleType name="JpaVersion">
+    <restriction base="string">
+      <enumeration value="V_2_0"/>
+      <enumeration value="V_2_1"/>
+      <enumeration value="V_2_2"/>
```
lukaseder: I understand that you've introduced this enumeration syntax for XJC to pick up the values as actual enum constants, which wouldn't be possible if we were using `2.0`, `2.1`, and `2.2` instead. But still, the configuration should be XML / declarative first, and programmatic second, so I'd really prefer this to be simple strings that correspond to the JPA version scheme on a 1:1 basis.

jOOQ-meta/src/main/java/org/jooq/util/jaxb/JpaVersion.java
```diff
@@ -0,0 +1,102 @@
...
+
+    public boolean supportedIndex() {
+        return annotations.contains("Index");
+    }
+
+    public String getName() {
+        return name().substring(2).replace('_', '.');
+    }
+
+    public static JpaVersion latest() {
```
lukaseder: These classes are generated by XJC. It's not possible (or better: not recommended) to put custom logic inside of them. We can easily create a utility though...

jOOQ-manual/src/main/resources/org/jooq/web/manual-3.11.xml
```diff
@@ -16931,6 +16931,7 @@ implements IBook {
 								<li><strong>daos</strong>: Records are a pre-requisite for DAOs. If DAOs are generated, records are generated as well</li>
 								<li><strong>interfaces</strong>: If interfaces are generated, records will implement them</li>
 								<li><strong>jpaAnnotations</strong>: JPA annotations are used on generated records</li>
+								<li><strong>jpaVersion</strong>: Version of JPA specification is to be used to generate version-specific annotations. If it is omitted, the latest version is used by default.</li>
```
lukaseder: Cool, thanks for thinking of this!

lukaseder:
Thanks again for your contribution. I have merged this and will adapt the implementation for the enumeration not to use a custom versioning scheme.

jOOQ-codegen/src/main/java/org/jooq/util/JavaGenerator.java
```diff
@@ -278,6 +279,9 @@ public final void generate(Database db) {
             + ((!generateGeneratedAnnotation && (useSchemaVersionProvider || useCatalogVersionProvider)) ?
                 " (forced to true because of <schemaVersionProvider/> or <catalogVersionProvider/>)" : ""));
         log.info("  JPA annotations", generateJPAAnnotations());
+        log.info("  JPA version", generateJpaVersion()
+            + (!generateJPAAnnotations && generateJpaVersion != null ? " (forced to null because of <jpaAnnotations/>)"  :
+              (generateJPAAnnotations && generateJpaVersion == null ? " (forced to " + JpaVersion.latest().name() + " because of <jpaVersion/>)" : "")));
```
lukaseder: I think in this case, it's sufficient to simply print the version itself. No need to add any logic (given that we won't have the enum in Java)

jOOQ-codegen/src/main/java/org/jooq/util/JavaGenerator.java
```diff
@@ -4439,43 +4443,45 @@ protected void printTableJPAAnnotation(JavaWriter out, TableDefinition table) {
                 out.print(scala ? ")" : "}");
             }
 
-            StringBuilder sb2 = new StringBuilder();
-            String glue2 = "\n";
+            if (generateJpaVersion() != null && generateJpaVersion().supportedIndex()) {
```
lukaseder: If the configured version is null, then the expression should yield true. Thus:

```java
if (generateJpaVersion() == null || ...) {
```

I'll fix this as well...

PR #6997, opened 2018-01-05
Support system properties in Maven code generation

pvorb:
Adds a property `jooq.codegen.skip`, so generation can be skipped dynamically via `-Djooq.codegen.skip`.

There may be situations where you want to skip execution of `jooq-codegen-maven` for a single run, for example when you know that your db schema has not changed and you want to save some time. If I don't miss anything, currently you can only skip codegen by changing the pom and configuring `<skip>true</skip>`. Most Maven plugins have a `<plugin>.skip` property to make this easier. This PR adds such a property.

With this change you can skip jOOQ codegen e.g. by running

    mvn install -Djooq.codegen.skip

which is equivalent to

    mvn install -Djooq.codegen.skip=true

This PR is related to #4505.

### DIFF
--- a/jOOQ-codegen-maven/src/main/java/org/jooq/util/maven/Plugin.java
+++ b/jOOQ-codegen-maven/src/main/java/org/jooq/util/maven/Plugin.java
@@ -95,7 +95,9 @@ public class Plugin extends AbstractMojo {
     /**
      * Whether to skip the execution of the Maven Plugin for this module.
      */
-    @Parameter
+    @Parameter(
+        property = "jooq.skip"
+    )
     private boolean                      skip;
 
     /**


pvorb:
As I found out, you can already skip it with another property

    mvn -Djooq-codegen-maven.skip

already works. I wasn't aware that Maven uses the plugin artifact name as a default prefix.

So probably this PR should be closed. It would be nice if you could document that property, though.

`<skip>true</skip>` is [already documented in section 6.26](https://www.jooq.org/doc/3.10/manual-single-page/#codegen-maven).

lukaseder:
You're right, thanks for the hint. Will document this in the manual: #7000

pvorb:
Heck, I was wrong about `-Djooq-codegen-maven.skip`. I missed that at one point in playing with properties I added `<skip>true</skip>` to my POM. Could you please verify?

Probably this PR is still relevant in that case? Of course you could also configure `<skip>${jooq.skip}</skip>` and use `-Djooq.skip` but every user could invent their own property.

lukaseder:
Will look into it

lukaseder:
I'll implement this via defined properties:

- `jooq.codegen.skip`
- `jooq.codegen.configurationFile`
- `jooq.codegen.logging`

Both the Maven plugin as well as the standalone code generator will support these

lukaseder:
Fixed with #7011

pvorb:
That's awesome. Thanks!

Issue #7049, opened 2018-01-13
Generated Tables.scala has unused imports


er1c:
### Expected behavior and actual behavior:

The generated Tables.scala file includes `import` statements for the 

```scala
import ta.db.generated.tables.AdminCatalogPermission
import ta.db.generated.tables.AdminSite
import ta.db.generated.tables.AdminUser

...

object Tables {

  /**
   * The table <code>tecalliance.admin_catalog_permission</code>.
   */
  val ADMIN_CATALOG_PERMISSION = ta.db.generated.tables.AdminCatalogPermission.ADMIN_CATALOG_PERMISSION

  /**
   * The table <code>tecalliance.admin_site</code>.
   */
  val ADMIN_SITE = ta.db.generated.tables.AdminSite.ADMIN_SITE

  /**
   * The table <code>tecalliance.admin_user</code>.
   */
  val ADMIN_USER = ta.db.generated.tables.AdminUser.ADMIN_USER
```

The project I'm working on happens to have `-Xfatal-warnings` turned on in scala, so this will generate a compile error with the generated jooq tables:

```
[error] C:\Users\Eric Peters\IdeaProjects\ta-catalog\orig\src\main\scala\ta\db\generated\Tables.scala:11:31: Unused import
[error] import ta.db.generated.tables.AdminCatalogPermission
[error]                               ^
[error] C:\Users\Eric Peters\IdeaProjects\ta-catalog\orig\src\main\scala\ta\db\generated\Tables.scala:12:31: Unused import
[error] import ta.db.generated.tables.AdminSite
[error]                               ^
```

I believe this is easily fixed by simply adding the `className` to the generated scala code here: https://github.com/jOOQ/jOOQ/blob/eec46ca0835bb5e5f433ad76adc75ed31fabb0bc/jOOQ-codegen/src/main/java/org/jooq/util/JavaGenerator.java#L2769

`out.tab(1).println("val %s = %s", id, fullId);` 

becomes

`out.tab(1).println("val %s : %s = %s", id, className, fullId);`

I've tested this locally and it fixes my  `-Xfatal-warnings` problem, I'll submit a pull request and reference this issue id with it.

The generated code looks like:

```
  /**
   * The table <code>tecalliance.admin_catalog_permission</code>.
   */
  val ADMIN_CATALOG_PERMISSION : AdminCatalogPermission = ta.db.generated.tables.AdminCatalogPermission.ADMIN_CATALOG_PERMISSION

  /**
[...]

PR #7050, opened 2018-01-13
[7049] Add type information in Tables.scala to avoid scalac unused import warning

er1c:
Fixes minor issue: https://github.com/jOOQ/jOOQ/issues/7049

I've previously signed/authorized the jOOQ contribution agreement


lukaseder:
Thanks for the suggestion. However, I prefer not to add the explicit type declaration (there had been many requests to remove "noise" in the past) but to simply avoid the imports. I'll fix this accordingly.

In the long run, we should try to prevent adding import statements eagerly and only import types if they are actually used. I've created another issue for this: #7058

lukaseder:
> but adding types doesn't always have to be avoided :)

Oh, I agree. But between a fix and a workaround, I prefer the fix :-)

PR #7077, opened 2018-01-16
Formatting tweaks

er1c:
This fetch based upon the mapper still seems weird: https://github.com/jOOQ/jOOQ/pull/7077/commits/02e7e62c5dd22460dcf69521b27f7f0f5168b63f#diff-c9f97bdfa70478b5ea299842ddbc660aL162

I'm not evem sure how this compile: https://github.com/jOOQ/jOOQ/blob/b2f9a4e5ad55b71b760a73f1e64c56cc2cce2fb6/jOOQ/src/main/java/org/jooq/ResultQuery.java#L652 isn't RecordMapper defined as `? extends R`?

IntelliJ doesn't like this line and complains about it, but the scala compiler "works"  Despite the fact we previously had a made-up bounds for `RecordMapper`

punting it for now because I couldn't get a typing to work properly, and I don't understand/didn't dive into how the mapper is suppose to work

### DIFF
--- a/jOOQ-codegen/src/main/java/org/jooq/util/JavaGenerator.java
+++ b/jOOQ-codegen/src/main/java/org/jooq/util/JavaGenerator.java
@@ -3075,7 +3075,7 @@ public class JavaGenerator extends AbstractGenerator {
                     separator,
                     generateImmutablePojos() ? "" : "private var ",
                     getStrategy().getJavaMemberName(column, Mode.POJO),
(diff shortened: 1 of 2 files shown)


lukaseder:
I'm sorry, I'm not sure I understand what you're criticising here, nor how this relates to your two commits.

Ideally, for each unit of work, there should be a separate issue and a separate commit that can be reviewed individually. I'm thus rejecting this, keeping the changes in https://github.com/jOOQ/jOOQ/pull/7077/commits/500a6a861ef4a4b89b8563de6e37292c9a404870, through #7078 and #7079

Issue #1969, opened 2012-11-16
Add support for views expressed in jOOQ


lukaseder:
It should be possible to register ...
- plain SQL
- jOOQ client code

... as "views" that are made accessible through code generation. Such views would then generate `org.jooq.Table` objects with related `org.jooq.Field` column objects. An example:

``` java
select(V.COL1, V.COL2).from(V).where(V.COL3.equal(1));
```

... where `V` is a generated object, generated from either:

``` java
select(A.X1.as("col1"), A.X2.as("col2"), A.X3.as("col3")).from(A).where(...)
```

or from:

``` java
String sql = "SELECT a.x1 col1, a.x2 col2, a.x3 col3 FROM a WHERE ...";
```

Read more about the original idea here:
- https://groups.google.com/d/topic/jooq-user/FhDh-kkdxhw/discussion
- https://groups.google.com/d/msg/jooq-user/7hfDlmEwfQI/pblBBG4EBwAJ

thiloplanz:
> The advantage of those views (over the feature discussed in #3796) is the fact that you can reference columns by name, rather than by index - which is much more natural and readable.

Today I hit the 22-columns limit for typesafe Records for the first time. In addition to it getting really tedious to keep track of columns by index, a view would presumably also remain typesafe for as many columns as you want ( or your application forces you to have )

lukaseder:
> Today I hit the 22-columns limit for typesafe Records for the first time.

It was inevitable :)

> In addition to it getting really tedious to keep track of columns by index

Yes, true record types (named-attribute tuples) would be awesome in Java. Unfortunately, it is currently not possible.

> a view would presumably also remain typesafe for as many columns as you want ( or your application forces you to have )

No, the generic type safety that is used in unions for instance, is always index-based. While it would be possible to have views with formal column declarations, there is actually no formal connection between the column and the view from a Java parser perspective.

Of course, I'm not quite sure what use-case you had in mind when you thought that these views might be a solution for it...

thiloplanz:
> a view would presumably also remain typesafe for as many columns as you want 

All I meant to say is that with the view, you could refer to columns by name (where name is a compile-time symbol pointing to a typed column definition), just as you can for tables (or database views) right now, even if the table has 321 columns. On the "receiving end",  you would not be dealing with `Select<Record21<T1,T2......>` anymore, but have a `Select<FooViewRecord>`.

lukaseder:
[...]

PR #7115, opened 2018-01-30
Allow usage of derived table as aliased parameter of TableImpl constructor

victorbr:
This is following up my [post in jOOQ UG](https://groups.google.com/forum/#!topic/jooq-user/YXUtvMP3GFI) 
I'm terribly sorry if it's a hack! I tested that it solves my specific scenario but I don't know how to check whether it doesn't create a problem.
Anyway, @lukaseder, please regard this as a suggestion, nothing more! Thank you!

### DIFF
--- a/jOOQ/src/main/java/org/jooq/impl/TableImpl.java
+++ b/jOOQ/src/main/java/org/jooq/impl/TableImpl.java
@@ -153,8 +153,14 @@ public class TableImpl<R extends Record> extends AbstractTable<R> {
 
         this.fields = new Fields<R>();
 
-        if (aliased != null)
-            alias = new Alias<Table<R>>(aliased, name);
+        if (aliased != null) {
+            Alias<Table<R>> existingAlias = Tools.alias(aliased);
+            if (existingAlias != null) {
+                alias = existingAlias;
+            } else {
+                alias = new Alias<Table<R>>(aliased, name);
+            }
+        }
         else
             alias = null;
 
--- a/jOOQ/src/main/java/org/jooq/impl/Tools.java
+++ b/jOOQ/src/main/java/org/jooq/impl/Tools.java
@@ -4198,11 +4198,11 @@ final class Tools {
             return null;
     }
 
-    static final Alias<? extends Table<?>> alias(Table<?> table) {
+    static final <R extends Record> Alias<Table<R>> alias(Table<R> table) {
         if (table instanceof TableImpl)
-            return ((TableImpl<?>) table).alias;
+            return ((TableImpl<R>) table).alias;
         else if (table instanceof TableAlias)
-            return ((TableAlias<?>) table).alias;
+            return ((TableAlias<R>) table).alias;
         else
             return null;
     }


lukaseder:
Thanks you for your suggestion, @victorbr. May I ask you two things:

- Can you please sign the transfer of rights agreement, which allows us to dual-license your contribution? https://github.com/jOOQ/jOOQ/blob/master/CONTRIBUTING.md
- Would you mind showing how to use your suggested change in an example jOOQ query?

victorbr:
Thank you @lukaseder !
I have signed the agreement.
Here's an imaginary piece of code one could use to test this:
```
package org.jooq.example;

import org.jooq.Record;
import org.jooq.Table;
import org.jooq.TableField;
import org.jooq.impl.DSL;
import org.jooq.impl.SQLDataType;
import org.jooq.impl.TableImpl;

public class DerivedTableExample {

    static class MyTable extends TableImpl<Record> {

        static final MyTable TABLE = new MyTable();

        public MyTable() {
            super("my_table");
        }
        MyTable(String alias, Table<Record> aliased) {
            super(alias, null, aliased);
        }
        public final TableField<Record, Long> tenant_id = createField("tenant_id", SQLDataType.BIGINT);
        public final TableField<Record, String> data1 = createField("data1", SQLDataType.VARCHAR);
        public final TableField<Record, Integer> data2 = createField("data2", SQLDataType.INTEGER);
    }

    private static MyTable tenantView(MyTable table, Long tenantId) {
        return new MyTable("tenant_view", DSL.selectFrom(table).where(table.tenant_id.eq(tenantId)).asTable());
    }

    public static void main(String[] args) {
        MyTable tenantTable = tenantView(MyTable.TABLE, 5L);
        DSL.select(tenantTable.data1, tenantTable.data2).from(tenantTable).fetch();
    }
}
```

lukaseder:
I'm now reviewing this. Of course, this is a bit of a hack, so I'm not convinced that we're going down the right path. Anyway, your unwrapping of the existingAlias works for this particular usage, but it's not entirely correct. The resulting SQL will ignore the name that is given to the view. In your case, the derived table should be named `my_table`. Instead, the generated alias from the `asTable()` call is used.

But I like the idea of using this `aliased` reference for other things than the `MyTable` type itself, even if this is internal API.

I'll investigate how your suggestion can be improved.

lukaseder:
I'll merge and fix afterwards.

lukaseder:
Fix was almost correct. New fix passes these tests:

```java
static class T7115_AB extends TableImpl<Record2<Integer, String>> {
    T7115_AB(String alias, Table<Record2<Integer, String>> aliased) {
        super(DSL.name(alias), null, aliased);
    }
    public final TableField<Record2<Integer, String>, Integer> A = createField("a", SQLDataType.INTEGER);
    public final TableField<Record2<Integer, String>, String> B = createField("b", SQLDataType.VARCHAR);
}

static class T7115_FIRST_NAME_LAST_NAME extends TableImpl<Record2<String, String>> {
    T7115_FIRST_NAME_LAST_NAME(String alias, Table<Record2<String, String>> aliased) {
        super(DSL.name(alias), null, aliased);
    }
    public final TableField<Record2<String, String>, String> F = createField("f", SQLDataType.VARCHAR);
    public final TableField<Record2<String, String>, String> L = createField("l", SQLDataType.VARCHAR);
}

public void testAliasingDerivedTablesWithTableType() throws Exception {
    Table<Record2<Integer, String>> t1 = values(row(1, "a"), row(2, "b")).as("t", "a", "b");
    T7115_AB a1 = new T7115_AB("alias", t1);
    Result<Record2<Integer, String>> r1 = create().select(a1.A, a1.B).from(a1).orderBy(a1.A).fetch();
    assertEquals(asList(1, 2), r1.getValues(a1.A));
    assertEquals(asList("a", "b"), r1.getValues(a1.B));

    Table<Record2<Integer, String>> t2 = selectFrom(values(row(1, "a"), row(2, "b")).as("t", "a", "b")).asTable();
    T7115_AB a2 = new T7115_AB("alias", t2);
    Result<Record2<Integer, String>> r2 = create().select(a2.A, a2.B).from(a2).orderBy(a2.A).fetch();
    assertEquals(asList(1, 2), r2.getValues(a2.A));
    assertEquals(asList("a", "b"), r2.getValues(a2.B));

    Table<Record2<String, String>> t3 =
        select(TAuthor_FIRST_NAME().as("f"), TAuthor_LAST_NAME().as("l"))
        .from(TAuthor())
        .where(TAuthor_ID().eq(1))
        .asTable();
    T7115_FIRST_NAME_LAST_NAME a3 = new T7115_FIRST_NAME_LAST_NAME("alias", t3);
    Record2<String, String> r3 = create().select(a3.F, a3.L).from(a3).fetchOne();
    assertEquals(AUTHOR_FIRST_NAMES.get(0), r3.get(a3.F));
    assertEquals(AUTHOR_LAST_NAMES.get(0), r3.get(a3.L));
}
```

With these tests, we'll guarantee that the expected functionality won't break in the future, even if it is using internal API.

Thanks again for the suggestion and for taking the time to investigate this!

I'm still going to keep #1969 open as I think this functionality deserves some cool API.

victorbr:
Wow, @lukaseder, thank you for diving into this and making sure this "hack" becomes a real solution! I'm now flattered to have contributed a tiny piece to jOOQ - my favorite library!

lukaseder:
> I'm now flattered to have contributed a tiny piece to jOOQ - my favorite library!

Thanks for your nice words. Looking forward to more contributions ;-)

lukaseder:
In fact, `CustomTable` should be adapted as well: #7145, and that will become actual public API, not internal API.

Relevant, similar discussion on the user group:
https://groups.google.com/forum/#!topic/jooq-user/Xff7ajkwDUk

PR #7195, opened 2018-02-16
More extensible LoggerListener

lau-sch:
Making it possible to extend the multi line logging behavior.

I send logging events to ElasticSearch, and multi line records with a table view is nearly useless. Changing the access modifier from private to protected, an extension can easily change these behaviors, achieving the result below with minimal code.

Original code output:
kibana-jooq-original

Modified code output:
kibana-jooq-modified

### DIFF
--- a/jOOQ/src/main/java/org/jooq/tools/LoggerListener.java
+++ b/jOOQ/src/main/java/org/jooq/tools/LoggerListener.java
@@ -80,7 +80,7 @@ public class LoggerListener extends DefaultExecuteListener {
      */
     private static final long serialVersionUID = 7399239846062763212L;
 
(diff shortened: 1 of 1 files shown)


lukaseder:

jOOQ is very opinionated about this sort of extensibility. Once we make such a method protected, we need to:

- Maintain it
- Design it much better
- Keep it backwards compatible

This is a high price to pay for a very low benefit.

Additionally, we're currently completely free to change the behaviour for the out-of-the-box logger (which is mainly useful for debug purposes) at any moment without worrying about breaking client code. If you really want to have your own log format, simply turn off jOOQ's built-in `LoggerListener` through `Settings.executeLogging = false` and roll your own logging `ExecuteListener`

Issue #7126, opened 2018-01-30
Add support for the PostgreSQL timestamp value [+-]INFINITY


ciaccia:
Starting with jOOQ 3.10, "timestamp with time zone" columns are mapped to `java.time.OffsetDateTime` instead of previous `java.sql.Timestamp`.

This new model is not able to handle values such as `'-infinity'::timestamptz` properly 
([documentation here](https://www.postgresql.org/docs/10/static/datatype-datetime.html#DATATYPE-DATETIME-SPECIAL-TABLE)).

Before the change, jOOQ internally called `java.sql.ResultSet#getTimestamp(int)` which knows how to handle special values such as +-infinity.

Now jOOQ gets the String version of the timestamp from the result set (using `java.sql.ResultSet#getString(int)` and passes the string value to `org.jooq.impl.DefaultBinding#offsetDateTime(String)`. Eventually an exception is throw complaining about `-infinity` not having the valid format.

Unfortunately I cannot fix this behavior using &lt;forcedTypes&gt; since jOOQ binding in this scenario is not passing Postgres' native type, rather `OffsetDateTime`. Ideally in this case the binding should receive a "native" type instead of a constructed one.

lukaseder:
Interesting, thanks for pointing this out.

> Unfortunately I cannot fix this behavior using <forcedTypes> since jOOQ binding in this scenario is not passing Postgres' native type

I think you're confusing the terms "converter" and "binding", here. You should be able to implement your own data type binding to work around the problem:
https://www.jooq.org/doc/latest/manual/code-generation/custom-data-type-bindings

Let me know if this helps

ciaccia:
Hi Lukas,
Please excuse me for the confusion. You are right, with a custom binder it is possible to avoid the problem. I got confused since the signature for the custom binder must be something like

    public class PGTimestampBinder implements org.jooq.Binding<OffsetDateTime, OffsetDateTime> {

which did not make any sense to me (why U and T in the template must be the same?). Nevertheless, with a custom binder on all timestamp column it is working fine. In the jOOQ configuration file I added it for all column with the timestamp with time zone type:

    <forcedTypes>
      <forcedType>
        <userType>java.time.OffsetDateTime</userType>
        <binding>com.example.PGTimestampBinder</binding>
        <expression>.*</expression>
        <types>.*timestamptz.*</types>
      </forcedType>
    </forcedTypes>

[...]

PR #7251, opened 2018-03-07
[#7126] Binding Postgres TIMESTAMP WITH TIME ZONE type doesn't suppor…

ciaccia:
…t values like INFINITY

### DIFF
--- a/jOOQ/src/main/java/org/jooq/impl/DefaultBinding.java
+++ b/jOOQ/src/main/java/org/jooq/impl/DefaultBinding.java
@@ -2392,26 +2392,13 @@ public class DefaultBinding<T, U> implements Binding<T, U> {
         // Probably a bug, let OffsetDateTime or OffsetTime report it
         else {
 
-            // [#7126] Postgresql can return Special Date/Time Inputs such as "epoch", "infinity", "now", ...
-            if ("epoch".equals(formatted)) {
-                return "1970-01-01T00:00Z";
-            } else if ("infinity".equals(formatted)) {
+            // [#7126] Binding Postgres TIMESTAMP WITH TIME ZONE type doesn't support values like INFINITY
+            if ("infinity".equals(formatted)) {
                 // DATE_POSITIVE_INFINITY taken from org.postgresql.PGStatement
                 return "+292278994-08-16T23:00Z";
             } else if ("-infinity".equals(formatted)) {
                 // DATE_NEGATIVE_INFINITY taken from org.postgresql.PGStatement
                 return "-292275055-05-16T23:00Z";
-            } else if ("now".equals(formatted)) {
-                return OffsetDateTime.now().toString();
-            } else if ("today".equals(formatted)) {
-                return OffsetDateTime.ofInstant(LocalDate.now().atStartOfDay(ZoneId.systemDefault()).toInstant(), ZoneId.systemDefault()).toString();
-            } else if ("tomorrow".equals(formatted)) {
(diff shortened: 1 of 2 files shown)


jOOQ/src/main/java/org/jooq/impl/DefaultBinding.java
```diff
@@ -2389,6 +2391,29 @@ private static final String preparse(String formatted, boolean includeDate) {
...
+            if ("epoch".equals(string)) {
+                return "1970-01-01T00:00Z";
+            } else if ("infinity".equals(string)) {
+                // DATE_POSITIVE_INFINITY taken from org.postgresql.PGStatement
+                return "+292278994-08-16T23:00Z";
+            } else if ("-infinity".equals(string)) {
+                // DATE_NEGATIVE_INFINITY taken from org.postgresql.PGStatement
+                return "-292275055-05-16T23:00Z";
+            } else if ("now".equals(string)) {
+                return OffsetDateTime.now().toString();
```
lukaseder: Hmm, the unfortunate side-effect of this usage is that `now` is an expression on the PostgreSQL *server*, which is now evaluated on the *client* (which may be off by whatever number of seconds/minutes/timezones/etc)

jOOQ/src/main/java/org/jooq/impl/DefaultBinding.java
```diff
@@ -2389,6 +2391,29 @@ private static final String preparse(String formatted, boolean includeDate) {
...
+            } else if ("today".equals(string)) {
+                return OffsetDateTime.ofInstant(LocalDate.now().atStartOfDay(ZoneId.systemDefault()).toInstant(), ZoneId.systemDefault()).toString();
```
lukaseder: ... likewise, calling `now()` or `ZoneId.systemDefault()` seams treacherous, as we're going to get the *client* default, not the *server* default. I had rather these continue failing than getting this wrong.

I agree with the epoch / infinity ones, though

ciaccia: Hallo Lukas, 
Please excuse me for this, I wrote the patch without testing it deeply. Actually the only two values which are returned as text (and not as ISO 8601) are `"-infinity"` and `"infinity"`. All the other values are returned in ISO 8601 format, using the time zone of the Postgresql server.

Therefore you can delete all the problematic lines (i.e. `"epoch"`, `"now"`, `"today"`, `"tomorrow"`, `"yesterday"`, `"allballs"`) since the server will make all the computations.

    create table timezoneissue
    (
        v timestamp with time zone
    );
    
    insert into timezoneissue(v) values ('-infinity'::timestamptz), ('infinity'::timestamptz), ('epoch'::timestamptz), ('now'::timestamptz), ('today'::timestamptz), ('tomorrow'::timestamptz), ('yesterday'::timestamptz); --('allballs'::timestamptz);

    select x.v from timezoneissue x;

lukaseder: I see, thanks for the clarification

ciaccia: Also `"epoch"`can be deleted, since the JDBC driver will return a nice `"1970-01-01 01:00:00+01"` string 😃

ciaccia:
Hi Lukas,
As I commented on 9th of March (https://github.com/jOOQ/jOOQ/pull/7251#discussion_r173503410), the only values which are possibly returned by the JDBC driver are `"infinity"` and `"-infinity"`.  Your concerns about the timezone of the PG server should not applicable to the fix anymore. I changed my repository accordingly and issued a merge request.

lukaseder:

The issue #7126 was fixed in the meantime. Looks like this PR was stale anyway.

Issue #5873, opened 2017-02-15
Add Setting.apostropheEscaping to allow for alternative escaping of apostrophes


lukaseder:
In most SQL databases, single quotes inside of string literals are escape by doubling them, e.g.:

```sql
SELECT 'The string''s length is 25' FROM dual
```

But in some SQL dialects (e.g. [SOQL](https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql_select_quotedstringescapes.htm)), they need to be escaped with a backslash:

```sql
SELECT 'The string\'s length is 25' FROM dual
```

While we will not add formal support for SOQL, it certainly makes sense to make this escaping configurable, similar to `Settings.backlashEscaping` (we'll need to study the interaction between these two)

sawyna:
Hi I would like to take up this issue. I was the one who raised question on stackoverflow.

lukaseder:
@sawyna Thank you very much for your offer. Let me know if you need any help.

sawyna:
@lukaseder I am new to open source contribution. I might need some help to get started. I assume I should fork jooq repo, work on the issue and create a PR. Please correct me if I am wrong

lukaseder:
@sawyna No worries, thanks for asking.

I've just updated our contribution guidelines with some ideas: https://github.com/jOOQ/jOOQ/blob/master/CONTRIBUTING.md

Let me know if you still have any open questions that aren't covered in there.

lukaseder:
This will not be implemented in jOOQ. Rationale here: https://github.com/jOOQ/jOOQ/pull/7314#issuecomment-374185946

PR #7314, opened 2018-03-15
[#5873] Add Setting.quoteEscaping to allow for alternative escaping of single quote

zhurs:
https://github.com/jOOQ/jOOQ/issues/5873

### DIFF
--- a/jOOQ/src/main/java/org/jooq/conf/QuoteEscaping.java
+++ b/jOOQ/src/main/java/org/jooq/conf/QuoteEscaping.java
@@ -1,0 +1,47 @@
+
+
+
+
+
+
+
+
+package org.jooq.conf;
+
+import javax.xml.bind.annotation.XmlEnum;
+import javax.xml.bind.annotation.XmlType;
+
+
+/**
+ * <p>Java class for QuoteEscaping.
+ * 
+ * <p>The following schema fragment specifies the expected content contained within this class.
+ * <p>
+ * <pre>
+ * &lt;simpleType name="QuoteEscaping">
+ *   &lt;restriction base="{http://www.w3.org/2001/XMLSchema}string">
+ *     &lt;enumeration value="DEFAULT"/>
+ *     &lt;enumeration value="QUOTE"/>
+ *     &lt;enumeration value="ESCAPE"/>
+ *   &lt;/restriction>
+ * &lt;/simpleType>
+ * </pre>
+ * 
+ */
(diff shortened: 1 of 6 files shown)


lukaseder:
Thank you very much for your suggested implementation. While I review, may I ask you to please read our contribution guidelines and sign the transfer of rights agreement linked in it?

lukaseder:
I've reviewed your suggestion. Very nice - seems complete, at first sight. Will check again with the integration tests after the agreement has been signed and the changes merged.

zhurs:
Thank you for the super-fast response!
I read documents and electronically signed the transfer of rights agreement.

lukaseder:
Thank you very much. Will merge on Monday

lukaseder:
I'll rename the flag to ApostropheEscaping. The term "quote" or "quoting" in jOOQ referes to quoting identifiers, such as:

- `"abc"` (standard SQL)
- `[abc]` (T-SQL)
- `` `abc` `` (MySQL)

lukaseder:
After giving this more thought, I'll revert your suggested implementation, it is incorrect. The original issue #5873 might have been a bit misleading because it suggested adding a new flag, when it shouldn't have (although there was a disclaimer: *we'll need to study the interaction between these two*).

The existing flag `Settings.backslashEscaping` currently governs MySQL-specific behaviour where both escaping through doubling the apostrophe *and* through using a backslash are possible. (This is of course nuts)

Now, the use-case described in #5873 is that *only* backslash escaping is possible in some dialects, e.g. SOQL. It is better to enhance the existing `BackslashEscaping` with an additional option:

- `OFF`: Backslash escaping is not possible (existing)
- `ON`: Backslash escaping is enabled, along with repeating apostrophe escaping (existing)
- `DEFAULT`: `ON` for MySQL and MariaDB, `OFF` for all the others
- `ONLY`: Backslash escaping is enabled, repeating apostrophe escaping is not enabled

The error in the suggested implementation lies in the fact that:

- The two flags are both conflicting and confusing
- When allowing to escape things using a backslash, then the backslash must escape *all* things, not just apostrophes. Very unfortunately, SOQL also proceeds to using the backslash for all sorts of other things: https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql_select_quotedstringescapes.htm

Since jOOQ doesn't actually support SOQL nor does jOOQ plan on doing so, I think I'll simply reject #5873. I'm sorry for the inconvenience this has caused.

Issue #7324, opened 2018-03-17
Support Kotlin metadata in DefaultRecordMapper


vojtapol:
Currently, `DefaultRecordMapper` supports many different mapping strategies including `@ConstructionProperties`. 

In Kotlin, it is common to define immutable data classes like so:

```kotlin
data class Person(val name: String, val age: Int)
```

This, when compiled, generates something that would be represented in Java like so:
```java
@Metadata(
   // note how property names are stored as metadata and can be accessed at runtime
   d2 = {"vojtapol/packing/Person;", "", "name", "", "age", "", "(Ljava/lang/String;I)V", "getAge", "()I", "getName", "()Ljava/lang/String;", "component1", "component2", "copy", "equals", "", "other", "hashCode", "toString", "production sources for module packing"}
)
public final class Person {
   @NotNull
   private final String name;
   private final int age;

   @NotNull
   public final String getName() {
      return this.name;
   }

   public final int getAge() {
      return this.age;
   }

   public Person(@NotNull String name, int age) {
      Intrinsics.checkParameterIsNotNull(name, "name");
      super();
      this.name = name;
      this.age = age;
   }

   // ommitted some convenience methods
}
```

As you know, when we tell `DefaultRecordMapper` to map a record into `Person` it will use the constructor and map it if the number of parameters in the constructor exactly matches the record. This is not ideal as it is very error-prone. If the parameters are all the same type and their order is altered (or the database column order is altered) the mapper will map the type incorrectly and no warning is logged. 

To mitigate this I propose leveraging Kotlin's `@Metadata` to match parameter names. Similarly to what is currently done with `@ConstructorProperties`. This would make using jOOQ from Kotlin's data classes seamless and no annotations would be required.

From Java, this could be done using the `kotlin-reflect` library like so:

```java
Person p = new Person("vojtapol", 4);
Class<? extends Person> type = p.getClass();
KClass<? extends Person> kotlinClass = JvmClassMappingKt.getKotlinClass(type);
Collection<? extends KProperty1<? extends Person, ?>> properties = KClasses.getDeclaredMemberProperties(kotlinClass);
```

Where `properties` now contain a list of a person's properties including their names.

I would be willing to help out or contribute via PR if necessary.

lukaseder:
Thank you very much for your suggestion. From what I can tell, this `Metadata` annotation isn't documented (https://stackoverflow.com/q/36816521/521799), so we shouldn't read it directly. Meta-
[...]

PR #7329, opened 2018-03-19
[#7324] Support Kotlin data classes in DefaultRecordMapper

vojtapol:
My naive attempt to leverage Kotlin `@Metadata` when mapping to Kotlin data classes.

Only [primary constructors](https://kotlinlang.org/docs/reference/classes.html#constructors) are considered. If the number of parameter matches the number of fields in the record, their names are checked. If all the names match the record is then mapped to the corresponding parameters by the names. 

If some of the conditions are not met an exception is thrown. I don't think we should default to any other mapping strategies for Kotlin data classes as not-matching the property names can lead to silent errors.

### DIFF
--- a/jOOQ/pom.xml
+++ b/jOOQ/pom.xml
@@ -101,6 +101,11 @@
             <scope>provided</scope>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-reflect</artifactId>
+            <version>1.2.30</version>
+        </dependency>
 
         <!-- From JDK 9 onwards, the JAXB dependency needs to be made explicit -->
         <dependency>
--- a/jOOQ/pom.xml
+++ b/jOOQ/pom.xml
@@ -101,11 +101,6 @@
             <scope>provided</scope>
             <optional>true</optional>
         </dependency>
-        <dependency>
-            <groupId>org.jetbrains.kotlin</groupId>
-            <artifactId>kotlin-reflect</artifactId>
-            <version>1.2.30</version>
-        </dependency>
 
         <!-- From JDK 9 onwards, the JAXB dependency needs to be made explicit -->
         <dependency>
--- a/jOOQ/src/main/java/org/jooq/impl/DefaultRecordMapper.java
+++ b/jOOQ/src/main/java/org/jooq/impl/DefaultRecordMapper.java
@@ -54,15 +54,8 @@ import static org.jooq.tools.reflect.Reflect.accessible;
 
 import java.beans.ConstructorProperties;
 import java.lang.invoke.MethodHandles.Lookup;
+import java.lang.reflect.*;
 import java.lang.reflect.Array;
-import java.lang.reflect.Constructor;
-import java.lang.reflect.Executable;
-import java.lang.reflect.InvocationHandler;
-import java.lang.reflect.Method;
-import java.lang.reflect.Modifier;
-import java.lang.reflect.Parameter;
-import java.lang.reflect.ParameterizedType;
-import java.lang.reflect.Proxy;
 import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -92,6 +85,7 @@ import org.jooq.exception.MappingException;
 import org.jooq.tools.Convert;
 import org.jooq.tools.StringUtils;
 import org.jooq.tools.reflect.Reflect;
+import org.jooq.tools.reflect.ReflectException;
 
 /**
  * This is the default implementation for <code>RecordMapper</code> types.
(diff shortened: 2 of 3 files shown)


lukaseder:
Thank you very much for your suggestion. I will soon review your PR. In the meantime, may I ask you to read the contribution guidelines and sign the transfer of rights agreement, which allows us to dual-license your contribution:

jOOQ/pom.xml
```diff
@@ -101,6 +101,11 @@
             <scope>provided</scope>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-reflect</artifactId>
+            <version>1.2.30</version>
+        </dependency>
```
lukaseder: Hmm, we cannot actually have this dependency in jOOQ, not even as `<scope>provided</scope>`. What I meant by "meta-reflecting" is that we need to reflect on this reflection API. I'll comment again in the actual code.

jOOQ/src/main/java/org/jooq/impl/DefaultRecordMapper.java
```diff
@@ -79,6 +80,11 @@
 
 import javax.persistence.Column;
 
+import kotlin.jvm.JvmClassMappingKt;
+import kotlin.reflect.KClass;
+import kotlin.reflect.KFunction;
+import kotlin.reflect.KParameter;
+import kotlin.reflect.full.KClasses;
```
lukaseder: Without the dependency, these imports are not possible.

jOOQ/src/main/java/org/jooq/impl/DefaultRecordMapper.java
```diff
@@ -343,6 +349,33 @@ private final void init(E instance) {
...
 
+        // [#7324] Map Kotlin data classes
+        KClass<? extends E> kotlinClass = JvmClassMappingKt.getKotlinClass(type);
+        if (kotlinClass.isData()) {
+            KFunction<? extends E> primaryConstructor = KClasses.getPrimaryConstructor(kotlinClass);
+            List<KParameter> parameters = primaryConstructor.getParameters();
+
+            if (parameters.size() != fields.length) {
+                throw new MappingException("Primary constructor of Kotlin data class " + type + " did not match row type " + rowType);
+            }
```
lukaseder: Once we reflect on field names, I don't think this limitation should still be maintained. A record may have more or less fields than the type we're mapping it into. See also the Java 8 parameter name reflection implementation, as an example.

jOOQ/src/main/java/org/jooq/impl/DefaultRecordMapper.java
```diff
@@ -343,6 +349,33 @@ private final void init(E instance) {
...
+                boolean found = false;
+                for (KParameter kParameter : parameters) {
+                    String parameterName = kParameter.getName();
+                    if (name.equals(parameterName)) {
+                        found = true;
+                    }
+                }
+                if (!found) {
+                    throw new MappingException("Primary constructor of Kotlin data class " + type + " did not match row type " + rowType);
+                }
```
lukaseder: Again, we don't require an exact match between source and target fields. If we don't find a specific field in the target type, then we simply won't map it.

There are pros and cons to each approach, but for consistency reasons with existing `DefaultRecordMapper` behaviour, this exception shouldn't be thrown.

jOOQ/src/main/java/org/jooq/impl/DefaultRecordMapper.java
```diff
@@ -981,6 +1014,44 @@ public final E map(R record) {
         }
     }
 
+    private class KotlinDataClassMapper implements RecordMapper<R, E> {
```
lukaseder: Is this class really necessary? Can't we use the existing `ImmutablePOJOMapperWithParameterNames`?

jOOQ/pom.xml
vojtapol: I see. I noticed there are almost no dependencies. I will try to work around it . To satisfy my curiosity, why can't jOOQ have that dependency?

lukaseder: It is a very very very *very* ***VERY*** bad idea for a library to have dependencies. Blog post here:
https://blog.jooq.org/2016/08/11/all-libraries-should-follow-a-zero-dependency-policy

vojtapol: This is what I propose as a solution:

Kotlin encodes metadata in a specific protobuf format. `kotlin-reflect` parses that and provides API to access all of the metadata. That approach was used in this PR. Since we can't depend on `kotlin-reflect` we would have to get the `Metadata` annotation using Java reflection and parse the protobuf data ourselves.

There is an official project `kotlinx.reflect.lite` which is a very lightweight implementation of `kotlin-reflect` which only supports reading parameter names and nullability. After reading the article you wrote I think you could accept that we rip out only the most necessary parts of this library to make extracting parameter names from the `Metadata` annotation possible. I would except this could consist of only one additional class being added to jOOQ. The `kotlinx.reflect.lite` library is licensed as Apache License 2.0 so it could be included in both the free and commercial version of jOOQ.

Here is a minimal example of obtaining the parameter names using this library:

```java
ClassMetadata classMetadata = ReflectionLite.INSTANCE.loadClassMetadata(PersonKotlin.class);
ConstructorMetadata constructorMetadata = classMetadata.getConstructor(p.getClass().getDeclaredConstructors()[0]);
List<ParameterMetadata> parameters = constructorMetadata.getParameters();
List<String> names = new ArrayList<>();
for (ParameterMetadata parameter : parameters) {
    names.add(parameter.getName());
}
System.out.println(names);
```

Let me know what you think.

vojtapol: It seems that the `kotlinx.reflect.lite` depends on `com.google.protobuf`. I don't see how we could get rid of that. It makes very little sense to reimplement protobuf parser from scratch.

lukaseder: Of course not. As I said, we can have an implicit runtime dependency on `kotlin-reflect` by meta-reflecting on its API. If it's on the classpath, we can use it. That's what we're doing with vendor-specific JDBC API as well, including ojdbc.

E.g. instead of 

```java
KClass<? extends E> kotlinClass = JvmClassMappingKt.getKotlinClass(type);
```

Write

```java
Reflect.on("com.example.JvmClassMappingKt").call("getKotlinClass", type);
```

vojtapol: Oh I see. I did not realize this could be done. I will look into it.

vojtapol:
I have reviewed contributors guidelines and signed the paperwork already.

I have pushed an updated version that does not use any additional dependencies and reuses `ImmutablePOJOMapperWithParameterNames`. 

We might want to think about exceptions. There are two conditions in general that might lead to exceptions:

1. `kotlin-reflect` is not on the classpath. Then the behavior defaults to what was there before. It should be however documented that it is advantageous to have `kotlin-reflect` if the user uses Kotlin classes.

2. jOOQ fell behind `kotlin-reflect` and some new version broke the reflection. This one is ugly but hopefuly the `kotlin-reflect` API remains stable.

lukaseder:
Thanks a lot. I'll merge and possibly adapt 1-2 minor things.

lukaseder:
Improvements that I'll apply:

- Java 6 compatibility (no diamond operator, no union types in catch blocks)
- Cache the reflection call on `kotlin.jvm.JvmClassMappingKt` (and others) in static variables

Maybe I'll apply further improvements once I've added integration tests...

lukaseder:
There's now also a log message when Kotlin is available (`kotlin.jvm.JvmClassMappingKt` can be found via reflection), but not the `kotlin-reflect` dependency (`kotlin.reflect.full.KClasses` cannot be found)

lukaseder:
Once jOOQ is modularised for JDK 9 / 10, I'll review if this still works. Might be that such dependencies can no longer be found on the classpaths if a module doesn't explicitly import it...

lukaseder:
Thanks again for your contribution! Works nicely.

oripwk:
What are the chances of this feature making it into the 3.10.8 release?

lukaseder:
@oripwk New features are released only in minor releases to minimise the risk of regressions in patch releases, so this will ship with 3.11

Issue #7301, opened 2018-03-14
Allow customization of LoggerListener


Diagoras:
JOOQ has an _awesome_ default logging implementation - it captures everything I'm interested in and makes very clear what's happening. However, there doesn't seem to be a way to customize how things are logged. For example, I'd be interested in both logging at the INFO level and also using a marker.

Of course I can just do something ugly (like copy paste the class and modify all the log calls to include my changes), but I was wondering if there was a way to make logging more customizable. I'm not sure what would be the best way to do that - my usecase only requires changing the log level and passing a marker, but I could see others wanting features like selective logging (eg. turn off query execution time or returned results).

Thanks, and let me know if you need any more details.

lukaseder:
Thanks for your suggestion. Yeah, I think we could add a few more flags to `Settings` to fine-tune how we log things out of the box. Something along the lines of:

- `Settings.executeLoggingBindVariables` (log level)
- `Settings.executeLoggingSqlString` (log level)
- `Settings.executeLoggingSqlStringInline` (log level)
- `Settings.executeLoggingResult` (log level)
- `Settings.executeLoggingResultNumberOfRows` (int)
- `Settings.executeLoggingException` (log level)

lukaseder:
@Diagoras Notice: I wrote "Something along the lines of ...", this wasn't a complete suggestion of what needs to be done. I'll revert your actual adaptations of the logger and re-implement them. Will keep the parts that make sense.

lukaseder:
On second thought, I'll revert the entire thing. This is going to overengineer the out of the box logger. It's super easy to write your own logger by simply copy pasting the `LoggerListener`...

lukaseder:
... a more elaborate rationale:

- The suggested PR was far from complete and introduced a few significant regressions
- These things are much more work than they seem to cause, at first
- One thing that hasn't been covered yet is the documentation in the manual. Once these things are documented, they're hard to change

Reviewing the above concerns, the suggestion simply doesn't add enough value to justify the additional costs of maintaining the solution.

PR #7346, opened 2018-03-26
#7301: Allow customization of LoggerListener

Diagoras:
Let me know what you think. Here's my random collection of questions/concerns/comments:

- The version of Log4J you're using doesn't have `isWarnEnabled()` or `isErrorEnabled()`. I used a method that takes a deprecated enum instead, which I'm not super happy about.
- Is there a reason Settings isn't using Object.equals and Object.hash? The custom implementation is kind of verbose.
- This doesn't let you log with an SFL4J marker. That's kind of tricky to pull off, since you also support Log4J and java.util.Logging. I guess let people set a marker in the Settings, then only use it if logging with SFL4J?
- I treat FATAL as ERROR. Let me know if that's okay.

### DIFF
--- a/jOOQ/src/main/java/org/jooq/Log.java
+++ b/jOOQ/src/main/java/org/jooq/Log.java
@@ -162,6 +162,11 @@ public interface Log {
     void info(Object message, Object details, Throwable throwable);
 
     /**
+     * Check if <code>WARN</code> level logging is enabled.
+     */
+    boolean isWarnEnabled();
+
+    /**
      * Log a message in <code>WARN</code> level.
      *
      * @param message The log message
@@ -196,6 +201,11 @@ public interface Log {
     void warn(Object message, Object details, Throwable throwable);
 
     /**
+     * Check if <code>ERROR</code> level logging is enabled.
+     */
+    boolean isErrorEnabled();
+
+    /**
      * Log a message in <code>ERROR</code> level.
      *
      * @param message The log message
@@ -228,6 +238,84 @@ public interface Log {
      *            message
      */
     void error(Object message, Object details, Throwable throwable);
+
+    default boolean isEnabled(final Log.Level level) {
+        switch (level) {
+            case TRACE: return isTraceEnabled();
+            case DEBUG: return isDebugEnabled();
+            case INFO: return isInfoEnabled();
+            case WARN: return isWarnEnabled();
+            case ERROR: return isErrorEnabled();
+            case FATAL: return isErrorEnabled();
+            default: return false;
+        }
+    }
+
+    default void log(final Object message, final Log.Level level) {
+        log(message, (Object) null, level);
+    }
+
+    default void log(final Object message, final Object details, final Log.Level level) {
+        switch (level) {
+            case TRACE: {
+                trace(message, details);
+                return;
+            }
+            case DEBUG: {
+                debug(message, details);
+                return;
+            }
+            case INFO: {
+                info(message, details);
+                return;
+            }
+            case WARN: {
+                warn(message, details);
+                return;
+            }
+            case ERROR: {
+                error(message, details);
+                return;
+            }
+            case FATAL: {
+                error(message, details);
+                return;
+            }
+        }
+    }
+
+    default void log(final Object message, final Throwable throwable, final Log.Level level) {
+        log(message, null, throwable, level);
+    }
+
+    default void log(final Object message, final Object details, final Throwable throwable, final Log.Level level) {
+        switch (level) {
+            case TRACE: {
+                trace(message, details, throwable);
+                return;
+            }
+            case DEBUG: {
+                debug(message, details, throwable);
+                return;
+            }
+            case INFO: {
+                info(message, details, throwable);
+                return;
+            }
+            case WARN: {
+                warn(message, details, throwable);
+                return;
+            }
+            case ERROR: {
+                error(message, details, throwable);
+                return;
+            }
+            case FATAL: {
+                error(message, details, throwable);
+                return;
+            }
+        }
+    }
 
     /**
      * The log level.
--- a/jOOQ/src/main/java/org/jooq/conf/Settings.java
+++ b/jOOQ/src/main/java/org/jooq/conf/Settings.java
@@ -8,6 +8,8 @@
 
 package org.jooq.conf;
 
+import org.jooq.Log;
+
 import java.io.Serializable;
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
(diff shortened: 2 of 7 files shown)


lukaseder:
Thanks for your suggestion. I'll review this now.

jOOQ/src/main/java/org/jooq/Log.java
```diff
@@ -229,6 +239,84 @@
      */
     void error(Object message, Object details, Throwable throwable);
 
+    default boolean isEnabled(final Log.Level level) {
```
lukaseder: We cannot have default methods in jOOQ yet as we're still supporting Java 6 and 7 in our commercial distributions. The additional API definitely makes sense, though.

jOOQ/src/main/java/org/jooq/conf/Settings.java
```diff
@@ -65,6 +67,22 @@
     protected StatementType statementType = StatementType.PREPARED_STATEMENT;
     @XmlElement(defaultValue = "true")
     protected Boolean executeLogging = true;
+    @XmlElement(defaultValue = "DEBUG")
+    protected Log.Level executeLoggingBindVariables = Log.Level.DEBUG;
```
lukaseder: Did you generate these using XJC? Because this file is governed by and derived from https://www.jooq.org/xsd/jooq-runtime-3.10.2.xsd. You should also adapt https://github.com/jOOQ/jOOQ/blob/master/jOOQ/src/main/resources/xsd/jooq-runtime-3.11.0.xsd accordingly.

jOOQ/src/main/java/org/jooq/tools/LoggerListener.java
```diff
@@ -38,6 +38,7 @@
 package org.jooq.tools;
 
 import static java.lang.Boolean.TRUE;
+import static java.util.Objects.requireNonNull;
```
lukaseder: Another thing we cannot use in Java 6...

jOOQ/src/main/java/org/jooq/impl/ExecuteListeners.java
```diff
@@ -102,7 +102,15 @@ static ExecuteListener get(ExecuteContext ctx) {
...
-                (result = init(result)).add(new LoggerListener());
+                (result = init(result)).add(new LoggerListener(
+                        ctx.settings().getExecuteLoggingBindVariables(),
+                        ctx.settings().isExecuteLoggingAbbreviatedBindVariables(),
+                        ctx.settings().getExecuteLoggingSqlString(),
+                        ctx.settings().getExecuteLoggingResult(),
+                        ctx.settings().getExecuteLoggingResultNumberOfRows(),
+                        ctx.settings().getExecuteLoggingResultNumberOfColumns(),
+                        ctx.settings().getExecuteLoggingRoutine(),
+                        ctx.settings().getExecuteLoggingException()));
```
lukaseder: Does the logger listener really need to be aware of all of these at instantiation time? It will get a reference to `Settings` via the `ExecuteContext.settings()` method, so I don't see the point of this particular change...

jOOQ/src/main/java/org/jooq/tools/LoggerListener.java
```diff
@@ -82,91 +84,121 @@
...
+                          final Log.Level executeLoggingRoutine,
+                          final Log.Level executeLoggingException) {
+        this.executeLoggingBindVariables = requireNonNull(executeLoggingBindVariables);
+        this.executeLoggingAbbreviatedBindVariables = requireNonNull(executeLoggingAbbreviatedBindVariables);
+        this.executeLoggingSqlString = requireNonNull(executeLoggingSqlString);
+        this.executeLoggingResult = requireNonNull(executeLoggingResult);
+        this.executeLoggingResultNumberOfRows = requireNonNull(executeLoggingResultNumberOfRows);
+        this.executeLoggingResultNumberOfColumns = requireNonNull(executeLoggingResultNumberOfColumns);
+        this.executeLoggingRoutine = requireNonNull(executeLoggingRoutine);
+        this.executeLoggingException = requireNonNull(executeLoggingException);
```
lukaseder: See comment above... LoggerListener shouldn't have this new constructor.

jOOQ/src/main/java/org/jooq/tools/LoggerListener.java
```diff
@@ -82,91 +84,121 @@
...
-            if (!log.isTraceEnabled())
-                configuration = abbreviateBindVariables(configuration);
+        // [#2939] Prevent excessive logging of bind variables only in DEBUG mode, not in TRACE mode.
+        if (executeLoggingAbbreviatedBindVariables)
+            configuration = abbreviateBindVariables(configuration);
 
-            String[] batchSQL = ctx.batchSQL();
-            if (ctx.query() != null) {
+        String[] batchSQL = ctx.batchSQL();
+        if (log.isEnabled(executeLoggingSqlString) && ctx.query() != null) {
```
lukaseder: I suggest you inverse these checks. The `null` check is much cheaper than the enabled check, which can get quite hairy, checking several `volatile` fields, depending on the logging provider.

lukaseder:
Regarding your comments:

> The version of Log4J you're using doesn't have isWarnEnabled() or isErrorEnabled(). I used a method that takes a deprecated enum instead, which I'm not super happy about.

I'll check it out.

> Is there a reason Settings isn't using Object.equals and Object.hash? The custom implementation is kind of verbose.

`Object.equals()` and `Object.hashCode()` compare identities, not values. This doesn't make sense in `Settings`, which is essentially a value. The methods are generated in XJC, so don't worry about verbosity.

> This doesn't let you log with an SFL4J marker. That's kind of tricky to pull off, since you also support Log4J and java.util.Logging. I guess let people set a marker in the Settings, then only use it if logging with SFL4J?

What does that mean?

> I treat FATAL as ERROR. Let me know if that's okay.

Where?

jOOQ/src/main/java/org/jooq/Log.java
```diff
@@ -229,6 +239,84 @@
...
+            case DEBUG: return isDebugEnabled();
+            case INFO: return isInfoEnabled();
+            case WARN: return isWarnEnabled();
+            case ERROR: return isErrorEnabled();
+            case FATAL: return isErrorEnabled();
+            default: return false;
+        }
+    }
+
+    default void log(final Object message, final Log.Level level) {
```
lukaseder: I'll reverse the order of arguments here. Seems more consistent to always have the level at the beginning.

jOOQ/src/main/java/org/jooq/conf/Settings.java
```diff
@@ -65,6 +67,22 @@
...
+    @XmlElement(defaultValue = "false")
+    protected Boolean executeLoggingAbbreviatedBindVariables = false;
```
lukaseder: Wrong default. Should be true. (not abbreviating bind variables can explode the logs, e.g. when a bind variable has several mega bytes of LOB data)

jOOQ/src/main/java/org/jooq/conf/Settings.java
lukaseder: I will rename this to `executeLoggingSqlStringInlined`. This isn't about logging bind variables themselves, but about logging the SQL string with inlined bind variables.

jOOQ/src/main/java/org/jooq/tools/LoggerListener.java
```diff
@@ -82,91 +84,121 @@
...
+                for (String sql : batchSQL)
+                    log.log("Executing batch query", newline + sql, executeLoggingSqlString);
         }
     }
 
     @Override
     public void recordEnd(ExecuteContext ctx) {
-        if (log.isTraceEnabled() && ctx.record() != null)
-            logMultiline("Record fetched", ctx.record().toString(), Level.FINER);
+        if (log.isEnabled(executeLoggingResult) && ctx.record() != null)
```
lukaseder: This is too simplistic and rather dangerous! Before, records were usually not logged individually (except on trace level). Now, there's no way to distinguish between logging records and the entire result. This results in way too verbose logging. More flags will be needed.

jOOQ/src/main/java/org/jooq/tools/LoggerListener.java
lukaseder: Java 8 APIs are allowed in sections that are removed in the Java 6 build. But that's only available in the commercial repositories, not the open source ones. It's a minor thing, which can easily be fixed after merging a PR, don't worry about it too much.

Issue #3777, opened 2014-11-18
Deadlock when loading DataType classes


tstack:
We're seeing a deadlock when two different threads load different DataType classes at the same time.  It looks like loading one DataType class causes the others to be loaded as well during class initialization.  But, this can cause a deadlock when multiple threads are trying to load these DataType classes because they each hold a lock on the class and end up waiting for the others to finish.

The following code reproduces the deadlock for me:

``` java
import java.io.*;
import java.net.*;

class jooqlock {
    public static void main(String[] args) throws Exception {
        for (int i = 0; i < 1; i++) {
            doit();
        }
    }

    private static void doit() throws Exception {
        final ClassLoader loader = new URLClassLoader(new URL[] { new File("jooq-3.4.4.jar").toURI().toURL() });

        Runnable oload = new Runnable() {
                public void run() {
                    try {
                        Class.forName("org.jooq.util.sqlite.SQLiteDataType", true, loader);
                    } catch (Throwable th) {
                        th.printStackTrace();
                    }
                }
            };
        Runnable sload = new Runnable() {
                public void run() {
                    try {
                        Class.forName("org.jooq.util.postgres.PostgresDataType", true, loader);
                    } catch (Throwable th) {
                        th.printStackTrace();
                    }
                }
            };

        Thread th1 = new Thread(oload);
        Thread th2 = new Thread(sload);
        th1.start();
        th2.start();
        th1.join();
        th2.join();
    }
}
```

Here is a stack dump from after the deadlock happens:

```
Full thread dump Java HotSpot(TM) 64-Bit Server VM (24.65-b04 mixed mode):

"Thread-1" prio=5 tid=0x00007fd119841800 nid=0x5303 in Object.wait() [0x000000011b619000]
   java.lang.Thread.State: RUNNABLE
    at org.jooq.util.postgres.PostgresDataType.<clinit>(PostgresDataType.java:76)
    at java.lang.Class.forName0(Native Method)
    at java.lang.Class.forName(Class.java:270)
    at jooqlock$2.run(jooqlock.java:27)
    at java.lang.Thread.run(Thread.java:745)

"Thread-0" prio=5 tid=0x00007fd119841000 nid=0x5103 in Object.wait() [0x000000011b515000]
   java.lang.Thread.State: RUNNABLE
    at java.lang.Class.forName0(Native Method)
    at java.lang.Class.forName(Class.java:190)
[...]

PR #7350, opened 2018-03-26
Initialize data types when creating context

victorbr:
@lukaseder This is an attempt to solve #3777 without a circular dependency between `SQLDataType` and dialect-specific data type classes.
This is far from ideal and I'm really not sure that it is safe to do what I'm doing - for you to judge!
Also, the fact that `SQLDialect` is an enum is really unfortunate... It creates a coupling between dialects which better be avoided. I didn't find a better way to solve this however...
Even if this is a promising direction, `TYPES_BY_SQL_DATATYPE` and its friends are better be made thread-safe otherwise we risk a race condition while populating them if several threads are involved.

### DIFF
--- a/jOOQ/src/main/java/org/jooq/SQLDialect.java
+++ b/jOOQ/src/main/java/org/jooq/SQLDialect.java
@@ -38,6 +38,17 @@
 
 package org.jooq;
 
+import org.jooq.impl.SQLDataType;
+import org.jooq.util.cubrid.CUBRIDDataType;
+import org.jooq.util.derby.DerbyDataType;
+import org.jooq.util.firebird.FirebirdDataType;
+import org.jooq.util.h2.H2DataType;
+import org.jooq.util.hsqldb.HSQLDBDataType;
+import org.jooq.util.mariadb.MariaDBDataType;
+import org.jooq.util.mysql.MySQLDataType;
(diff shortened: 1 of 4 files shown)


lukaseder:
This looks quite hacky, I'm afraid. Also, I'm not sure I really want to create a dependency from `SQLDialect` to specific `DataType` classes.

I think this really needs to be fixed thoroughly, or we'll live with the status quo. Hacking around in this area will introduce great risks for little value.

victorbr:
Right, as I said, the central problem is `SQLDialect` being an enum. If each of them were a separate class I don't think there would be a problem for every specific one to reference its types, would it?
Other than that I wouldn't say it's a hack 😄

lukaseder:
> Right, as I said, the central problem is SQLDialect being an enum

No, that's definitely not a central problem. It may be central for the approach you've taken, but it just shows that your approach was wrong.

> Other than that I wouldn't say it's a hack 😄

Any solution that doesn't solve the *actual* problem (the fact that there are global and static caches, instead of caches inside of the `Configuration`) is a hack. In fact, the status quo is a hack, too!

With this in mind, I insist again, we don't worsen the situation by working around the status quo hack via additional hacks, but solve this thoroughly.

PR #7360, opened 2018-03-29
Initialize data types when creating context

victorbr:
Revised version of #7350 , this time without touching the `SQLDialect` enum.
Hope this time it does not look "hacky" 😄

### DIFF
--- a/jOOQ/src/main/java/org/jooq/impl/DefaultDSLContext.java
+++ b/jOOQ/src/main/java/org/jooq/impl/DefaultDSLContext.java
@@ -335,6 +335,9 @@ public class DefaultDSLContext extends AbstractScope implements DSLContext, Seri
 
     public DefaultDSLContext(Configuration configuration) {
         super(configuration, configuration == null ? null : configuration.data());
+        if (configuration != null) {
+            SQLDialectRegistry.initializeDialect(configuration.dialect());
+        }
     }
 
     // -------------------------------------------------------------------------
@@ -4474,19 +4477,6 @@ public class DefaultDSLContext extends AbstractScope implements DSLContext, Seri
         DeleteQuery<R> delete = deleteQuery(record.getTable());
         delete.addConditions(condition);
         return delete.execute();
-    }
-
-    // -------------------------------------------------------------------------
-    // XXX Static initialisation of dialect-specific data types
-    // -------------------------------------------------------------------------
-
-    static {
-        // Load all dialect-specific data types
-        // TODO [#650] Make this more reliable using a data type registry
-
-        try {
-            Class.forName(SQLDataType.class.getName());
-        } catch (Exception ignore) {}
     }
 
     // -------------------------------------------------------------------------
--- a/jOOQ/src/main/java/org/jooq/impl/DefaultDataType.java
+++ b/jOOQ/src/main/java/org/jooq/impl/DefaultDataType.java
@@ -233,12 +233,6 @@ public class DefaultDataType<T> implements DataType<T> {
         }
 
         SQL_DATATYPES_BY_TYPE = new LinkedHashMap<Class<?>, DataType<?>>();
-
-        // [#2506] Transitively load all dialect-specific data types
-        try {
-            Class.forName(SQLDataType.class.getName());
-        }
-        catch (Exception ignore) {}
     }
 
     public DefaultDataType(SQLDialect dialect, DataType<T> sqlDataType, String typeName) {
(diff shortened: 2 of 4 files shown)


lukaseder:
Victor! This is crazy. How is initialising these classes from a random `DefaultDSLContext` instance not a hack? Please stop sending PRs for this. The correct solution is a data type registry inside of the `Configuration`, but not in any way you seem to think right now. I'll fix this eventually. For the time being, please just make sure the jOOQ generated code is initialised from a single thread, not from two concurrent threads.

victorbr:
@lukaseder I'm sorry if I trespassed over sacred jOOQ boundaries but please believe me I'm sincerely trying to help! It's just the things that are obvious to you are not as nearly obvious to me! For instance, I assumed an instance of `DSLContext` is bound to a specific dialect, so I thought it's actually natural to initialize dialect-specific classes there. I believe you have a good reason to call it a "random instance" but I thought this is where it all starts for the jOOQ user - it is the first time they specify which dialect they work with - so I thought it actually makes sense to do the initialization there.
Apparently I'm mistaken but believe me it's out of ignorance and not out of desire to spam you with irrelevant PRs! I'm only trying to help..

lukaseder:
Hey Victor, this is nothing personal. It just takes time to review PRs and PRs for complex stuff has had a 20% success chance in the past. So, you're in good company :-)

The best help in this particular case would be to simply trust me and accept the fact that a thorough fix of the mess that is the status quo is very difficult, and any attempt to work around the deadlock you've experienced is a workaround in a very delicate area, and thus not a good idea due to the tremendous regression risk. You don't have the integration tests, so it is impossible for you to verify if any of your suggested fixes are actually going to work well (in this case).

Please be patient. Eventually, I will fix this thoroughly, but given the fact that this has been happening so rarely over the past years, I am not willing to prioritise this fix right now.

lukaseder:
Side note: In a way, the real blame here goes to the github user experience. It should be impossible to create a PR without prior permission by the maintainer.

victorbr:
Thanks Lukas, I'm glad we talked because I had a really bad feeling about this.

Yeah, I can see this is a very delicate area with quite nasty workarounds in place - just look at the lines deleted in this PR! I completely understand your concern about possible regressions.

The reason I dived into this was an hour of production outage of our system caused by the deadlock. In the end we had to explicitly call `Class.forName` for `PostgresDataType` to make sure there's no collision between threads (we work with Guice and its initialization is multi-threaded). So I tried to understand the underlying reason and to find a better solution.

Trust you to get to this some day - if by chance I could help I'd be happy to!

lukaseder:
I'm sorry about that production outage. I absolutely understand how scary such a situation is. After all, deadlocks are non-deterministic events and can happen anytime. Yet, again, this problem doesn't appear to happen very often in general, so I'm really curious what actually causes it. I don't mean what classes are initialised in the wrong order, but why this happens *to you* in the first place. Does your own code, or your generated code reference the `PostgresDataType` class by any chance? If so, would you mind sharing the relevant bits? I'm assuming that by removing this reference, we can work around the issue *in your case*. I will spend a bit more time next week thinking about another workaround that is more thorough than the existing ones. After all, even if this is fixed thoroughly through an actual type registry, that fix cannot be downported to older versions. A workaround could be.

victorbr:
It's OK, with that outage it's more our deployment system that is to blame - the failure to start up should not have caused the outage. Never mind!
Yes, indeed, we reference `PostgresDataType` - for example for the `JSON` datatype which is unique to Postgres. While in other places we use the default `SQLDataType` directly. And this deadlocks pretty much with 100% chance as shown in my comment [here](https://github.com/jOOQ/jOOQ/issues/3777#issuecomment-374224056).
What is the general recommendation - to use the generic types or the dialect-specific ones? We are sort of unsure, perhaps this is where the problem comes from.
I personally am not a big fan of generated code in general so I prefer to define the table classes "by hand".

lukaseder:
I see - I have regretted the premature addition of that `PostgresDataType.JSON` value in the past. It is rather unusual to actually use those dialect specific `DataType` references directly, which is probably the reason why this dead lock happens so rarely in user code.

I suspect the best way forward for you right now is to avoid using `PostgresDataType`, and resort to using data type converters or bindings instead:

- https://www.jooq.org/doc/latest/manual/code-generation/custom-data-types
- https://www.jooq.org/doc/latest/manual/code-generation/custom-data-type-bindings

These work the same way, regardless if you're using the code generator or not.

lukaseder:
... in the meantime, I'll think about how these `PostgresDataType` et al classes can be deprecated in the future.

lukaseder:
I've verified various usages of the `<dialect>DataType` classes throughout the integration test code base, including generated code, and there's really no reason why this type should be referenced directly. I'll mark the types as deprecated: #7375

Their usage is no longer recommended. Instead of using `PostgresDataType.JSON`, please use a custom data type and a binding.

PR #7497, opened 2018-05-17
Specify capacity for internal ArrayLists whose size is known

twilson-palantir:


lukaseder:
Thank you very much for your contribution. I agree with the general idea of improving this. Will look through each individual change and merge afterwards

jOOQ-codegen/src/main/java/org/jooq/codegen/JavaWriter.java
```diff
@@ -202,7 +202,7 @@ protected String beforeClose(String string) {
 
     @Override
     protected List<String> ref(List<String> clazz, int keepSegments) {
-        List<String> result = new ArrayList<String>();
+        List<String> result = new ArrayList<String>(clazz == null ? 0 : clazz.size());
```
lukaseder: Do note that `new ArrayList()` and `new ArrayList(0)` do not behave the same way once the list grows. Irrelevant in this case, but worth mentioning

lukaseder:
Btw, how did you find all of these instances? Using some static code analysis tool?

jOOQ/src/main/java/org/jooq/impl/AbstractField.java
```diff
@@ -947,7 +947,7 @@ public final Condition in(Field<?>... values) {
 
     @Override
     public final Condition in(Collection<?> values) {
-        List<Field<?>> fields = new ArrayList<Field<?>>();
+        List<Field<?>> fields = new ArrayList<Field<?>>(values.size());
```
lukaseder: This can be taken one step further, because the list is going to be converted to an array again, later. So why not just use an array, instead?

jOOQ/src/main/java/org/jooq/impl/AbstractRoutine.java
```diff
@@ -1606,7 +1606,7 @@ public void accept(Context<?> ctx) {
             RenderContext local = create(ctx).renderContext();
             toSQLQualifiedName(local);
 
-            List<Field<?>> fields = new ArrayList<Field<?>>();
+            List<Field<?>> fields = new ArrayList<Field<?>>(getInParameters().size());
```
lukaseder: Same here, this could be an array, rather than an arraylist

jOOQ/src/main/java/org/jooq/impl/BatchSingle.java
```diff
@@ -200,7 +200,7 @@ private final void checkBindValues() {
         // [#4062] Make sure we collect also repeated named parameters
         ParamCollector collector = new ParamCollector(configuration, false);
         collector.visit(query);
-        List<Param<?>> params = new ArrayList<Param<?>>();
+        List<Param<?>> params = new ArrayList<Param<?>>(collector.resultList.size());
```
lukaseder: Another candidate for an array (I'll just fix these. Further comments will be implied)

jOOQ/src/main/java/org/jooq/impl/Tools.java
```diff
@@ -1421,11 +1421,12 @@ else if (value instanceof QueryPart)
      *         values created from the argument objects.
      */
     static final <T> List<Field<T>> fields(T[] values) {
-        List<Field<T>> result = new ArrayList<Field<T>>();
+        int length = values == null ? 0 : values.length;
```
lukaseder: This null check will lead to potential null pointer access warnings further down. I'll change this into:

```java
if (values == null)
    return new ArrayList<Field<T>>();
```

lukaseder:
Excellent, I've reviewed and can approve each single change. Thanks a lot for doing this!

The hassle of keeping in mind exact sizes of arrays / lists would be remedied by using the Java 8 Stream API - if only we could upgrade our internals to Java 8...

twilson-palantir:
I just used Intellij to find callers of `new ArrayList<>()` and audited each one.

I knew some (many?) of these could probably just be arrays, but I wanted to keep things simple and particular did not want to have to consider whether the commercial version might be adding additional elements. I completely support replacing these with raw arrays where appropriate.

lukaseder:
> I just used Intellij to find callers of new ArrayList<>() and audited each one.

Thanks for the hint, and thanks again for doing this!

> I knew some (many?) of these could probably just be arrays, but I wanted to keep things simple and particular did not want to have to consider whether the commercial version might be adding additional elements.

Sure, no worries

PR #7578, opened 2018-06-13
Spring boot example test repeats identical assertion

bbaumgartner:
I guess it should assert rollback1 and rollback2 instead.


lukaseder:
Sharp eye, thank you very much for the contribution!

PR #7652, opened 2018-07-07
Upgrade to maven-compiler-plugin 3.7.1-SNAPSHOT

sormuras:
Prior to this commit, the `maven-compiler-plugin` **3.7.0** places jars on the class path which need to be placed on the module path. See errors at https://travis-ci.org/jOOQ/jOOQ/builds/400924646#L2003

Using `maven-compiler-plugin` **3.7.1-SNAPSHOT** solves this issue.

Related bug: https://issues.apache.org/jira/browse/MCOMPILER-336

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -174,6 +174,13 @@
         </dependencies>
     </dependencyManagement>
 
+    <pluginRepositories>
+        <pluginRepository>
+            <id>apache.snapshots</id>
+            <url>http://repository.apache.org/snapshots/</url>
+        </pluginRepository>
+    </pluginRepositories>
+
     <build>
         <finalName>${project.artifactId}-${project.version}</finalName>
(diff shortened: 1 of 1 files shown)


lukaseder:
Oh, that's very interesting. Cool, thank you so much for researching this! Still interesting that the build works locally (for me) but not on travis...

sormuras:
> ... the build works locally ...

Had a similar working run, but only once. Thereafter, all build runs failed due the classpath/modulepath issue.

lukaseder:
> Had a similar working run, but only once. Thereafter, all build runs failed due the classpath/modulepath issue.

Hmm indeed, I've noticed something similar, recently:

- `mvn clean install` ran fine
- `mvn test` (right after, no clean) didn't

Sounds similar to your case

Issue #5218, opened 2016-04-23
Add support for FOR SHARE OF in Postgres dialect


kbalyasnikov:
At the moment there is no support for this syntax in jOOQ. The syntax is same as for FOR UPDATE and described in [official documentation](http://www.postgresql.org/docs/current/static/sql-select.html#SQL-FOR-UPDATE-SHARE).

lukaseder:
Thank you very much for your suggestion. That certainly makes sense. For the time being, this can be emulated by using `FOR UPDATE` instead, and replacing the `FOR UPDATE` clause in the generated SQL string by `FOR SHARE` via a `ExecuteListener.renderEnd()` event

kbalyasnikov:
@lukaseder, thanks for accepting suggestion!

knutwannheden:
I may be missing something, but `SelectForUpdateStep` has had a `forShare()` method for a very long time already (oldest Git revision has it as well: https://github.com/jOOQ/jOOQ/blob/5666cfd019b650f117d61672de7cd366e024befc/jOOQ/src/main/java/org/jooq/SelectForUpdateStep.java).

However, I noticed that the method doesn't return `SelectForUpdateOfStep` as for example `forUpdate()` does. As a result the `FOR SHARE` clause doesn't allow the `NOWAIT` option. This is wrong and should be corrected and we could do that in the scope of this issue.

knutwannheden:
I probably missed the `OF` in the issue title. That would also be resolved by changing the return type of `forShare()` as described.

knutwannheden:
OK, wasn't quite as trivial as I thought. The `SelectQuery` API is rather different here and would require the addition of a set of `setForShareOf()` methods as well as `setForShareWait()` and `setForShareNoWait()` methods and even a `setForShareSkipLocked()` method.

knutwannheden:
The DSL API has been changed as proposed and the `SelectQuery` now also supports specifying the `OF` and `WAIT | NOWAIT | SKIP LOCKED` clauses for all lock modes. Whether a given combination of lock mode and flags makes sense, depends on the SQL dialect.

PR #7690, opened 2018-07-26
Added support for additional postgresql 9.3+ locking modes

jmgrimes:
Implemented postgresql 9.3+ SELECT FOR NO KEY UPDATE support via forUpdate().noKey().
Implemented postgresql 9.3+ SELECT FOR KEY SHARE support via forShare().key().

My team had some issues with Postgres's FOR UPDATE locking inserts and updates on foreign key rows due to the possibility of a "FOR UPDATE" updating the primary key.

SELECT FOR NO KEY UPDATE is Postgres's slightly less pessimistic locking mode to prevent this.

Likewise, SELECT FOR KEY SHARE allows FOR SHARE semantics on a row locked by "FOR NO KEY UPDATE".

Figured I'd just take a shot at implementing this and sending a PR.

### DIFF
--- a/jOOQ/src/main/java/org/jooq/SelectForShareKeyStep.java
+++ b/jOOQ/src/main/java/org/jooq/SelectForShareKeyStep.java
@@ -1,0 +1,10 @@
+package org.jooq;
+
+import static org.jooq.SQLDialect.POSTGRES_9_3;
+
+public interface SelectForShareKeyStep<R extends Record> extends SelectOptionStep<R> {
+
+    @Support({ POSTGRES_9_3 })
+    SelectOptionStep<R> key();
+
+}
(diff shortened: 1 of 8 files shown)


lukaseder:
Thank you very much for your suggestion. I agree that we could support this feature, see also #5218. However, I don't agree with the suggested API. Why not just add straightforward method names like `forNoKeyUpdate()`? We're not going to save much by reusing keywords like `forUpdate()` and then turning around keywords in a way that makes them harder to discover.

I will reject this PR and create a separate issue for this feature: #7705

----
See also: #5218

lukaseder:
Duplicate of #7705

lukaseder:
#7705 implemented for jOOQ 3.12.0

jmgrimes:
Sounds great, thank you!

Issue #7764, opened 2018-08-16
Race condition in cached ImmutablePOJOMapperWithParameterNames


llucherini:
### Expected behavior and actual behavior:
When `RecordMapper`s caching is turned on, which is the default configuration, race conditions in `DefaultRecordMapper.ImmutablePOJOMapperWithParameterNames` lead to inconsistent results when using `<E> E fetchOneInto(Class<? extends E> type)` and potentially other methods.
### Steps to reproduce the problem (include example code if possible):
I've created a small project to replicate the issue [here](https://github.com/llucherini/jOOQ-race-condition-poc). it can be run simply by issuing `gradle run`.
### Versions:

- jOOQ: 3.11.3
- Java: 10.0.1
- Database (include vendor): H2 1.4.197 and Postgres on Amazon RDS
- OS: Ubuntu 16.04.1
- JDBC Driver (include name if unofficial driver): h2 1.4.197 and postgres 42.2.1

lukaseder:
Fixed for jOOQ 3.12. Backports scheduled for 3.11.5 (#7777) and 3.10.9 (#7778)

lukaseder:
Thanks again for your helpful report and test case!

llucherini:
No worries! Happy to be of help!

PR #7765, opened 2018-08-16
[#7764] Race condition in DefaultRecordMapper 

llucherini:
`parameterValues` is shared between threads and can be concurrently modified, causing the inconsistencies described by the issue.


lukaseder:
Thank you very much for your report and for your suggested fix. I can reproduce the problem and confirm your fix

Issue #6071, opened 2017-04-13
Add Name Sequence.getQualifiedName() and getUnqualifiedName()


lukaseder:
See also: https://github.com/jOOQ/jOOQ/issues/6068

lukaseder:
Change committed in https://github.com/jOOQ/jOOQ/commit/294b06ac1e41ee6c7fa8ef408ddfe5a6ad176807

Issue #5997, opened 2017-03-22
Field.as(Field) and Table.as(Table) should retain quotation flag from Name


lukaseder:
If the argument `Field` of an aliasing call is a `field(Name)`, we must retain the newly introduced quotation flag from `DSL.quotedName()` or `DSL.unquotedName()` (see #5993)

Issue #5996, opened 2017-03-22
Add Field.as(Name) and Table.as(Name), Table.as(Name, Name...)


lukaseder:
Users who want to alias fields or tables with `DSL.unquotedName()` (see #5993) currently cannot do so, as the existing `Field.as(String)` and `Table.as(String)` methods will always produce `DSL.name()` behind the scenes (quoted identifiers).

The only option is to use a `RenderNameStyle` (or jOOQ 3.10 `Settings.renderQuotedNames == false`) but that operates globally, not on a per-alias basis.

These new methods allow for overriding the quoting on a per-alias basis.

lukaseder:
The code generator needs to be adapted as well

PR #7851, opened 2018-09-10
In Coerce use the qualified name so that fields can be compared again

dhs3000:
The Coerce'ed field renders the full qualified name in SQL, but the field-name is used for comparison, so the name should match the rendered SQL.

After updating to jOOQ 3.11.4 some of our query creation code fails as it compares Coerce'ed table fields and they are not equal anymore (as they where before). With the proposed change the code works again as expected.

### DIFF
--- a/jOOQ/src/main/java/org/jooq/impl/Coerce.java
+++ b/jOOQ/src/main/java/org/jooq/impl/Coerce.java
@@ -54,7 +54,7 @@ final class Coerce<T> extends AbstractField<T> {
     private final Field<?>    field;
 
     public Coerce(Field<?> field, DataType<T> type) {
-        super(field.getUnqualifiedName(), type);
+        super(field.getQualifiedName(), type);
 
         this.field = (field instanceof Coerce) ? ((Coerce<?>) field).field : field;
     }


lukaseder:
Thanks for your report and for the suggested fix. Would you mind showing a test case that worked before (with what version) and fails now?

dhs3000:
I created a test case for this which fails with v3.11.4 but works with v3.6.2 (in Scala as I don't have a Java project up and running). 

In the real project we create 'copy' statements and and replace some columns with new columns before the query is finally created, this fails as some coerced columns cannot be found anymore.

```scala
package jooqfieldequals

import org.jooq._
import org.jooq.impl._
import org.scalatest._

class FieldEqualsTest extends FunSpec with MustMatchers {

  describe("Fields can be compared") {
    import jooqfieldequals.FieldEqualsTest.Schema.TestTable._

    it("when it is the same field") {
      val field1 = TEST_TABLE.INT_FIELD
      val field2 = TEST_TABLE.INT_FIELD

      field1 must equal(field2)
    }

    it("when it is the same field but one is coerced to a different type") {
      val field1 = TEST_TABLE.INT_FIELD
      val field2 = TEST_TABLE.INT_FIELD.coerce(classOf[java.lang.Long])

      field1 must equal(field2) // <-- This works with v3.6.2, but fails with v3.11.4
    }

    it("when it is two different fields") {
      val field1 = TEST_TABLE.INT_FIELD
      val field2 = TEST_TABLE.LONG_FIELD

      field1 must not equal field2
    }
  }

}

// ----------------------------------------------

object FieldEqualsTest {

  object Schema {
    object TestTable {
      val TEST_TABLE = new TestTable()
    }

    class TestTable(alias: String = "TEST_TABLE") extends TableImpl[TestTableRecord](alias) {

      override def getRecordType: Class[TestTableRecord] = classOf[TestTableRecord]

      val INT_FIELD: TableField[TestTableRecord, java.lang.Integer] =
        createField("INT_FIELD", org.jooq.impl.SQLDataType.INTEGER, "")
      val LONG_FIELD: TableField[TestTableRecord, java.lang.Long] =
        createField("LONG_FIELD", org.jooq.impl.SQLDataType.BIGINT, "")
    }

    class TestTableRecord
        extends UpdatableRecordImpl[TestTableRecord](TestTable.TEST_TABLE)
        with Record2[java.lang.Integer, java.lang.Long] {

      override def fieldsRow: Row2[java.lang.Integer, java.lang.Long] =
        super.fieldsRow.asInstanceOf[Row2[java.lang.Integer, java.lang.Long]]
      override def valuesRow: Row2[java.lang.Integer, java.lang.Long] =
        super.valuesRow.asInstanceOf[Row2[java.lang.Integer, java.lang.Long]]
      override def field1: Field[java.lang.Integer] = TestTable.TEST_TABLE.INT_FIELD
      override def field2: Field[java.lang.Long] = TestTable.TEST_TABLE.LONG_FIELD
      override def value1: java.lang.Integer = throw new NotImplementedError
      override def value2: java.lang.Long = throw new NotImplementedError
      override def value1(value: java.lang.Integer): TestTableRecord = throw new NotImplementedError
      override def value2(value: java.lang.Long): TestTableRecord = throw new NotImplementedError
      override def values(value1: java.lang.Integer, value2: java.lang.Long): TestTableRecord = throw new NotImplementedError
    }
  }
}
```

lukaseder:
Thank you very much for the example. I'll investigate. The changes that might be involved in a regression here are:

### In `Coerce.java` (constructor)

- #5996
- #5997

### In `AbstractField.java` (equals method)

- #6071 (moved `AbstractField.equals()` to `AbstractNamed.equals()`), which now checks the fully qualified name, rather than just the field name. This might be the problem.

dhs3000:
Yes, sound as they could be involved. It sounds right to me to compare the fully qualified field names and not just the field name.
The above test case (and our other tests) work when using `field.getQualifiedName()` instead of `field.getUnqualifiedName()` for the `name` in `Coerce` constructor. The name is used in `AbstractNamed.equals()` for the comparison.

dhs3000:
@lukaseder Could you find anything out related to this issue?

lukaseder:
I'm sorry for the delay.

I have re-considered this change. True to the nature of the "coerce" operation (which is not reflected in SQL but purely in client logic), I agree that the behaviour you're expecting is reasonable. Just the same, you could do a coercion manually by re-creating a qualified column name and attaching a different data type to it:

```java
Field<String> field1 = A.B.C;
Field<Integer> field2a = field1.coerce(Integer.class);
Field<Integer> field2b = DSL.field(field1.getQualifiedName(), Integer.class);
```

It is safe to expect `field2a` and `field2b` to be the same.

lukaseder:
While I do agree that the qualified name should be present, I'm reluctant to backport this to jOOQ 3.11 given that the change may again be seen as a breaking change if someone relies on the qualified name of a coerced field to be effectively unqualified.

PR #7939, opened 2018-10-13
Fix Javadoc entry for Field#containsIgnoreCase

### LINKED ISSUE #7938: Incorrect Javadoc entry for Field#containsIgnoreCase
I came across a slip-up in the Javadoc for `Field#containsIgnoreCase(Field<String> value)`:

https://github.com/jOOQ/jOOQ/blob/5a958a8df9843d8b2cbe58d41e6664b417b78d63/jOOQ/src/main/java/org/jooq/Field.java#L1674-L1690

This method is described as: _translating to `this not ilike ...`_ while it should be: `this ilike ...`, just like the `#likeIgnoreCase(String, char)` it is conceptually similar to.

That's just a problem with the documentation: the actual implementation in `AbstractField` is fine.

I've had a thorough look for other potential mishaps of the same sort in that same interface but couldn't find any. I hope it's not too insignificant for a PR...

ccjmne:
This fixes #7938.

---

The SQL-equivalent part of this method's documentation was wrongly indicating that it:
> translates to `this not ilike` ... in POSTGRES, or to `lower(this) not like lower` ... in all other dialects.

---

I removed the `not`.

Could not find any other similar mistake – I'm pretty sure that was an isolated case.

### DIFF
--- a/jOOQ/src/main/java/org/jooq/Field.java
+++ b/jOOQ/src/main/java/org/jooq/Field.java
@@ -1676,9 +1676,9 @@ extends
      * proper adding of wildcards and escaping.
      * <p>
      * This translates to
-     * <code>this not ilike ('%' || escape(value, '\') || '%') escape '\'</code>
+     * <code>this ilike ('%' || escape(value, '\') || '%') escape '\'</code>
      * in {@link SQLDialect#POSTGRES}, or to
-     * <code>lower(this) not like lower(('%' || escape(value, '\') || '%') escape '\')</code>
+     * <code>lower(this) like lower(('%' || escape(value, '\') || '%') escape '\')</code>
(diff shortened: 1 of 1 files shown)


ccjmne:
I did sign the [Transfer of Rights Agreement](http://www.jooq.org/legal/contributions).
I did read the [contributing guidelines](https://github.com/jOOQ/jOOQ/blob/master/CONTRIBUTING.md) but figured adding my name to the ABOUT.txt was largely overkill.

lukaseder:
Thanks for your suggested fix. I totally agree with it.

> I did read the contributing guidelines but figured adding my name to the ABOUT.txt was largely overkill.

Sure, no worries

Issue #8090, opened 2018-12-05
XMLDatabase does not support array types for postgres dialect


Klobersaurus:
### Expected behavior and actual behavior:
Expected behavior: varchar[] types get generated as String[]
Actual behavior: varchar[] types get generated as Object

#### Sample column
```
<column>
    <table_catalog></table_catalog>
    <table_schema>my_schema</table_schema>
    <table_name>my_table</table_name>
    <column_name>my_column</column_name>
    <data_type>ARRAY</data_type>
    <character_maximum_length>0</character_maximum_length>
    <numeric_precision>0</numeric_precision>
    <numeric_scale>0</numeric_scale>
    <ordinal_position>4</ordinal_position>
    <is_nullable>false</is_nullable>
    <comment>Text representation of the type of an action.</comment>
</column>
```
#### Java code to set up configuration
```
public Configuration build() {
	final Generator generator = new Generator();
	generator.setName(JavaGenerator.class.getName());

	final Database database = new Database();
	database.setName(XMLDatabase.class.getName());
	
	//Dynamically specify schemas via SPI
	final List<Schema> schemata = new ArrayList<Schema>();
	JooqCodeGenSchemaService.instance().getAllProviders()
			.forEach(provider -> schemata.addAll(provider.getSchemata()));
	database.setSchemata(schemata);

	List<Property> dbProperties = new ArrayList<Property>();
	
	Property xmlFileProp = new Property();
	xmlFileProp.setKey("xml-file");
	xmlFileProp.setValue("test.xml");
	dbProperties.add(xmlFileProp);
	
	Property xmlDialectProp = new Property();
	xmlDialectProp.setKey("dialect");
	xmlDialectProp.setValue(SQLDialect.POSTGRES.getName());
	dbProperties.add(xmlDialectProp);
	
	database.setProperties(dbProperties);
	
	//Dynamically load forced types via SPI
	final List<ForcedType> forcedTypes = new ArrayList<ForcedType>();
	JooqForcedTypeService.instance().getAllProviders()
			.forEach(provider -> forcedTypes.addAll(provider.getForcedTypes()));
	database.setForcedTypes(forcedTypes);
	generator.setDatabase(database);

	final Generate generate = new Generate();
	generate.setDaos(true);
	generate.setInterfaces(true);
	generate.setSerializableInterfaces(true);
	generate.setJpaAnnotations(false);
	generate.setValidationAnnotations(true);
	generator.setGenerate(generate);
	
[...]

PR #8101, opened 2018-12-08
[#8090] - xmldatabase array support

PatrickJSweeney:
to be utilized from XMLGenerator and XMLTableDefinition

https://github.com/jOOQ/jOOQ/issues/8090

### DIFF
--- a/jOOQ-codegen/src/main/java/org/jooq/codegen/XMLGenerator.java
+++ b/jOOQ-codegen/src/main/java/org/jooq/codegen/XMLGenerator.java
@@ -165,6 +165,7 @@ public class XMLGenerator extends AbstractGenerator {
                         column.setCharacterMaximumLength(type.getLength());
                         column.setColumnDefault(type.getDefaultValue());
                         column.setDataType(type.getType());
+                        column.setUdtName(type.getUserType());
                         if (co.isIdentity())
                             column.setIdentityGeneration("YES");
                         column.setIsNullable(type.isNullable());
--- a/jOOQ-codegen/src/main/java/org/jooq/codegen/XMLGenerator.java
+++ b/jOOQ-codegen/src/main/java/org/jooq/codegen/XMLGenerator.java
@@ -383,6 +383,7 @@ public class XMLGenerator extends AbstractGenerator {
         else {
             routine.setRoutineType(RoutineType.FUNCTION);
             routine.setDataType(r.getReturnType().getType());
+            routine.setTypeUdtName(r.getReturnType().getUserType());
             routine.setCharacterMaximumLength(r.getReturnType().getLength());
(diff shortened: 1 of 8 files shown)


PatrickJSweeney:
@lukaseder I need this fix under jdk8 for my personal needs.  It cherry-picked cleanly back to v3.11.7.  Please let me know if there is anything else I can do.

PatrickJSweeney:
K.  I promise I'm done with the force-pushes now.  I was just trying to match the https://github.com/jOOQ/jOOQ/blob/master/CONTRIBUTING.md

lukaseder:
Thank you for offering your help with this task, @PatrickJSweeney. Unfortunately, I must reject this PR.

The SQL standard `INFORMATION_SCHEMA` specifies a variety of columns in the `INFORMATION_SCHEMA.COLUMNS` view, including `(UDT_CATALOG, UDT_SCHEMA, UDT_NAME)`, which are derived from the `DEFINITION_SCHEMA.DATA_TYPE_DESCRIPTOR` table. If we do decide to add support for arrays, it is crucial to specify their contents in a new standards compliant `INFORMATION_SCHEMA.ELEMENT_TYPES` table.

PatrickJSweeney:
@lukaseder  I don't think I understood your original comments in #8090, thus my incomplete implementation.  I thought referencing the bare minimum udt columns from information_schema would be least invasive.

Could you supply a link to the spec you are referencing?  My google searches do not yield any results that line up with your description.  I may want to propose a number of improvements to XMLDatabase and would like to avoid any future miscommunication about your guidance on the implementation.

PatrickJSweeney:
@lukaseder After some digging, I see that the SQL standard is not free (or cheap for an individual hobbyist). Disregard my previous request.  If I do more work in this area, I'll reference your implementation for this bug as the pattern to follow and hope for the best.

lukaseder:
@PatrickJSweeney I'm sorry for the confusion. The point I'm trying to mainly hint at here is that changing things in jOOQ is not as simple as it may look at first. While I do appreciate the intent, usually I prefer doing a change myself rather than discussing the various requirements of a possible PR, which is both more time consuming for everyone and far from a guarantee to a better result.

In addition to this, from a project and product management perspective, one would not expect that a feature can be implemented so quickly without deeper knowledge of the product internals and/or roadmap, and this doesn't touch the topic of prioritisation. Here, we have two edge case features: arrays and the `XMLDatabase`. Fixing this quickly will deteriorate the quality of the product in an already edge casey area. Fixing this thoroughly will be very time consuming and definitely not something that could be expected from an external contributor.

I just wish github would allow for turning on the issue tracker feature without also necessarily turning on the pull request feature. The workflow should always be: Discuss first, and then probably, not implement later :-)

Issue #8004, opened 2018-11-01
Handle edgecase with Kotlin boolean types


amir20001:
### Steps to reproduce the problem (include example code if possible):

`Foo(val id: String, var isDeleted: Boolean)`

Will use ImmutablePOJOMapperWithParameterNames and try to map to a field called `isDeleted` but the field should be `deleted`

### Versions:

- jOOQ: 3.11
- Java: 8
- Database (include vendor): MySql
- OS: 10.13.6
- JDBC Driver (include name if inofficial driver):

lukaseder:
Thank you very much for your request.

I'm curious: Based on what specific convention do you think that the `is` prefix should somehow be recognised by jOOQ as not being part of the attribute name? jOOQ can recognise it on getters, as that has been common practice for many years according to the JavaBeans convention, but you're not referring to a getter, you're referring to an attribute...

amir20001:
Thanks for getting back to me.
Just to make sure we are on the same page.
 `Foo(val id: String, var isDeleted: Boolean)`

will generate a java class equivalent to 

`
public class Foo {

    private String id;
    private Boolean deleted;

    public Foo(String id, Boolean isDeleted) {
        this.id = id;
        this.deleted = deleted;
    }

    public String getId() {
        return id;
    }

    public Foo setId(String id) {
        this.id = id;
        return this;
    }

    public Boolean isDeleted() {
        return deleted;
    }

    public Foo setDeleted(Boolean deleted) {
        this.deleted = deleted;
        return this;
    }
}
`

The problem that we are running into if im not mistaken, we are reading the property names from the constructor, which is what causes the mismatch.

amir20001:
Unfortunately i can not find anything in the reference guide. All I can find are some stackoverflow posts 
https://stackoverflow.com/questions/43317026/kotlin-naming-convention-for-boolean-returning-methods

vojtapol:
Hi Luke,

This is a widely used feature of Kotlin. 

The corresponding documentation is here:
https://kotlinlang.org/docs/reference/java-interop.html#getters-and-setters

As you know, Kotlin generates getters and setters when you define a class using this syntax:

```kotlin
data class User(val name: String, val age: Int)
```

This generates `getName()` and `getAge()` methods in the bytecode.

The problem is though that often but not always, boolean getters in Java start with `is` instead of `get`. JetBrains had to come up with a way to let users choose whether they want the generated getter to start with `is` or with `get` while maintaining the concise syntax.

[...]

PR #8114, opened 2018-12-13
[#8004] Handle edgecase for Kotlin boolean type

amir20001:
Fixes: #8004

trim Kotlin property names to handle edgecase for booleans

### DIFF
--- a/jOOQ/src/main/java/org/jooq/impl/DefaultRecordMapper.java
+++ b/jOOQ/src/main/java/org/jooq/impl/DefaultRecordMapper.java
@@ -402,7 +402,9 @@ public class DefaultRecordMapper<R extends Record, E> implements RecordMapper<R,
 
                     for (int i = 0; i < parameterTypes.length; i++) {
                         Reflect parameter = Reflect.on(parameters.get(i));
-                        parameterNames.add(parameter.call("getName").<String>get());
+                        // [#8004] Clean up kotlin field name for boolean types
+                        String name = parameter.call("getName").<String>get();
+                        parameterNames.add(getPropertyName(name));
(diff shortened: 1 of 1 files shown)


jOOQ/src/main/java/org/jooq/impl/DefaultRecordMapper.java
```diff
@@ -402,7 +402,9 @@ public int compare(Constructor<E> c1, Constructor<E> c2) {
 
                     for (int i = 0; i < parameterTypes.length; i++) {
                         Reflect parameter = Reflect.on(parameters.get(i));
-                        parameterNames.add(parameter.call("getName").<String>get());
+                        // [#8004] Clean up kotlin field name for boolean types
+                        String name = parameter.call("getName").<String>get();
+                        parameterNames.add(getPropertyName(name));
```
lukaseder: Thank you very much for your suggestion. `Tools.getPropertyName()` extracts the property name `xyz` from getters of the form `isXyz()` and `getXyz()` as well as from setters of the form `setXyz()`. This seems to have the unwanted side effect of translating too many potential property names. Also, surely we should add a check to make sure this is about a boolean type. I assume that a property of the form `isXyz: String` does not follow the same rules in Kotlin?

amir20001: thanks for the feedback, i'll go ahead and make some updates to my pr

vojtapol:
@lukaseder, what is missing from this PR at the moment?

lukaseder:
Thanks for the ping. This looks fine to me. Will merge soon and validate with some integration tests on our side.

Issue #8139, opened 2018-12-21
scala.UninitializedFieldError in runtime using generated data objects


astrofed:
### Steps to reproduce the problem:
Using org.jooq.codegen.ScalaGenerator there are generated objects and classes for database objects (schema, catalogs, tables, records, etc). Then instantiating these objects in runtime the next error is excepted:

Caused by: scala.UninitializedFieldError: Uninitialized field: .../db/generated/Public.scala: 28

### Investigation results
It's caused by the fact that class Table has an argument in constructor Schema that's called from object Public (val PUBLIC = new Public), but constructor of Public contains reference on Catalog which called from object DefaultCatalog, but during initialization of DefaultCatalog in the body of this class there should be initialized val PUBLIC = com.opso.db.generated.Public.PUBLIC. As we can see there is a cross-reference and initialization loop. 

Looks like it's better to use **lazy val** instead **val** in scala case.

### Versions:
- jOOQ : 3.11.7
- Scala: 2.12.8
- Database (include vendor): PostgreSQL 11
- JDBC Driver (include name if inofficial driver): org.postgresql.Driver (42.2.5)

astrofed:
Looks like https://github.com/jOOQ/jOOQ/issues/5489 has the same root

lukaseder:
Thanks for your report. Our integration tests work smoothly with Scala 2.12.8, but perhaps there's an edge case that cannot be reproduced so easily? Would it be possible to provide an MCVE (template here: https://github.com/jOOQ/jOOQ-mcve) that helps reproduce the issue?

lukaseder:
Unfortunately, we couldn't reproduce this issue in our integration tests so far. jOOQ 3.14 will have quite a few changes / improvements that were made to the `ScalaGenerator` in the context of the implementation of the new `KotlinGenerator`: https://github.com/jOOQ/jOOQ/issues/6248

It includes reordering of constructor declarations. Perhaps, this issue will resolve itself?

I'm closing this as "works for me". Will be happy to re-iterate a discussion with some more information on how to reproduce this.

costa100:
I have the exact same issue reported by the OP. I am using jooq 3.13.4. He described the accurately the issue. My database has exactly one table in the public schema and I generated the code using this config file:

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<configuration xmlns="http://www.jooq.org/xsd/jooq-codegen-3.7.0.xsd">
  <jdbc>
[...]

Issue #5489, opened 2016-08-10
UninitializedFieldError when using scala compiler option "-Xcheckinit"


Juraj-Masiar:
I'm using jOOQ in Play Framework 2.5.4 and after I upgraded and regenerated jOOQ from 2.7.4 to 2.8.4 I'm getting this runtime exception:

```
2016-08-10 12:44:32 - [error] - a.a.ActorSystemImpl - Uncaught fatal error from thread [application-contexts.database-6] shutting down ActorSystem [application]
java.lang.ExceptionInInitializerError: null
        at jooq_gen.Public.<init>(Public.scala:74)
        at jooq_gen.Public$.<init>(Public.scala:61)
        at jooq_gen.Public$.<clinit>(Public.scala)
        at jooq_gen.tables.Person.<init>(Person.scala:53)
        at jooq_gen.tables.Person.<init>(Person.scala:186)
Caused by: scala.UninitializedFieldError: Uninitialized field: Public.scala: 61
        at jooq_gen.Public$.PUBLIC(Public.scala:61)
        at jooq_gen.DefaultCatalog.<init>(DefaultCatalog.scala:42)
        at jooq_gen.DefaultCatalog$.<init>(DefaultCatalog.scala:24)
        at jooq_gen.DefaultCatalog$.<clinit>(DefaultCatalog.scala)
        at jooq_gen.Public.<init>(Public.scala:74)
```

After some googling I find out it's connected with scala compile option "-Xcheckinit" that I was using. Without that everything works fine but still sounds like a bug.

lukaseder:
Thank you very much for reporting.

Are you sure this is a jOOQ bug? I would suspect that this means there is something wrong in the bytecode that's produced by the Scala compiler... What Scala version are you using?

icl7126:
I'm using current 2.11.8 version.  
And these are the compiler options from my `build.sbt`:
```
scalacOptions in ThisBuild ++= Seq(
  "-target:jvm-1.8",
  "-encoding", "UTF-8",
  "-deprecation", // warning and location for usages of deprecated APIs
  "-feature", // warning and location for usages of features that should be imported explicitly
  "-unchecked", // additional warnings where generated code depends on assumptions
  "-Xlint", // recommended additional warnings
//  "-Xcheckinit", // runtime error when a val is not initialized due to trait hierarchies (instead of NPE somewhere else)
  "-Ywarn-adapted-args", // Warn if an argument list is modified to match the receiver
[...]

PR #8141, opened 2018-12-21
[#8139] preferred lazy instantiation for vals in object instances

astrofed:
this commit should fix [#5489] too

### DIFF
--- a/jOOQ-codegen/src/main/java/org/jooq/codegen/JavaGenerator.java
+++ b/jOOQ-codegen/src/main/java/org/jooq/codegen/JavaGenerator.java
@@ -749,7 +749,7 @@ public class JavaGenerator extends AbstractGenerator {
 
                     printDeprecationIfUnknownType(out, columnTypeFull);
                     if (scala)
-                        out.tab(1).println("val IDENTITY_%s = Identities%s.IDENTITY_%s",
+                        out.tab(1).println("lazy val IDENTITY_%s = Identities%s.IDENTITY_%s",
                                 identityId, block, identityId);
                     else
                         out.tab(1).println("public static final %s<%s, %s> IDENTITY_%s = Identities%s.IDENTITY_%s;",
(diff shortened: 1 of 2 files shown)


lukaseder:
Thank you very much for your suggestion. I think this is a reasonable improvement in general, but I would definitely like to first understand the root problem and be sure how to reproduce it...

astrofed:
I created a simple projecthttps://github.com/astrofed/jooq-codegen/commits/master

There are several commits, 2 latest are about this issue and potential fix. 

1. https://github.com/astrofed/jooq-codegen/commit/fa4add1cf218d346638895efbffde8e6b40c3f04
Here I just committed generated classes

2. https://github.com/astrofed/jooq-codegen/commit/1005329beaf87c8af44d9f655c4d62de2ab7bcc9
Here, please take a look line 37. I corrected manually this line replacing val with def and test completed successfully. Potentially I see several ways to fix it: 
    a. Replace vals with lazy vals like in my PR
    b. Replace this place in catalog generation method with def
   c. Remove this method from DefaultCatalog generally because we have a method getSchemas0  

The last decision, surely, is up to you.

The main root, the vals in the body of scala class is part of the constructor, so this val should be initialized during DefaultCatalog instantiation and it calls com.test.db.generated.Public.PUBLIC that's instantiating at the same time. This situation causes `Caused by: scala.UninitializedFieldError: Uninitialized field: .../com/test/db/generated/Public.scala: 28
	at com.test.db.generated.Public$.PUBLIC(Public.scala:28)
	at com.test.db.generated.DefaultCatalog.<init>(DefaultCatalog.scala:42)
	at com.test.db.generated.DefaultCatalog$.<init>(DefaultCatalog.scala:24)
	at com.test.db.generated.DefaultCatalog$.<clinit>(DefaultCatalog.scala)
	... 53 more
`

lukaseder:

lukaseder:
I was never able to reproduce this issue on my side. In the meantime, jOOQ 3.14 has fixed a lot of issues in the `ScalaGenerator` (https://github.com/jOOQ/jOOQ/issues/10191) - perhaps this issue here will go away

Issue #6604, opened 2017-09-22
Add multi Row[N] or Record[N] VALUES support


lukaseder:
The `INSERT .. VALUES` API is quite useful for multi-row inserts as well:

```java
ctx.insertInto(T)
   .columns(T.C1, T.C2, T.C3)
   .values(1, 2, 3)
   .values(4, 5, 6)
   .values(7, 8, 9)
   .execute();
```

But that works well only for constant numbers of rows. When the rows are determined dynamically, users have to assign the `InsertValuesStep` type to a local variable and loop through their rows, copying values to the DSL API, which is ugly.

We should have similar API like `DSL.values()` for the `INSERT` statement as well. Hypothetical syntax:

```java
Integer[][] values = {{ 1, 2, 3 }, { 4, 5, 6 }, { 7, 8, 9 }};
ctx.insertInto(T)
   .columns(T.C1, T.C2, T.C3)
   .values(values)
   .execute();
```

The method name probably cannot be overloaded easily, because of the existing `values(Collection<?>)` overload.

Several variants of this new API will be made available, supporting arrays, collections, etc, perhaps taking inspiration from the `Loader` API.

dmoebius:
How about a new values() method which accepts a stream, e.g. for the various InsertValuesStepN classes:
```
values(Stream<RowN> rows)
```
Could be used like that:
```java
Collection<A> mycoll = ...;
ctx.insertInto(T)
    .columns(T.C1, T.C2, T.C3)
    .values(mycoll.stream().map(a -> DSL.row(a.c1, a.c2, a.c3)))
    .execute()
```
Advantage: type safe, user is forced to take care about proper value mapping.

dmoebius:
Or, if a stream is not acceptable, use a differently named method `rowValues(Collection<RowN>)` instead, so that it doesn't collide with `values(Collection<?>)`.

lukaseder:
Regarding the `Stream` part, that will be done as part of https://github.com/jOOQ/jOOQ/issues/5133. It was delayed among other reasons because I wanted to be sure that it is acceptable for an API to accept a `Stream` (most APIs only return `Stream`). Background info here: https://twitter.com/lukaseder/status/897914723299143681

The right choice of `<T>` type will be a bit tricky. All of these are good and reasonable candidates:

- `Row[N]`
- `Record[N]`
- `Map`
- `Object[]`

The design of this API needs careful thought...

Diagoras:
[...]

PR #8148, opened 2018-12-30
[#6604] Improve INSERT .. VALUES API for dynamic multi-row inserts

Diagoras:
https://github.com/jOOQ/jOOQ/issues/6604

### DIFF
--- a/jOOQ/src/main/java/org/jooq/InsertSetMoreStep.java
+++ b/jOOQ/src/main/java/org/jooq/InsertSetMoreStep.java
@@ -37,6 +37,7 @@
  */
 package org.jooq;
 
+import java.util.Collection;
 import java.util.Map;
 
 /**
@@ -110,6 +111,9 @@ public interface InsertSetMoreStep<R extends Record> extends InsertOnDuplicateSt
     @Support
     InsertSetMoreStep<R> set(Map<?, ?> map);
 
+    @Support
+    InsertSetMoreStep<R> setAllMaps(Collection<Map<?, ?>> maps);
+
     /**
      * Set values in the <code>INSERT</code> statement.
      * <p>
@@ -120,6 +124,9 @@ public interface InsertSetMoreStep<R extends Record> extends InsertOnDuplicateSt
      */
     @Support
     InsertSetMoreStep<R> set(Record record);
+
+    @Support
+    InsertSetMoreStep<R> setAllRecords(Collection<Record> records);
 
     /**
      * Add an additional record to the <code>INSERT</code> statement
--- a/jOOQ/src/main/java/org/jooq/InsertSetStep.java
+++ b/jOOQ/src/main/java/org/jooq/InsertSetStep.java
@@ -278,6 +278,9 @@ public interface InsertSetStep<R extends Record> {
     @Support
     InsertSetMoreStep<R> set(Map<?, ?> map);
 
+    @Support
+    InsertSetMoreStep<R> setAllMaps(Collection<Map<?, ?>> maps);
+
     /**
      * Set values in the <code>INSERT</code> statement.
      * <p>
@@ -290,6 +293,9 @@ public interface InsertSetStep<R extends Record> {
      */
     @Support
     InsertSetMoreStep<R> set(Record record);
+
+    @Support
+    InsertSetMoreStep<R> setAllRecords(Collection<Record> records);
 
     /**
      * Add values to the insert statement with implicit field names.
@@ -308,6 +314,9 @@ public interface InsertSetStep<R extends Record> {
      */
     @Support
     InsertValuesStepN<R> values(Collection<?> values);
+
+    @Support
+    InsertValuesStepN<R> allValues(Collection<Collection<?>> values);
 
     /**
      * Add an empty record with default values.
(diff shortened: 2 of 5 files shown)


lukaseder:
Thank you for your suggestion. We're far from reaching consensus on what should be done here syntactically or semantically, nor whether anything should be done at all. I can thus not accept this PR.

Diagoras:
@lukaseder Got it. What would you need to see in order to accept PRs for this issue? A more definitive conversation and conclusion in the related issue? Something else?

Also, is there something I should be looking for to determine whether an issue is good to work on? Like a "contributions-wanted" label or something on issues that are considered open to contributions? Just want to avoid doing this again.

lukaseder:
> What would you need to see in order to accept PRs for this issue? A more definitive conversation and conclusion in the related issue? Something else?

There's a big problem on Github. It does not allow to activate the issue tracker without activating the PR subsystem. In fact, apart from typos and obvious bug fixes, I prefer not to get any PRs at all. jOOQ has become quite complex and adding a feature is a very costly endeavour (not as complex as adding something to the Java language, yet, but we're getting there).

The idea that PRs are a thing implies that users can re-prioritise efforts around an individual feature, which I don't think is such a high priority. Also, believe it or not, I don't believe in "community driven" open source. The efforts for the maintainer to review, improve, merge contributions are in now reasonable relation to the benefit of not "having to do the work". After all, the actual implementation is usually the least amount of work. Designing a feature is the big part. And as you're already suggesting, this will need a *"more definitive conversation"*, which I currently don't want to have (because of priorities).

> Also, is there something I should be looking for to determine whether an issue is good to work on? Like a "contributions-wanted" label or something on issues that are considered open to contributions? Just want to avoid doing this again.

Yes. Please don't :) I'm sorry to disappoint you here, but think of jOOQ like the Oracle Database. You usually wouldn't want to provide code, let alone features to it (the exception being "obvious" bug fixes). But you like using it, so the best thing to contribute is feedback, bug reports, etc.

A long time ago, there was a contributions-wanted label. I've removed it because it caused more trouble than benefit with tons "contributors" who needed tons of training to get going only to contribute 1-2 minor features. It simply doesn't scale.

Diagoras:
@lukaseder Ah, got it. That clears a lot of things up, thanks.

> By the way. This feels like RTFM'ing... Sorry about that. We have this document here: https://github.com/jOOQ/jOOQ/blob/master/CONTRIBUTING.md

Yeah, I took that paragraph to mean "don't open a PR without opening an issue and talking about it first", not "don't open a PR for issues beyond bug fixes".

How would you feel about modifying the "CONTRIBUTING.md" to make what you've stated in this PR explicit to other wannabe contributors? Specifically, that jOOQ isn't interested in people opening PRs for anything beyond minor and obvious bug fixes, and that people should especially never open PRs for new features.

That was non-obvious to me, because of how different it is from the usual GitHub culture of open contribution for triaged issues. I understand and respect your reasons for not adopting that model, but I think that given it's the norm you might be better off if you make your policy clear. At the very least, you'd get fewer randos like me opening PRs against issues you haven't prioritized.

BTW, I'd be willing to open an example PR for that change. I'm guessing that you'll probably want to handle it yourself, though. ;)

lukaseder:
> Yeah, I took that paragraph to mean "don't open a PR without opening an issue and talking about it first", not "don't open a PR for issues beyond bug fixes".

Well, we didn't "talk" about it yet, in the sense of having agreed to 1) what the API should look like, 2) whether this should be done at all :)

> How would you feel about modifying the "CONTRIBUTING.md" to make what you've stated in this PR explicit to other wannabe contributors? Specifically, that jOOQ isn't interested in people opening PRs for anything beyond minor and obvious bug fixes, and that people should especially never open PRs for new features.

I don't say never, but again, would you just go fork the Open JDK, implement tons of new Collection methods, provide a PR, and expect that PR to be merged?

Trivial PRs can be merged easily, because there's nothing to discuss. This here is quite big. You may not see it (yet), but adding new API like this needs to be added in a consistent way. This includes:

- Thinking about all the remaining API that deals with sets of records that needs to be considered
- Thinking about possible existing functionality, that should be merged with this one
- Thinking about the many places in the manual that need to be updated

This is just something that most aspiring contributors underestimate.

> That was non-obvious to me, because of how different it is from the usual GitHub culture of open contribution for triaged issues.

Well, jOOQ works well along the lines of this "culture". There are no such issues. You can browse them and contribute, except that you won't find any ;-)

> I understand and respect your reasons for not adopting that model, but I think that given it's the norm 

Who says this is the norm? There is bad UX on github's behalf, and there are hobbyist's projects, where quality doesn't matter. But I don't think you'll be able to add features that easily in more serious projects, without first discussing and *agreeing* with the maintainers about what should be done.

> you might be better off if you make your policy clear. At the very least, you'd get fewer randos like me opening PRs against issues you haven't prioritized.

I guess I will adapt the wording, indeed.

Diagoras:
Sounds good. Let me know if I should open an issue for changing the wording of the "CONTRIBUTING.md", but for now I'll assume that's overkill unless told otherwise. Thanks for the feedback, and for maintaining this awesome database tool.

lukaseder:
Hah, how meta :) I've updated the wording here: https://github.com/jOOQ/jOOQ/blob/master/CONTRIBUTING.md

I hope this makes it (would have made it) more clear. Thanks again for your understanding

Diagoras:
@lukaseder That's perfect! I think that's absolutely clear about what jOOQ is looking for in PRs. Thanks.

Issue #5733, opened 2016-12-27
Add a configuration option for varargs setters


lukaseder:
The new generated varargs setters (https://github.com/jOOQ/jOOQ/issues/4164) should have been introduced with a new configuration option to avoid warnings in generated code.

See also @ben-manes's feedback on the mailing list:
https://groups.google.com/d/msg/jooq-user/zlDXGePK7e8/85Hdjo9aEAAJ

etiennestuder:
@lukaseder A fix would be great since we are affected by that ;-)

lukaseder:
@etiennestuder I'm curious: how are you affected?

etiennestuder:
@lukaseder We make the file build on compiler warnings. So now the compile fails.

lukaseder:
OK, I see. I've just had a similar case before: https://github.com/jOOQ/jOOQ/issues/5775#issuecomment-270889860.

Hmm, this is third-party generated code - so perhaps you could be a bit more lenient with the warnings on that?

codeturner:
I'm seeing the same thing.  Unfortunately, it's polluting our previously clean list of java warnings in Eclipse, and violating our team policy.  And Eclipse doesn't allow you to disable this warning in generated code only, so we're ignoring them for now, but ... broken windows.

However, we're only seeing this for byte[] values.  I cannot think of a use case where a user of jooq would want to set the value of a byte[] by passing in a vararg list of individual bytes.

Can the generator simply not convert an array to varargs if the value type is a primitive?

lukaseder:
OK, message received. This feature was a mistake :)

lukaseder:
Fixed for jOOQ 3.10. This will be merged to version 3.9.2 (https://github.com/jOOQ/jOOQ/issues/5861) hopefully next week.

PR #8212, opened 2019-01-12
Add "overrides" to @SuppressWarnings class annotations. Fixes compile…

sLite:
This fixes a problem we encountered recently when trying to enable strict javac flags on one of my builds.

Database is postgres and some tables have array columns.

```
    /**
     * Set the first value.
     */
    Record1<T1> value1(T1 value);
```

gets overridden with

```
    /**
     * {@inheritDoc}
     */
    @Override
    public FooRecord value1(String... value) {
        setFooArray(value);
        return this;
    }
```

which leads to a ```overridden method has no '...' ``` 
compiler warning.

compiler options are
`def javacOptions = [ "-Xlint:all" , "-Xlint:-processing", "-Werror" ]`

### DIFF
--- a/jOOQ-codegen/src/main/java/org/jooq/codegen/JavaGenerator.java
+++ b/jOOQ-codegen/src/main/java/org/jooq/codegen/JavaGenerator.java
@@ -5822,7 +5822,7 @@ public class JavaGenerator extends AbstractGenerator {
         }
 
         if (!scala)
-            out.println("@%s({ \"all\", \"unchecked\", \"rawtypes\" })", out.ref("java.lang.SuppressWarnings"));
+            out.println("@%s({ \"all\", \"unchecked\", \"rawtypes\", \"overrides\" })", out.ref("java.lang.SuppressWarnings"));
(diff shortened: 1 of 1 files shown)


lukaseder:
Thanks for the suggestion. Indeed, this isn't nice - but I wonder if we should simply not generate the `@Override` annotation. Will review shortly.

lukaseder:
Notice, you can also turn off the generation of these varargs setters by specifying `<varargSetters>false</varargSetters>`

lukaseder:
Indeed, as discussed in #5733, the introduction of the generation of varargs setters has not been too well received in the community. Eventually, we're going to deprecate this feature. So I don't think it is worth adding additional compiler warnings (the one you suggested doesn't seem to work in Eclipse, for example). Instead, the recommendation is to simply turn off `<varargSetters/>`

sLite:
Very nice, thank you for the recommendation.
Wasn't aware that we can turn that off, that's the perfect solution for our usecase.

PR #8222, opened 2019-01-15
Resolves 8221: StringIndexOutOfBoundsException parsing PSQL Timestamp

berngp:


berngp:
Thanks @lukaseder for the quick reply.

PR #8322, opened 2019-02-13
Add text datatype

Xierip:


lukaseder:
Thank you very much for your suggestion. I need to reject this PR. We cannot make such an important change without first discussing your use case and the various options / workarounds you've already tried. Please review our contribution guidelines:

Thanks for your understanding

Issue #8398, opened 2019-03-09
support Datepart.Week in trunc() for Postgres


gunkelolaf:
### Use case:
We have some queries aggregating attributes on timeseries bases accumulated over weeks, therefore we use the Postgres datetrunc function.

### Possible solution you'd like to see:
change 
org.jooq.impl.TruncDate 

line 127

```
   case POSTGRES: {
                switch (part) {
                    case YEAR:   keyword = "year";   break;
                    case MONTH:  keyword = "month";  break;
                    case DAY:    keyword = "day";    break;
                    case HOUR:   keyword = "hour";   break;
                    case MINUTE: keyword = "minute"; break;
                    case SECOND: keyword = "second"; break;
                    default: throwUnsupported();
                }

                return DSL.field("{date_trunc}({0}, {1})", getDataType(), inline(keyword), date);
            }
```
to
```
case POSTGRES: {
                switch (part) {
                    case YEAR:   keyword = "year";   break;
                    case MONTH:  keyword = "month";  break;
                    case WEEK:   keyword = "week";   break;
                    case DAY:    keyword = "day";    break;
                    case HOUR:   keyword = "hour";   break;
                    case MINUTE: keyword = "minute"; break;
                    case SECOND: keyword = "second"; break;
                    default: throwUnsupported();
                }

                return DSL.field("{date_trunc}({0}, {1})", getDataType(), inline(keyword), date);
            }
```

### Versions:

- jOOQ: 3.11.10
- Java: 11
- Database (include vendor): Postgres

lukaseder:
Thank you very much for your request. This functionality is going to be implemented through #2132. We currently have non-standard date part support in `EXTRACT()`, and will also add support for such date parts in `TRUNC()`.

I'm closing this as a duplicate

Issue #2132, opened 2013-01-30
Add support for non-standard date-time fields in the EXTRACT() function


lukaseder:
Postgres supports a variety of other "date parts" than the ones specified in the SQL standard:
https://www.postgresql.org/docs/current/static/functions-datetime.html#FUNCTIONS-DATETIME-EXTRACT

When adding new parts, we'll have to:

- Annotate parts with `SQLDialect`
- Implement native support where available
- Emulate support where possible (in separate tickets)

The followint date parts (from PostgreSQL) would be useful:

- `CENTURY` (emulation based on `YEAR`)
- `DECADE` (emulation based on `YEAR`)
- `ISO_DAY_OF_WEEK` (see #7792)
- `DAY_OF_WEEK` (see #7792)
- `DAY_OF_YEAR` (see #7793)
- `EPOCH` (see #7794)
- `MICROSECOND`
- `MILLENNIUM`
- `MILLISECOND`
- `QUARTER` (see #7796)
- `TIMEZONE`
- `TIMEZONE_HOUR` (actually part of the SQL standard)
- `TIMEZONE_MINUTE` (actually part of the SQL standard)
- `WEEK`

Support will be needed in the following functions:

- [x] `extract()`

In a separate task, we'll add support for these: 

- [x] `dateAdd()` (see #8823)
- [x] `dateSub()` (see #8823)
- [x] `localDateAdd()` (see #8823)
- [x] `localDateSub()` (see #8823)
- [x] `localDateTimeAdd()` (see #8823)
- [x] `localDateTimeSub()` (see #8823)
- [x] `timestampAdd()` (see #8823)
- [x] `trunc()` (see #8418)

----
This was requested here:
- https://groups.google.com/d/topic/jooq-user/IVy74kfg988/discussion
- https://groups.google.com/d/msg/jooq-user/Xff7ajkwDUk/5KfnZl_0CAAJ
- http://stackoverflow.com/q/23335674/521799

lpireyn:
I want to implement this by adding values to the `DatePart` enum and ensuring that all its uses (`switch`, etc.) are still valid. Is there any reason against modifying the enum? Can values be inserted among the existing values (therefore modifying their ordinal value) without disrupting the existing code?

lukaseder:
[...]

PR #8399, opened 2019-03-09
[#8398] add datepart week to datetrunc for postgres

gunkelolaf:


lukaseder:
Thank you very much for your suggestion. In #2132, we'll be adding support for many other vendor-specific date parts in the context of the `TRUNC()` / `DATE_TRUNC()` function of various RDBMS dialects, so for the time being, I prefer not to include a solution that just fixes this for `WEEK` and PostgreSQL only.

As always, you can easily work around missing functionality in jOOQ by using plain SQL templating:
https://www.jooq.org/doc/latest/manual/sql-building/plain-sql-templating/

lukaseder:
This is a duplicate of #2132

Issue #6799, opened 2017-11-08
Improve performance of Convert logic


lukaseder:
Similar to the recent improvement of the `DefaultBinding` (#4930), we should avoid the long branching in `Convert` and find a more efficient way to apply auto conversion in jOOQ.

In a benchmark that only consists of mapping `Result` to various data types through the `DefaultRecordMapper`, the data type conversion does appear quite a bit in a profiling overview:

![image](https://user-images.githubusercontent.com/734593/32546361-1fa0a8b0-c47f-11e7-89ac-a68f61c74d41.png)

When drilling down into the different mapper implementations being tested:

![image](https://user-images.githubusercontent.com/734593/32546392-40002716-c47f-11e7-9409-d32e63f95fdd.png)

Then it can be seen that a lot of time is spent in the `Convert` class:

![image](https://user-images.githubusercontent.com/734593/32546448-70cf0b78-c47f-11e7-83b1-23bd5dedeabb.png)

----

A part of this waste might also be attributed to the excessive allocation of `ConvertAll` converters, which are used only for API reuse convenience:

![image](https://user-images.githubusercontent.com/734593/32546633-0d3ea86a-c480-11e7-831a-e0532d3d8753.png)

If escape analysis and/or stack allocation don't kick in here, that's some quite heavy burden.

lukaseder:
Some improvements will be implemented in https://github.com/jOOQ/jOOQ/issues/5865

lukaseder:
Some fixes have been implemented via #11099

lukaseder:
This will be implemented via a `DefaultConverterProvider` cache (see https://github.com/jOOQ/jOOQ/issues/6799)

PR #8432, opened 2019-03-18
Make ConvertAll.from static to avoid pointless allocations, partially addresses #6799

angusholder:
I've got this as 2 commits currently as this change reveals some apparently unnecessary invocations of the conversion routine, but I can't be sure that they're definitely unnecessary without running the test suite to check their removal wouldn't be a breaking change.

I could also now completely dissolve the ConvertAll class into the outer Convert class, but I've erred on the side of making fewer changes rather than more for now, so I'll leave it up to you

### DIFF
--- a/jOOQ/src/main/java/org/jooq/tools/Convert.java
+++ b/jOOQ/src/main/java/org/jooq/tools/Convert.java
@@ -298,7 +298,7 @@ public final class Convert {
     }
 
     public static final <U> U[] convertCollection(Collection from, Class<? extends U[]> to){
-        return new ConvertAll<U[]>(to).from(from);
+        return ConvertAll.from(from, to);
     }
 
     /**
@@ -317,8 +317,7 @@ public final class Convert {
      * Conversion type-safety
      */
     private static final <T, U> U convert0(Object from, Converter<T, ? extends U> converter) throws DataTypeException {
-        ConvertAll<T> all = new ConvertAll<T>(converter.fromType());
-        return converter.from(all.from(from));
+        return converter.from(ConvertAll.from(from, converter.fromType()));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


lukaseder:
Thank you very much for your suggestion and for signing the CLA. 

As a general remark (for next time): you should prefix the issue number with the # symbol, so Github will link the commits from the issue #6799 itself. I will comment on the commits directly.

jOOQ/src/main/java/org/jooq/tools/Convert.java
```diff
@@ -380,7 +380,7 @@ else if (from[0] != null && from[0].getClass() == toComponentType) {
      * @throws DataTypeException - When the conversion is not possible
      */
     public static final <T> T convert(Object from, Class<? extends T> toClass) throws DataTypeException {
-        return ConvertAll.from(ConvertAll.from(from, toClass), toClass);
+        return ConvertAll.from(from, toClass);
```
lukaseder: This change has been applied on top of your own change, so it is difficult to assess from the change itself whether it is correct.

I do remember there's a case where this was needed, but I can't track it down to an individual regression test right now

lukaseder:
After a quick review, I don't think that the regression risk / benefit are in favour of merging your suggestion. I will hence reject it. Thanks for your understanding.

angusholder:
I was meaning I could revert the second commit if you thought it risked regressions, the first commit alone makes no semantic changes to the code

lukaseder:
> the first commit alone makes no semantic changes to the code

Are you 100% *sure* :)

From my point of view, this is really a cost/benefit ratio assessment right here. Reviewing takes time. Making a very minor improvement without actually proving the benefit is not worth it, e.g. I will have to run the JMC and JMH benchmarks again, because your change doesn't include *proof* that the fix is actually better.

Without such proof, I right now prefer to simply drop the change as this would not have been a priority at all, otherwise.

PR #8645, opened 2019-05-17
Sakila/PostgreSQL: Replace nonexisting IF with CASE

mrehkopf:
Function `get_customer_balance` declared in postgres_sakila_schema.sql uses a function called IF that doesn't exist in PostgreSQL (probably a remnant from the original MySQL version).
Replace the IF with a corresponding CASE statement, including appropriate conversions for the intervals.


lukaseder:
Thanks a lot! Too bad PostgreSQL doesn't check procedure syntax when creating them (by default?). I hadn't noticed this yet...

PR #8839, opened 2019-06-22
Make project compile again

luvarqpp:
I have upgraded parent spring project to latest available and removed jooq property with version override to make project compile again.

I have examined target/generated sources and it seems that things are there as expected. Also mvn test finished with 6 tests run and no fail/ignored. Java version 12, zulu build.

### DIFF
--- a/jOOQ-examples/jOOQ-spring-boot-example/pom.xml
+++ b/jOOQ-examples/jOOQ-spring-boot-example/pom.xml
@@ -6,7 +6,7 @@
     <parent>
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-starter-parent</artifactId>
-        <version>2.0.2.RELEASE</version>
+        <version>2.1.6.RELEASE</version>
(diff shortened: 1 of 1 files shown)


lukaseder:
Thank you very much for your report.

The examples project compiles just fine for me. Be sure to build jOOQ 3.12.0-SNAPSHOT before building the examples, or, if you want to pull a published jOOQ version from Maven Central instead, use the examples from the matching version, e.g. 3.11.11: https://github.com/jOOQ/jOOQ/tree/version-3.11.11/jOOQ-examples

lukaseder:
Ah, I see now. It worked for me, because I had all sorts of pre-built jOOQ versions in my local Maven repository, of course. The fix, however, is not to exclude the version number, but to exclude the Spring Boot dependency, and use the one from github instead.

lukaseder:
I will fix this through https://github.com/jOOQ/jOOQ/issues/8857

Thanks again for your report!

Issue #8939, opened 2019-07-13
Support Flyway file ordering in DDLDatabase


yuecelm:
### Expected behavior and actual behavior:

Flyway SQL scripts are imported in the wrong order into DDLDatabase (with sort=semantic)

excepted:
src/main/resources/db/migration/V1__init_schema.sql
src/main/resources/db/migration/V1.1__add_indexes.sql

actual:
src/main/resources/db/migration/V1.1__add_indexes.sql <- triggers sql error due to missing table
src/main/resources/db/migration/V1__init_schema.sql

### Steps to reproduce the problem (if possible, create an MCVE: https://github.com/jOOQ/jOOQ-mcve):

src/main/resources/db/migration/V1__init_schema.sql:
```
CREATE TABLE company (
  id UUID PRIMARY KEY,
  name VARCHAR NOT NULL,
  created_at TIMESTAMP,
  updated_at TIMESTAMP
);
```
src/main/resources/db/migration/V1.1__add_indexes.sql:
```
CREATE INDEX company_index_name ON company (name);
```
pom.xml:
```
                <database>
                  <name>org.jooq.meta.extensions.ddl.DDLDatabase</name>
                  <inputCatalog></inputCatalog>
                  <inputSchema>PUBLIC</inputSchema>
                  <outputSchemaToDefault>true</outputSchemaToDefault>
                  <outputCatalogToDefault>true</outputCatalogToDefault>
                  <properties>
                    <property>
                      <key>sort</key>
                      <value>semantic</value>
                    </property>
                    <property>
                      <key>scripts</key>
                      <value>src/main/resources/db/migration/*</value>
                    </property>
                  </properties>
                </database>
```

### Versions:

- jOOQ: 3.11.11
- Java: OpenJDK 1.8.0_212
- Database (include vendor): org.jooq.meta.extensions.ddl.DDLDatabase
- OS: Ubuntu 18.04.2 LTS

yuecelm:
I would suggest to introduce sort=flyway for org.jooq.meta.extensions.ddl.DDLDatabase if the actual order is intended with sort=semantic

yuecelm:
I am also willing to implement `int compare(File o1, File o2)` for the new sort

yuecelm:
https://flywaydb.org/documentation/migrations#naming

lukaseder:
Thank you very much for your suggestion.

[...]

PR #8976, opened 2019-07-23
[jOOQ/jOOQ#8939] Support Flyway file ordering in DDLDatabase

yuecelm:


jOOQ-meta/src/main/java/org/jooq/meta/tools/FlywayFileComparator.java
```diff
@@ -0,0 +1,57 @@
...
+public final class FlywayFileComparator implements Comparator<File> {
+
+    public static final FlywayFileComparator INSTANCE = new FlywayFileComparator();
+
+    @Override
+    public final int compare(File o1, File o2) {
+        String s1 = o1 == null ? null : o1.getName();
+        String s2 = o2 == null ? null : o2.getName();
+
+        FlywayVersion v1 = FlywayVersion.fromVersion(s1 == null ? null : s1.substring(1, s1.indexOf("__")));
```
yuecelm: introduce SEPARATOR for "__"

yuecelm:
also forgot to add my name to ABOUT.txt, will do it after review...

lukaseder:
Thank you very much for your contribution. Will review later today.

jOOQ-examples/jOOQ-flyway-ddl-example/pom.xml
```diff
@@ -0,0 +1,111 @@
...
+        <dependency>
+            <groupId>org.jooq</groupId>
+            <artifactId>jooq</artifactId>
+            <version>${org.jooq.version}</version>
+        </dependency>
+
+    </dependencies>
+
+    <build>
+        <plugins>
```
lukaseder: Great idea to create an example project for this. It would be cool to actually use the flyway plugin, too and run a quick query on the resulting database. I'll make something up. I'd like to play around with Flyway's programmatic API for this, to create an in-memory H2 database.

lukaseder:
Thanks again for this PR. I'll merge and make some adaptations. This will be a very useful addition.

jOOQ-meta/src/main/java/org/jooq/meta/tools/FlywayFileComparator.java
lukaseder: This substring call can lead to:

```
Caused by: java.lang.StringIndexOutOfBoundsException: begin 1, end -1, length 12
    at java.lang.String.checkBoundsBeginEnd (String.java:3319)
    at java.lang.String.substring (String.java:1874)
    at org.jooq.meta.tools.FlywayFileComparator.compare (FlywayFileComparator.java:52)
    at org.jooq.meta.tools.FlywayFileComparator.compare (FlywayFileComparator.java:43)
```

e.g. when comparing non-flyway files or directories

lukaseder: This seems to work better:

```java
public final class FlywayFileComparator implements Comparator<File> {

    public static final FlywayFileComparator INSTANCE = new FlywayFileComparator();

    @Override
    public final int compare(File o1, File o2) {
        String s1 = o1 == null ? null : o1.getName();
        String s2 = o2 == null ? null : o2.getName();

        if (s1 != null && s2 != null) {
            int i1 = s1.indexOf("__");
            int i2 = s2.indexOf("__");

            if (i1 != -1 && i2 != -1) {
                FlywayVersion v1 = FlywayVersion.fromVersion(s1.substring(1, i1));
                FlywayVersion v2 = FlywayVersion.fromVersion(s2.substring(1, i2));

                return v1.compareTo(v2);
            }
        }

        // In case we're not comparing two Flyway files, fall back to default comparator logic
        return FileComparator.INSTANCE.compare(o1, o2);
    }
}
```

yuecelm: its not better: (i1 >=1 && i2 >=1), otherwise a File starting with "__" will still trigger an exception

lukaseder: You're right. Are empty version numbers possible? Or should we check for `(i1 > 1 && i2 > 1)`?

lukaseder: Never mind. The `FlywayVersion.fromVersion()` method already handles this.

yuecelm: :+1:

lukaseder:
> also forgot to add my name to ABOUT.txt, will do it after the review...

Just saw this now. Already did it :)

PR #8978, opened 2019-07-23
Make generatePojoMultiConstructor in JavaGenerator protected

kaandok:
Tiny PR, contributor agreement signed.


lukaseder:
Thanks a lot for your fix. Clearly, this was a bug as the Javadoc of the method suggests that subclasses can override the method, which they couldn't

lukaseder:
Fixed in 3.12, backport scheduled for 3.11.12 (#8981)

PR #8994, opened 2019-07-29
Fix NullPointerException when JDBC driver returns null on getGeneratedKeys call

### LINKED ISSUE #8993: NullPointerException when JDBC driver returns null on getGeneratedKeys call using ZDAL5 driver on MySQL
### Expected behavior and actual behavior:

When underlying JDBC drive doesn't support `getGeneratedKeys` and returns `null` when calling `getGeneratedKeys()`, jOOQ raises `NullPointerException` in `AbstractDMLQuery. executeReturningGeneratedKeysFetchAdditionalRows `:

https://github.com/jOOQ/jOOQ/blob/4d6bbc902baf6e667c64eec69030a9a78b644b21/jOOQ/src/main/java/org/jooq/impl/AbstractDMLQuery.java#L1116-L1121

The `if rs != null` check does not work because the variable `rs` is a `DefaultResultSet` which is not `null` but its private member variable `delegate` could be `null`. So when calling `DefaultResultSet.next()` which in turn calls `this.delegate.next()` raises NPE.

I think it might be good to check JDBC's `DatabaseMetaData.supportsGetGeneratedKeys` before calling `getGeneratedKeys`.

### Versions:

- jOOQ: 3.11.11
- Java: java version "1.8.0_151"
- Database (include vendor): MySQL
- OS: macOS
- JDBC Driver (include name if inofficial driver): ZDAL5

messense:
Closes #8993

### DIFF
--- a/jOOQ/src/main/java/org/jooq/tools/jdbc/DefaultStatement.java
+++ b/jOOQ/src/main/java/org/jooq/tools/jdbc/DefaultStatement.java
@@ -266,7 +266,13 @@ public class DefaultStatement extends JDBC41Statement implements Statement {
 
     @Override
     public ResultSet getGeneratedKeys() throws SQLException {
-        return new DefaultResultSet(getDelegateStatement().getGeneratedKeys(), this);
+        ResultSet rs = getDelegateStatement().getGeneratedKeys();
+        // Some JDBC drivers seem to illegally return null
+        // from getGeneratedKeys() sometimes
+        if (rs != null) {
+            return new DefaultResultSet(rs, this);
+        }
+        return null;
     }
 
     @Override
--- a/jOOQ/src/main/resources/META-INF/ABOUT.txt
+++ b/jOOQ/src/main/resources/META-INF/ABOUT.txt
@@ -29,6 +29,7 @@ Authors and contributors of jOOQ or parts of jOOQ in alphabetical order:
 - Laurent Pireyn
 - Luc Marchaud
 - Lukas Eder
+- Lusheng Lyu
 - Matti Tahvonen
 - Michael Doberenz
 - Michael Simons


jOOQ/src/main/java/org/jooq/tools/jdbc/DefaultStatement.java
```diff
@@ -266,7 +266,13 @@ public boolean getMoreResults(int current) throws SQLException {
...
     @Override
     public ResultSet getGeneratedKeys() throws SQLException {
-        return new DefaultResultSet(getDelegateStatement().getGeneratedKeys(), this);
+        ResultSet rs = getDelegateStatement().getGeneratedKeys();
+        // Some JDBC drivers seem to illegally return null
+        // from getGeneratedKeys() sometimes
+        if (rs != null) {
+            return new DefaultResultSet(rs, this);
+        }
+        return null;
```
knutwannheden: I wouldn't want to return `null` here, because as we have seen this is a violation of the JDBC API. I would then rather return an empty `ResultSet` (possibly using jOOQ's own `MockResultSet`).

messense: Could you elaborate on how to use `MockResultSet` to create a empty `ResultSet`?

knutwannheden: Thinking more about this, I don't really like the idea of incorporating JDBC driver specific hacks into `DefaultStatement`, as it only implements the delegate pattern.

If the `Settings#withReturnIdentityOnUpdatableRecord()` workaround works for you for now, I think we should leave it at that and then see if we also need to address #8996.

lukaseder:
I agree with @knutwannheden, this change seems to work around some issues, but it is very confusing for the `DefaultStatement` to handle this particular JDBC spec violation by drivers. We can certainly handle this elsewhere in a much better way.

lukaseder:
After a bit more investigation:

Yes, the JDBC driver violates the spec. But many do. By wrapping a `null` `ResultSet` in a `DefaultResultSet` inside our proxy, we're simply obfuscating the actual problem, which we've already handled in `AbstractDMLQuery`.

Perhaps, this workaround isn't such a bad idea after all - however, we'd have to implement it consistently for all methods, and not just for `ResultSet`.

Issue #9019, opened 2019-08-08
Render DROP TEMPORARY TABLE in MySQL


knutwannheden:
As discussed in #6371 the `TEMPORARY` keyword in `DROP TEMPORARY TABLE` actually has some semantics, and it would thus make sense if `DSL#dropTemporaryTable()` and `DSLContext#dropTemporaryTable()` would render this keyword for supporting dialects (like MySQL).

Let's see what the MySQL manual (https://dev.mysql.com/doc/refman/8.0/en/drop-table.html) has to say on this:

> The TEMPORARY keyword has the following effects:
>
> * The statement drops only TEMPORARY tables.
> * The statement does not cause an implicit commit.
> * No access rights are checked. A TEMPORARY table is visible only with the session that created it, so no check is necessary.
> 
> Including the TEMPORARY keyword is a good way to prevent accidentally dropping non-TEMPORARY tables.

AnthonyKot:
https://github.com/jOOQ/jOOQ/pull/9042

javolek:
This one should definitely be implemented, because, as stated above, without keyword `TEMPORARY` MySQL is going to silently commit the transaction. This makes a huge difference. 
I've spent several hours now to figure out, why my transactions are not properly rolled back, before I noticed in the log that the keyword `TEMPORARY` was missing.

lukaseder:
MariaDB also supports the keyword. MemSQL doesn't

lukaseder:
Fixed for 3.13.0. This improvement is of significant use and little risk, so we'll backport it to 3.12.2 (#9245)

lukaseder:
Related improvement: #9244 (support for `DROP TEMPORARY TABLE IF EXISTS`)

PR #9042, opened 2019-08-15
[#9019] MySQL supports DROP TEMPORARY tables and recommends to use it

AnthonyKot:
It should work. But soon there's no tests in the repo, could you give me idea how to check it's actually adding TEMPORARY kewword for MySQL? Thank you.

### DIFF
--- a/jOOQ/src/main/java/org/jooq/DSLContext.java
+++ b/jOOQ/src/main/java/org/jooq/DSLContext.java
@@ -9660,7 +9660,7 @@ public interface DSLContext extends Scope , AutoCloseable  {
      *
      * @see DSL#dropTemporaryTable(String)
      */
-    @Support
+    @Support({ MYSQL })
     DropTableStep dropTemporaryTable(String table);
 
     /**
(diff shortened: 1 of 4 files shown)


lukaseder:
Thank you very much for your PR. May I ask you to sign our transfer of rights agreement that allows us to dual license your contribution? A link to the document can be found in our contribution guidelines:

jOOQ/src/main/java/org/jooq/DSLContext.java
```diff
@@ -9660,15 +9660,15 @@
...
+    @Support({ MYSQL })
     DropTableStep dropTemporaryTable(String table);
 
     /**
      * Create a new DSL <code>DROP TEMPORARY TABLE</code> statement.
      *
      * @see DSL#dropTemporaryTable(Name)
      */
-    @Support
+    @Support({ MYSQL })
```
lukaseder: I'm not sure about this change. This would mean that *only* `MYSQL` supports the syntax, which would be backwards incompatible, and does not correspond to your suggested implementation. I'd prefer to keep the annotation as it is for now and fix this separately, as we validate other dialects in the future.

AnthonyKot:
@lukaseder thank you for your review. And no problems if you need to close PR. I really enjoyed this project when generating SQL script Java, just want to return something back. If you have open Issues I can work in, just ping me ;)

Issue #9055, opened 2019-08-17
should NO table qualify if NO table alias


trydofor:
if there is not table alias, (1) is better than (2).
```
-- (1) 
SELECT * FROM TABLE WHERE ID=?
DELETE FROM `TST_DEMO` WHERE `ID` = ?
-- (2) 
SELECT TABLE.* FROM TABLE WHERE TABLE.ID=?
DELETE FROM `TST_DEMO` WHERE `TST_DEMO`.`ID` = ?
```

AND  bad syntax if alias to Insert and Delete
### Expected behavior and actual behavior:
```kotlin
val t = TstDemoTable.TST_DEMO.`as`("t")
//val t = TstDemoTable.TST_DEMO
dsl.insertInto(t)
        .columns(t.ID,t.CREATE_DT,t.COMMIT_ID)
        .values(1L, LocalDateTime.now(),1L)
        .execute()
dsl.delete(t)
      .where(t.ID.eq(id))
      .execute()
```
**Expected** without  alias (or withou `tablename.` qulify as plain sql)
```mysql
insert into `TST_DEMO` (`ID`, `CREATE_DT`, `COMMIT_ID`) values (?, ?, ?)
delete from `TST_DEMO`  where `ID` = ?
```
**actual**
```
2019-08-17 09:59:57.930 DEBUG 12416 --- [           main] org.jooq.tools.LoggerListener            : Executing query          : 
insert into `TST_DEMO` as `t` (`ID`, `CREATE_DT`, `COMMIT_ID`) values (?, ?, ?)
delete `t` from `TST_DEMO` as `t` where `t`.`ID` = ?

....
.... without alias
....
delete from `TST_DEMO` where `TST_DEMO`.`ID` <= ?
select `TST_DEMO`.`ID` from `TST_DEMO` where `TST_DEMO`.`ID` <= ? limit ?
```
### Versions:

- jOOQ: 3.11.11
- Java: 8
- Database (include vendor): mysql 5.7
- OS:ubuntu 18.04
- JDBC Driver (include name if inofficial driver):

trydofor:
``` java
// TableFieldImpl
    @Override
    public final void accept(Context<?> ctx) {
        ctx.data(DATA_OMIT_CLAUSE_EVENT_EMISSION, true);

        if (ctx.qualify()) {
            // [#9055] should NO table qualify if NO table alias
            boolean qualify = true;
            if (table instanceof TableImpl) {
[...]

PR #9057, opened 2019-08-18
#9055 should NO table qualify if NO table alias

trydofor:
#9055 should NO table qualify if NO table alias


lukaseder:

There are many many reasons why we would want to fully qualify columns if qualification is available. Your suggested fix would break more things than it would fix.

I will thus have to reject it. Will discuss on the main issue #9055

PR #9181, opened 2019-09-08
Support INSERT...RETURNING and DELETE...RETURNING with MariaDB

nathan818fr:
MariaDB support INSERT...RETURNING since [10.5.1](https://mariadb.com/kb/en/library/mariadb-1051-release-notes/) and DELETE...RETURNING since [10.0.5](https://mariadb.com/kb/en/library/mariadb-1005-release-notes/) !
Adding support for those features would be great.

I'm not sure how to deal with backwards compatibility. Should we add some config flags? or check if the server version support those features? or something else?

### DIFF
--- a/jOOQ/src/main/java/org/jooq/DeleteReturningStep.java
+++ b/jOOQ/src/main/java/org/jooq/DeleteReturningStep.java
@@ -41,6 +41,7 @@ package org.jooq;
 // ...
 import static org.jooq.SQLDialect.FIREBIRD;
 // ...
+import static org.jooq.SQLDialect.MARIADB;
 import static org.jooq.SQLDialect.POSTGRES;
 // ...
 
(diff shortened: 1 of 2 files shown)


lukaseder:
Thank you very much for your suggestion. I have created a feature request for this: #9212. We'll definitely implement this for jOOQ 3.13.

I think it won't be as simple as your PR suggests. A lot of integration tests might be affected (they're not open source), and such an important change might suggest we need to introduce versioned MariaDB dialects, where the unversioned version (the family) would support `RETURNING`, whereas the old dialects would use the current emulations.

We'll take care of it. I will close this PR for the time being.

lukaseder:
This is now implemented for jOOQ 3.14: https://github.com/jOOQ/jOOQ/commit/f40a591bf44e99e29afcee3148dbafc090508b80

PR #9256, opened 2019-09-22
maven-plugin-api 3.6.2

sullis:


lukaseder:
Thank you very much for your suggestions. I appreciate the effort. 

Notice that we periodically upgrade *all* of our dependencies already. Allocating the time for managing individual patch release upgrades across all distributions (the commercial distributions are not affected yet by this PR) without any immediate benefits (e.g. actual bugs that we run into) is not worth the effort. This is why I am rejecting these PRs for now.

We will upgrade the dependencies eventually.

Thanks again for your efforts.

PR #9257, opened 2019-09-22
upgrade to Flyway 6.0.3

sullis:


lukaseder:
See comment here about the rationale behind this rejection: https://github.com/jOOQ/jOOQ/pull/9256#issuecomment-533990658

PR #9326, opened 2019-10-03
optimization: if no queries to execute in batch, nothing to do

evernat:
This case happened for a javamelody user: https://github.com/javamelody/javamelody/issues/863
Finding how it can happen lead me to suggest this PR.


lukaseder:
Thank you very much for your suggestion. We'll think about it.

lukaseder:

Reviewing the suggested change, I prefer not to go ahead with this because it might break existing code without any obvious gains.

Issue #8893, opened 2019-07-03
Add Settings.renderTable


lukaseder:
Some users might want to avoid table qualification of columns where this is not strictly necessary. For example:

```sql
-- This
SELECT a, b, c FROM t;

-- Instead of this
SELECT t.a, t.b, t.c FROM t;
```

This would work in addition to the omitted schema qualification:

```sql
SELECT s.t.a, s.t.b, s.t.c FROM s.t;
```

There are numerous ways such column names can be ambiguous. Hence, we offer the following flag values:

- `ALWAYS` (always qualify columns with a table. The default and current behaviour)
- `NEVER` (never qualify columns with a table. This risks breaking a lot of queries)
- `WHEN_MULTIPLE_TABLES` (qualify columns only when there are mutliple tables in the `FROM` clause)
- `WHEN_AMBIGUOUS_COLUMNS` (qualify columns only when we *know* there is an ambiguity. This requires a bit of calculation when generating the SQL string. It is currently unclear whether this is worth implementing for general purpose jOOQ, but it could definitely be useful for the parser).

A special use-case is the parser when it uses meta lookups (#7163), which currently produce qualified column references from what the parser sees as unqualified column references.

lukaseder:
A previous version of the description mentioned these options:

- `ALWAYS` (always qualify columns with a table. The default and current behaviour)
- `NEVER` (never qualify columns with a table. This risks breaking a lot of queries)
- `WHEN_NEEDED` (qualify columns only when we *know* there is an ambiguity. This requires a bit of calculation when generating the SQL string. It is currently unclear whether this is worth implementing for general purpose jOOQ, but it could definitely be useful for the parser).

We will only implement a boolean value that will never qualify columns with their table name, irrespective of possible ambiguities.

internetstaff:
It'd be great if `WHEN_NEEDED` was possible. Our particular issue is the Jooq-generated SQL is extremely large and somewhat unreadable for audit logging.

lukaseder:
> It'd be great if `WHEN_NEEDED` was possible. Our particular issue is the Jooq-generated SQL is extremely large and somewhat unreadable for audit logging.

Because of rendered tables? I mean, you can already remove the schema using `Settings.renderSchema`, then remove the quoting, and use aliases in your code to make it smaller...

internetstaff:
We've removed schemas and quoting, but we have a lot of long table names with lots of columns! :)

[...]

PR #9408, opened 2019-10-19
[#9055] should NO table qualify if NO table alias

trydofor:


trydofor:
PR for master
---------
I PR again, with `withRenderTable(false)` setting.
`renderTable` with default value `true` affects nothing.
when  renderTable is false, it only affects non-aliased table' field.

| renderTable | dsl.getSQL                            | affects |
|-------|---------------------------------------|---------|
| true(default)  | SELECT table.f1, table.f2 FROM table  | nothing   |
| true(default)  | SELECT t.f1, t.f2 FROM table **as** t     | nothing   |
| false | SELECT t.f1, t.f2 FROM table **as** t     | nothing   |
| false | SELECT f1, f2 FROM table              |  **NOT qulify table name**    |

this PR is most important in sharding scenario,
because the `table` will be replaced to `table_##`, 
the table qualified field may goes wrong.

thanks.

lukaseder:

I'm not sure we're ready to look into specific implementations yet. Your suggestion already shows that there are edge cases we need to take care of in different ways. For example, why would your suggestion not apply to aliases? Would it apply to plain SQL fields? Or to `field(Name)` fields? What about `MappedTable`?

There are many open questions, including whether this is really a good idea in the first place.

I will reject the issue for now. When we see more traction with the feature request itself, we might reconsider the functionality, but as of now, I prefer not to eagerly add this particular feature.

lukaseder:
Notice, the feature request this is for is https://github.com/jOOQ/jOOQ/issues/8893

PR #9640, opened 2019-12-07
Update dependency coordinates for Jakarta EE

napstr:
Ahoy, this PR updates the various javax dependency coordinates to their new homes under jakarta.

Didn't see the `activation-api` dependency anywhere so I'm assuming it's in the jooq enterprise code, it will require an adjustment too.

AFAIK no breaking stuff has been done with these APIs yet, and so far my projects have also been fine having a mix of the old and new ones, so here's to hoping that your secret tests will pass this PR :)

### DIFF
--- a/jOOQ-examples/jOOQ-jpa-example-entities/pom.xml
+++ b/jOOQ-examples/jOOQ-jpa-example-entities/pom.xml
@@ -64,18 +64,18 @@
                             <version>${org.hibernate.version}</version>
                         </annotationProcessorPath>
                         <annotationProcessorPath>
-                            <groupId>javax.xml.bind</groupId>
-                            <artifactId>jaxb-api</artifactId>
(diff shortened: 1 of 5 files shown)


lukaseder:
Thanks a lot for bringing this to our attention. I have created a feature request for this: https://github.com/jOOQ/jOOQ/issues/9641

We might look into this eventually, when we have the need to upgrade any of these dependencies. Currently, we don't, so I prefer not to rush this incompatible change yet.

napstr:
That's probably a good idea. It might be worth looking at how other projects handles this upgrade. Spring Boot for example additionally also excluded the old packages from transitive dependencies: https://github.com/spring-projects/spring-boot/commit/d6a869fa98eaeb3c180a8a3d94dcce8e3f51e204

lukaseder:

Might reopen if this gets more traction.

PR #9700, opened 2019-12-26
Sakila/PostgreSQL: Don't fail if language plpgsql already exists

rweisleder:
When using the Sakila DB with the Docker image postgres:12.1 the initialisation fails with:
`ERROR:  language "plpgsql" already exists`

With PostgreSQL 8.4 (see [1]) and earlier it was only possible to execute:
`CREATE [ PROCEDURAL ] LANGUAGE name`

Starting with PostgreSQL 9.0 (see [2]) it is possible to execute:
`CREATE [ OR REPLACE ] [ PROCEDURAL ] LANGUAGE name`

This commit makes the SQL script incompatible with PostgreSQL 8.4 and earlier.

[1] https://www.postgresql.org/docs/8.4/sql-createlanguage.html
[2] https://www.postgresql.org/docs/9.0/sql-createlanguage.html


knutwannheden:
@rweisleder Thank you very much for your contribution!

knutwannheden:
@rweisleder Thank you very much for your contribution!

Issue #9633, opened 2019-12-05
Add a new Settings.interpreterNameLookupCaseSensitivity 


lukaseder:
In the new DDL interpretation (#8528), some objects may refer some other objects by name, and we have to get case sensitivity right when looking up these names. For example:

```sql
create table t1 (
  id int,
  primary key (ID)
);
create table "t2" (
  "id" int,
  primary key ("PK")
);
```

In the above examples some identifiers are quoted, others are unquoted. Depending on the `Settings.interpreterDialect` (see #7337), and depending on this new flag, the statements are correct or incorrect.

- In some databases, identifiers are always case insensitive, and quoting only serves to escape keywords or to allow special characters in identifiers. Such dialects include SQL Server
- In other databases, identifiers are case insensitive only if unquoted. Quoted identifiers are case sensitive. Such dialects include Oracle
- In other databases, identifiers are always case sensitive, regardless of their being quoted. Again, quoting only serves to escape keywords or to allow special characters in identifiers. Such dialects include Sybase ASE
- Some dialects mix the above rules depending on the object type, operating system, configuration, and other random events. Looking at you, MySQL.

The new flag `Settings.interpreterNameLookupCaseSensitivity` supports these values:

- `DEFAULT`: the default value depending on the `Settings.interpreterDialect`, see list below
- `ALWAYS`: identifiers are always case sensitive
- `WHEN_QUOTED`: only quoted identifiers are case sensitive
- `NEVER`: identifiers are never case sensitive

The different values for `DEFAULT` are:

a | `interpreterDialect` | `interpreterNameLookupCaseSensitivity`
--- | --- | ---
✅ | `DEFAULT` | `WHEN_QUOTED`
✅ | `ACCESS` | `NEVER`
✅ | `ASE` | `ALWAYS`
✅ | `COCKROACHDB` | `WHEN_QUOTED`
✅ | `DB2` | `WHEN_QUOTED`
✅ | `DERBY` | `WHEN_QUOTED`
✅ | `FIREBIRD` | `WHEN_QUOTED`
✅ | `H2` | `WHEN_QUOTED`
✅ | `HANA` | `WHEN_QUOTED`
✅ | `HSQLDB` | `WHEN_QUOTED`
✅ | `INFORMIX` | `WHEN_QUOTED`
✅ | `INGRES` | `WHEN_QUOTED`
✅ | `MARIADB` | `NEVER` (might need additional object-type specific configuration)
[...]

Issue #9238, opened 2019-09-18
Add a new Settings.parseNameCase


lukaseder:
The `DDLDatabase` has a configuration option called `defaultNameCase`, which governs the name case for unquoted identifiers. This is relevant when e.g. a statement like

```sql
create table x (i int);
```

Is executed. Depending on the dialect, this would produce either a table `X` (e.g. Oracle) or a table `x` (e.g. PostgreSQL).

This feature needs to be moved into the core as a `Settings`, such that new functionality, like DDL simulation can decide what case it should translate unquoted identifiers to.

The different options are:

- `AS_IS` (`Abc` -> `Abc`)
- `UPPER` (`Abc` -> `ABC`)
- `UPPER_IF_UNQUOTED` (`Abc` -> `ABC`)
- `LOWER` (`Abc` -> `abc`)
- `LOWER_IF_UNQUOTED` (`Abc` -> `abc`)
- `DEFAULT` (this depends on the `parseDialect`)

The different values for `DEFAULT` are:

a | `parseDialect` | `parseNameCase`
--- | --- | ---
✅ | `DEFAULT` | `AS_IS`
✅ | `ACCESS` | `AS_IS` (?)
✅ | `ASE` | `AS_IS`
✅ | `COCKROACHDB` | `LOWER_IF_UNQUOTED`
✅ | `DB2` | `UPPER_IF_UNQUOTED`
✅ | `DERBY` | `UPPER_IF_UNQUOTED`
✅ | `FIREBIRD` | `UPPER_IF_UNQUOTED`
✅ | `H2` | `UPPER_IF_UNQUOTED`
✅ | `HANA` | `UPPER_IF_UNQUOTED` (?)
✅ | `HSQLDB` | `UPPER_IF_UNQUOTED`
✅ | `INFORMIX` | `LOWER_IF_UNQUOTED`
✅ | `INGRES` | `UPPER_IF_UNQUOTED` (?)
✅ | `MARIADB` | `AS_IS`
✅ | `MEMSQL` | `AS_IS`
✅ | `MYSQL` | `AS_IS` (table names in fact depend on the `lower_case_table_names` flag)
✅ | `ORACLE` | `UPPER_IF_UNQUOTED`
✅ | `POSTGRES` | `LOWER_IF_UNQUOTED`
✅ | `REDSHIFT` | `LOWER_IF_UNQUOTED`
✅ | `SQLDATAWAREHOUSE` | `AS_IS`
[...]

PR #9773, opened 2020-01-29
DDLDatabase: preserve casing of names

bennofs:
Without the `DATABASE_TO_UPPPER` option, H2 automatically translates
unquoted lowercase identifiers such as table names to upper case. This
is especially problematic in combination with liquibase, which generates
unquoted table names by default. Reverse-engineering a schema from a
liquibase changelog using the H2 DDL emulation will thus generate the
wrong casing for table names. If you use a case-sensitive database in
production (such as MySQL), then you get runtime errors.

### DIFF
--- a/jOOQ-meta-extensions/src/main/java/org/jooq/meta/extensions/AbstractInterpretingDatabase.java
+++ b/jOOQ-meta-extensions/src/main/java/org/jooq/meta/extensions/AbstractInterpretingDatabase.java
@@ -95,7 +95,7 @@ public abstract class AbstractInterpretingDatabase extends H2Database {
                 Properties info = new Properties();
                 info.put("user", "sa");
                 info.put("password", "");
-                connection = new org.h2.Driver().connect("jdbc:h2:mem:jooq-meta-extensions-" + UUID.randomUUID(), info);
+                connection = new org.h2.Driver().connect("jdbc:h2:mem:jooq-meta-extensions-" + UUID.randomUUID() + ";DATABASE_TO_UPPER=false", info);
 
                 export();
             }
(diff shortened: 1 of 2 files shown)


lukaseder:
Thank you very much for your report and for your suggested fix.

Unfortunately, this cannot be fixed as simply as you suggested. It may work for you and your specific setups, but not for other setups where upper case identifiers are fine.

We've reworked quite a bit how jOOQ deals with identifier cases, including the parser, and we've implemented an interpreter that should replace our simulation on H2 in the future. The interpreter will be able to reproduce the case sensitivity and default case behaviour of your target database product.

I will be happy to discuss specific feature requests or actual problems encountered, but I'd like to discourage suggestions of quick fixes like these, as such changes always need much more careful thought than it may first appear.

bennofs:
Thanks for the quick feedback. My impression was that the flag would only change things if you actually use lowercase identifiers, but I don't know much about H2 database, so I wasn't aware that this can have unintended side effects.

lukaseder:
Here's an overview (look for the checkboxes) of identifier case sensitivity among different dialects: https://github.com/jOOQ/jOOQ/issues/9633

Now, combine this with the following list that specifies the default name case of identifiers per dialect: https://github.com/jOOQ/jOOQ/issues/9238

Without these two flags (which will be supported starting from 3.13), it will be nearly impossible to fix this for everyone. And even with the flags, it will not be 100% correct, because some dialects, such as MySQL, have object type specific behaviour, combined with operating system specific behaviour. E.g. tables may be case sensitive when unquoted, whereas columns might not be.

So, if you want to be completely safe, I generally recommend using something like https://www.testcontainers.org to create your schema on an actual MySQL database and reverse engineer that in the code generator.

Issue #9872, opened 2020-02-25
LiquibaseDatabase should use configured changeLogParameters.contexts when updating


Fadelis:
### Use case:

Currently liquibase `contexts` parameter is not applied correctly as it is not a database property and needs to be passed to the `database.update(String)` method to work correctly.

### Possible solution you'd like to see:

Include explicit `contexts` property and pass it to the `update` method

### Versions:

- jOOQ: 3.13.1
- Java: 11
- Database (include vendor):
- OS:
- JDBC Driver (include name if inofficial driver):

lukaseder:
Thanks a lot for your suggestion. I agree that this `contexts` string would be a useful property to pass along.

I will comment on your commit

lukaseder:
Implemented in jOOQ 3.14.0. Backport scheduled for 3.13.2 (#9876).

Thanks again!

PR #9874, opened 2020-02-25
[#9872] LiquibaseDatabase should use provided contexts on update

Fadelis:


jOOQ-meta-extensions/src/main/java/org/jooq/meta/extensions/liquibase/LiquibaseDatabase.java
```diff
@@ -97,7 +97,8 @@
     @Override
     protected void export() throws Exception {
         String scripts = getProperties().getProperty("scripts");
-        includeLiquibaseTables = Boolean.valueOf(getProperties().getProperty("includeLiquibaseTables", "false"));
+        String liquibaseContexts = getProperties().getProperty("contexts", "");
```
Fadelis: I don't think this should be prefixed with the `database.` prefix as it is not a database property. But maybe it could be prefixed with something else, currently just went based on the `scripts` property.

lukaseder: Inside of Liquibase, the `Contexts` object gets passed along to `changeLogParameters`. I'd say, we call it `changeLogParameters.contexts`.

lukaseder:
Thanks a lot for your suggestion. I'll merge and fix according to my comments. I think we can safely backport this to 3.13.2, for immediate availability as well.

Issue #9873, opened 2020-02-25
LiquibaseDatabase should use actual changeLog database table names


Fadelis:
### Expected behavior and actual behavior:

Currently when excluding liquibase `changeLog` tables, static table name values are used, but those might be overridden by the configuration setters.

### Versions:

- jOOQ: 3.13.1
- Java: 11
- Database (include vendor):
- OS:
- JDBC Driver (include name if inofficial driver):

lukaseder:
Thanks a lot for your suggestion. Agreed, that currently doesn't work correctly for non-out-of-the-box configurations.

lukaseder:
Integrated in jOOQ 3.14.0. Backport scheduled for 3.13.2 (#9877).

Thanks again!

PR #9875, opened 2020-02-25
[#9873] LiquibaseDatabase should use actual changeLog table names

Fadelis:


lukaseder:
Thanks a lot for the suggested improvement. Agreed

PR #9934, opened 2020-03-11
Fix Sakila install scripts for Oracle Database

gvenzl:
This PR fixes the Sakila install scripts for Oracle Database and cleans up some of the others.

It also provides an installation ReadMe.


lukaseder:
Thanks a lot @gvenzl. I agree with your improvements.

Issue #10154, opened 2020-05-02
Improved default codegen naming strategy


benmccann:
### Use case:

My database has fields with camelCase names such as `accountId`. The records, by default, are generated like `account.getAccountid()` losing the casing that's in the DB.

### Possible solution you'd like to see:

The default naming strategy currently works well if some cases like all caps with underscores. If there are no underscores and mixed casing then I think the default could be improved and could preserve the user's casing. E.g. we could simply capitalize the first letter and prepend `get` or `set` in that case

### Possible workarounds:

I'm aware I can specify a code generation strategy that will fix this, but I thought this was a common and simple enough case that the default could be improved. The default obviously cannot work for everyone all the time and probably shouldn't be made too complicated, but I do think it could handle this common case better

### Versions:

- jOOQ: 3.13.1
- Java: 8
- Database (include vendor): PostgreSQL
- OS: Ubuntu
- JDBC Driver (include name if inofficial driver): PostgreSQL Java driver

lukaseder:
Thanks for your report. I think this is a duplicate, but I can't seem to find the other issue(s) right now.

I don't think using such casing is a very common use case outside of SQL Server, and even there a lot of names are in PascalCase (not camelCase), in case of which the `AS_IS` naming strategy can be used very easily. In PostgreSQL, once you create case sensitive identifiers, you have to quote them in all tools, which is something that people don't tend to appreciate when writing ad-hoc SQL. (SQL Server identifiers aren't case sensitive)

There should definitely be a solution that works better with camel cased identifiers and the generated getters and setters, but I just wanted to explain why this hasn't been a priority so far.

benmccann:
Ah. I wasn't that familiar with Postgres and didn't realize that it dropped the casing. Thank you very much for the tip. I will rename my columns to use underscores. I think you saved me a lot of headaches down the road

PR #10198, opened 2020-05-17
Improve default getter/setter casing

benmccann:
Closes https://github.com/jOOQ/jOOQ/issues/10154


jOOQ-codegen/src/main/java/org/jooq/codegen/DefaultGeneratorStrategy.java
```diff
@@ -18,36 +18,14 @@
...
- *
- *
- *
- *
- *
- *
- *
- *
- *
- *
```
lukaseder, line 36: Please don't change these things. They're here for good reasons: https://github.com/jOOQ/jOOQ/blob/master/CONTRIBUTING.md

PR #10252, opened 2020-06-08
Add empty comment after stmt in trigger.

lelle1234:
This makes it possible to run script from clp without changing stmt terminator. Verified with:

#> db2 +c -s -tf  db2-sakila-schema.sql

### DIFF
--- a/jOOQ-examples/Sakila/db2-sakila-db/db2-sakila-schema.sql
+++ b/jOOQ-examples/Sakila/db2-sakila-db/db2-sakila-schema.sql
@@ -26,7 +26,7 @@ BEFORE UPDATE ON actor
 REFERENCING NEW AS new
 FOR EACH ROW 
 BEGIN
-  SET new.last_update = current_date;
+  SET new.last_update = current_date; --
 END;
 
 CREATE TABLE country (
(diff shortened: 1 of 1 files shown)


lelle1234:
The third party tool is the Db2 clp
[image: image.png]


The thing is that ; has a special meaning in sqlpl (or plsql, I can never
remember which is which). The hack was introduced more than 10 years ago,
by the chief architect (Serge Rielau at the time). We discussed it in a
usenet post that I can not find because google groups seem unavailable.
Anyhow, it is not that important, you can leave it as is. The problem is
BTW mentioned in:

 https://github.com/flyway/flyway/issues/1163

Finally, the whole purpose of downloading Sakila was because of your
excellent blogpost:

https://blog.jooq.org/2017/09/28/10-cool-sql-optimisations-that-do-not-depend-on-the-cost-model/

I just wanted to check whether things have changed in Db2 11.5

Cheers
/Lennart


Den mån 8 juni 2020 17:25Lukas Eder <notifications@github.com> skrev:

lelle1234:
It turned out to be 2004 :-)

https://groups.google.com/forum/#!searchin/comp.databases.ibm-db2/serge$20guilty$20db2look%7Csort:date/comp.databases.ibm-db2/goEiaG3jsAI/DcImHGDCXisJ

lukaseder:
Alright, thanks for the links. I guess it doesn't hurt, even if it's ugly :)

Issue #7014, opened 2018-01-09
Add API to bind named parameters to plain SQL QueryParts


lukaseder:
The following test fails:

```java
@Test
public void testPlainSQLParamRenderingRepeatedParams() {
    ResultQuery<?> q1 = resultQuery(
        "select :first as a1, coalesce(:second, :second) as b1, :second as b2, :first as a2",
        param("first", 1),
        param("second", 2)
    );

    assertEquals("select 1 as a1, coalesce(2, 2) as b1, 2 as b2, 1 as a2", create.renderInlined(q1));
    assertEquals("select ? as a1, coalesce(?, ?) as b1, ? as b2, ? as a2", create.render(q1));
    assertEquals("select :first as a1, coalesce(:second, :second) as b1, :second as b2, :first as a2", create.renderNamedParams(q1));
    assertEquals("select :first as a1, coalesce(:second, :second) as b1, :second as b2, :first as a2", create.renderNamedOrInlinedParams(q1));
}
```

The reason is that `resultQuery(String, QueryPart...)` binds parameters by their indexes, not by their names.

While the change is very intuitive and the status quo is confusing, changing this behaviour is very backwards incompatible and would break a lot of assumptions and "accidentally working" code. We must be careful when implementing this. Perhaps this can be done in a backwards-compatible way, e.g. by specifying additional API.

MMeent:
I've tried my hand on this one to get a local application working. 

What I ended up doing was:

- Extend Settings with an extra config flag `bindByParamName`, default `false`, to allow opt-in usage of the feature
- Update `Tools.renderAndBind` to do the following (when `bindByParamName` is configured)
  - at the start, find `Param`s in the list of substitutes, and create a map from name to param.
  - in the parameter binding section, when a named bind variable declaration is found, use the name of the parameter to find a matching parameter and use that instead of the indexed method. If no such parameter is found for the name, then the variable declaration is skipped / rendered as-is. 

This worked fine for my use case (adding user-specified parameters to admin-customizable sql queries).

MMeent:
I've added a MR with the patch I created, see #10703

KirillKurdyukov:
Lukas, hi!
Is it true that substituting numbered parameters will not cause SQL injection?

lukaseder:
[...]

PR #10703, opened 2020-09-30
[#7014] Add API to bind named parameters to plain SQL QueryParts

MMeent:
This MR implements #7014:

It adds a setting to opt-in to the new behaviour, and implements named parameter binding

The added behaviour of binding parameters is:
1.) gather all named parameters into a `Map<param_name, List<NamedParam<name=param_name>>` to deal with possibly duplicated names being bound
2.) for each named parameter that is found that is the Nth of that name, it binds the Nth entry in the list of that parameter name, or if that is out of bounds, the last entry of that parameter name.

### DIFF
--- a/jOOQ/src/main/java/org/jooq/conf/Settings.java
+++ b/jOOQ/src/main/java/org/jooq/conf/Settings.java
@@ -105,6 +105,8 @@ public class Settings
     @XmlElement(defaultValue = "INDEXED")
     @XmlSchemaType(name = "string")
     protected ParamType paramType = ParamType.INDEXED;
+    @XmlElement(defaultValue = "false")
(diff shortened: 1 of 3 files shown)


lukaseder:
Thanks for the suggestion. We'll consider it once this is prioritised.

MuizMahdi:
Uninformative comment, but final issue for 3.15.0 🎉

lukaseder:
Sorry @MuizMahdi, the 3.15 ship has sailed 😁

lukaseder:
I'm closing this PR. While https://github.com/jOOQ/jOOQ/issues/7014 will be implemented eventually, the suggested implementation is probably not going to be the one that will make it.

Issue #10979, opened 2020-11-18
XMLDatabase does not produce parameters for overloaded routines


elenaem26:
### Expected behavior
Parameters are generated for overloaded postgreSql procedures pg_catalog. pg_advisory_xact_lock(bigint) and pg_catalog. pg_advisory_xact_lock(integer, integer) when being generated using information_schema.xml
### Actual behavior
Parameters are not generated for overloaded postgres procedures pg_catalog. pg_advisory_xact_lock(bigint) and pg_catalog. pg_advisory_xact_lock(integer, integer) when being generated using information_schema.xml.

### Steps to reproduce the problem
**Code generation configuration:**
```groovy
jooq {
    generateSchemaSourceOnCompilation = false
    xml(sourceSets.main) {
        jdbc {
            driver = 'org.postgresql.Driver'
            url = "URL"
            user = "USER"
            password = "PASSWORD"
        }
        generator {
            name = 'org.jooq.codegen.XMLGenerator'
            database {
                name = 'org.jooq.meta.postgres.PostgresDatabase'
                inputSchema = 'pg_catalog'
                includes = 'pg_catalog.pg_advisory_xact_lock'
            }
            target {
                directory = 'src/main/resources'
            }
        }
    }

    java(sourceSets.main) {
        generator {
            database {
                name = 'org.jooq.meta.xml.XMLDatabase'
                inputSchema = 'pg_catalog'
                includes = 'pg_catalog.pg_advisory_xact_lock'
                properties {
                    property {
                        key = 'dialect'
                        value = 'POSTGRES'
                    }
                    property {
                        key = 'xmlFile'
                        value = 'src/main/resources/information_schema.xml'
                    }
                }
            }
            generate {
                relations = true
                deprecated = false
                records = true
                pojos = true
                pojosEqualsAndHashCode = true
                fluentSetters = true
                javaTimeTypes = true
            }
            target {
                packageName = 'example.jooq'
                directory = 'src/main/java'
            }
        }
    }
}
```

**Steps to reproduce:**
1) Generate xml schema for routines pg_catalog.pg_advisory_xact_lock'. There will be different parameters in the schema that could be distinguished by specific_name tag.
Generated schema:
```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<information_schema>
    <schemata>
        <schema>
            <catalog_name></catalog_name>
            <schema_name>pg_catalog</schema_name>
            <comment></comment>
        </schema>
    </schemata>
[...]

PR #10980, opened 2020-11-18
[#10979] Compare parameters specific routine name with overloaded routine name

elenaem26:
Compare parameters specific routine name with overloaded routine name in XMLRoutineDefinition.java.
Otherwise there are no parameters will be generated for overloaded functions, like pg.pg_advisory_xact_lock in default schema pg_catalog in postgreSql

### DIFF
--- a/jOOQ-meta/src/main/java/org/jooq/meta/xml/XMLRoutineDefinition.java
+++ b/jOOQ-meta/src/main/java/org/jooq/meta/xml/XMLRoutineDefinition.java
@@ -40,6 +40,8 @@ package org.jooq.meta.xml;
 import static org.jooq.impl.DSL.name;
 import static org.jooq.tools.StringUtils.isBlank;
 
+import java.util.Arrays;
+
 import org.jooq.Name;
 import org.jooq.meta.AbstractRoutineDefinition;
 import org.jooq.meta.DataTypeDefinition;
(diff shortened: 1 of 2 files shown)


jOOQ-meta/src/main/java/org/jooq/meta/xml/XMLRoutineDefinition.java
```diff
@@ -167,4 +169,16 @@ protected void init0() {
...
     }
+
+    private Name getOverloadRoutineNamePart() {
+        Name qualifiedName = getQualifiedNamePart();
+        String lastName = qualifiedName.last();
+        if (lastName == null) {
+            return qualifiedName;
+        } else {
+            Name[] previousPath = Arrays.copyOf(qualifiedName.parts(), qualifiedName.parts().length - 1);
+            String overloadLastName = lastName + (isBlank(getOverload()) ? "" : "_" + getOverload());
```
lukaseder, line 180: I think it may well be a coincidence that our generated overload index is the same as the suffix of the specific name information. Running this query here in my own PostgreSQL installation:

```sql
select specific_name
from information_schema.routines
where routine_name = 'pg_advisory_xact_lock'
```

Yields:

```
specific_name             |
--------------------------|
pg_advisory_xact_lock_3093|
pg_advisory_xact_lock_3089|
```

Rather than using this suffix (which is only used for disambiguation by jOOQ), we should use the "specific name" information, explicitly.

I'll devise a fix.

lukaseder: In any case, the equality check you've fixed is definitely where the bug happens

lukaseder: Ah, I see why you've used this logic. It's the logic taken from `XMLGenerator`. But that's just one way to implement the "specific name", we shouldn't rely on it.

Issue #11118, opened 2020-12-13
Nested records produced by DefaultRecordMapper do not pass through RecordListener lifecycle


vkuzel:
### Current behaviour

Configuration (if there is any) of a parent record should be attached to a nested record created in the `DefaultRecordMapper`

* Location of the problem No 1: https://github.com/jOOQ/jOOQ/blob/824e689f19b712802cbf5e44928696ce574795d5/jOOQ/src/main/java/org/jooq/impl/DefaultRecordMapper.java#L825
* Location of the problem No 2: https://github.com/jOOQ/jOOQ/blob/824e689f19b712802cbf5e44928696ce574795d5/jOOQ/src/main/java/org/jooq/impl/DefaultRecordMapper.java#L1010

### Actual behaviour

Configuration is not attached, so for example configured custom converter provider won't be used in mapping of nested records.

### Steps to reproduce the problem

Following test should convert `CustomType`, but it doesn't.
```java
public static class CustomType {
    public String msg;
}

public static class NestedEntity {
    public CustomType customMessage;
}

public static class Entity {
    private NestedEntity nestedEntity;
}

public static class CustomConverterProvider implements ConverterProvider {
    @Override
    public @Nullable <T, U> Converter<T, U> provide(Class<T> tType, Class<U> uType) {
        if (tType == String.class && uType == CustomType.class) {
            return Converter.of(
                    tType, uType,
                    t -> {
                        CustomType customType = new CustomType();
                        customType.msg = (String) t;
                        return (U) customType;
                    },
                    u -> (T) ((CustomType) u).msg
            );
        } else {
            return null;
        }
    }
}

@Test
public void nestedMappingTest() {
    Entity entity = dslContext
            .configuration().set(new CustomConverterProvider())
            .dsl()
            .select(DSL.val("Hello World!").as("nestedEntity.customMessage"))
            .fetchOneInto(Entity.class);

    Assertions.assertEquals("Hello World!", entity.nestedEntity.customMessage.msg);
}
```

### Versions

- jOOQ: > 3.14.0 introduced by #6598

**Thank you for looking into this issue.**

lukaseder:
Thanks a lot for your thorough bug report. I can reproduce the problem. Agreed, the nested records should be attached, if `Settings.isAttachRecords() == true`.

I'll check your PR.

lukaseder:
As mentioned on the PR:

[...]

PR #11119, opened 2020-12-13
[jOOQ/jOOQ#11118] Attach configuration to a temporary record created in mapping of nested POJOs

vkuzel:
Fixes [jOOQ/jOOQ#11118]

Attaches a configuration (if set) from an existing record into short lived nested records used for mapping nested POJOs. It does use the static [`DefaultRecordMapper.attach()`](https://github.com/jOOQ/jOOQ/blob/824e689f19b712802cbf5e44928696ce574795d5/jOOQ/src/main/java/org/jooq/impl/DefaultRecordMapper.java#L1101-L1109) method to attach configuration, which checks whether configuration should be attached, even though this may not be necessary.

**Thank you, for reviewing this PR.**


lukaseder:
Thanks a lot for your suggestion.

This seems to fix the issue reported in #11118, but I think a more thorough solution wouldn't just take care of attaching the nested records, but to run through the entire `RecordListener` lifecycle, which in turn takes care of attaching the record, see `RecordDelegate::operate` (the `RecordDelegate` class has been moved to top level between jOOQ 3.14 and 3.15).

PR #11130, opened 2020-12-15
Scala deprecation fixes

arashi01:
Bump Scala and scala-maven-plugin versions from 2.13.2 to 2.13.4 and 4.3.0 to 4.4.0 respectively.

Fixes to allow compilation with Scala 3.0. Tested to compile with Scala 2.10 through 3.0.0-M2

Added Scala compiler arguments to build to display deprecation and/or Scala feature usage warnings.

### DIFF
--- a/jOOQ-scala_2.13/pom.xml
+++ b/jOOQ-scala_2.13/pom.xml
@@ -14,7 +14,7 @@
     <name>jOOQ Scala 2.13</name>
 
     <properties>
-        <scala.version>2.13.2</scala.version>
+        <scala.version>2.13.4</scala.version>
         <jooq-testdata-dir>${project.basedir}/../.data</jooq-testdata-dir>
     </properties>
 
@@ -35,6 +35,12 @@
                         </goals>
                     </execution>
                 </executions>
+                <configuration>
+                    <args>
+                        <arg>-deprecation</arg>
+                        <arg>-feature</arg>
+                    </args>
+                </configuration>
             </plugin>
 
             <!-- The jar plugin -->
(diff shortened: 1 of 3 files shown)


lukaseder:
Thanks a lot for your suggestions. I'll try to bump those versions on my side and see if all test still pass with your suggestions.

lukaseder:
In the meantime, the most up to date `scala-library` version is 2.13.5. I'll try that, instead.

lukaseder:
I'll do this in the context of a new issue: https://github.com/jOOQ/jOOQ/issues/11525

jOOQ-scala_2.13/pom.xml
```diff
@@ -35,6 +35,12 @@
                         </goals>
                     </execution>
                 </executions>
+                <configuration>
+                    <args>
+                        <arg>-deprecation</arg>
+                        <arg>-feature</arg>
+                    </args>
+                </configuration>
```
lukaseder, line 43: I'll put these in the parent pom.

jOOQ-scala_2.13/src/main/scala/org/jooq/scalaextensions/Conversions.scala
```diff
@@ -822,4 +821,30 @@ object Conversions {
...
+      object Inner {
+        import scala._
+        import jdk.CollectionConverters
+        val Converters = CollectionConverters
+      }
+    }
+
+    val Converters = Scope2.Inner.Converters
+  }
+
```
lukaseder, line 849: That looks laborious. Without an explanation, I don't think that's really necessary.

lukaseder:
Closing this as a duplicate of https://github.com/jOOQ/jOOQ/issues/11525.

Thanks again for your suggestions

PR #11346, opened 2021-02-01
Update FunWithKotlinAndJOOQ.kt

bk322:
I did a small study on reading config file stored inside JAR.

Java:

```
src/
├── demo/
│   └── App.java
└── config.properties
```

```java
package demo;

import java.util.Properties;
import java.io.InputStream;
import java.io.IOException;

public class App {
  public static void main(String[] args) {
    Properties properties = new Properties();

    try {
      InputStream config = App.class.getResourceAsStream("/config.properties");

      properties.load( config );

      System.out.println( properties.getProperty("foo") );
      System.out.println( properties.getProperty("baz") );

    } catch (IOException e) {
      e.printStackTrace();
    }
  }
}
```
`config.properties`:
```
foo=bar
baz=1
```
Compilation:
```
javac src/demo/App.java -d target
cp src/config.properties target/
cd target
  jar --create --file=../demo.jar --main-class=demo.App .
cd -
java -jar ./demo.jar
```
Same in Kotlin:
```
.
src/
├── demo/
│   └── App.kt
└── config.properties
```
`App.kt`:
```kotlin
package demo

import java.util.Properties

class App {
  companion object {
    @JvmStatic fun main(args: Array<String>) {
      var properties: Properties = Properties()

      properties.load( App::class.java.getResourceAsStream("/config.properties") )

      println( properties.getProperty("foo") )
      println( properties.getProperty("baz") )
    }
  }
}
```
Compilation:
```
cd src
  kotlinc demo/App.kt -include-runtime -d ../demo.jar
  jar --update --file=../demo.jar --main-class=demo.App config.properties
cd ..
java -jar demo.jar
```

### DIFF
--- a/jOOQ-examples/jOOQ-kotlin-example/src/main/kotlin/org/jooq/example/kotlin/FunWithKotlinAndJOOQ.kt
+++ b/jOOQ-examples/jOOQ-kotlin-example/src/main/kotlin/org/jooq/example/kotlin/FunWithKotlinAndJOOQ.kt
@@ -45,12 +45,14 @@ import org.jooq.impl.DSL.*
 
 import org.jooq.example.kotlin.db.h2.tables.references.*
 
-fun main(args: Array<String>) {
+class FunWithKotlinAndJOOQ {
+  companion object {
+    @JvmStatic fun main(args: Array<String>) {
 
     // This example project simply uses a standalone JDBC connection, but you're free to use any other
     // means to connect to your database, including standard DataSources
     val properties = Properties();
-    properties.load(properties::class.java.getResourceAsStream("/config.properties"));
+    properties.load(FunWithKotlinAndJOOQ::class.java.getResourceAsStream("/config.properties"));
 
     DSL.using(
         properties.getProperty("db.url"),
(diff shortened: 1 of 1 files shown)


lukaseder:
I'm sorry, I don't know what I'm looking at, or what problems this is addressing...

bk322:
that kotlin example, as it stands now, does not work. Because the code
which accesses config.properties in the Jar is wrong.

lukaseder:
Oh, I see, thanks for the hint. Wow, how could that have produced a regression...

This seems to work, too:

```kotlin
properties.load(object {}.javaClass.getResourceAsStream("/config.properties"));
```

bk322:
Indeed, `object {}.javaClass` works with Maven build.

Invoking `kotlinc` compiler directly on
```kotlin
package demo

import java.util.Properties

fun main(args: Array<String>) {
  var properties: Properties = Properties()

  properties.load( object {}.javaClass.getResourceAsStream("/config.properties") )

  println( properties.getProperty("foo") )
  println( properties.getProperty("baz") )
}
```

results in

> Error: Could not find or load main class demo.App

lukaseder:
But there's no main class, I think? I don't have a `kotlinc` compiler outside of IntelliJ ready, but should main not be invoked on the package `demo`, directly?

Issue #9335, opened 2019-10-08
Add a jooq-kotlin-coroutines module to support transactional coroutines based on reactive streams Publishers


lukaseder:
It should be possible to add some Kotlin extensions directly in the jOOQ core library and ship them to users integrating with Kotlin, without affecting Java users. For example, we could support coroutines as `DSLContext` extensions or `ResultQuery` extensions

A proof of concept has been made here:
https://github.com/marshallpierce/jOOQ/commit/72b0b3b0127afc758094016e124b0f7d1da98cd4

As suggested on the user group:
https://groups.google.com/d/msg/jooq-user/NS5anAYmIgI/FlV3H6gvAwAJ

marshallpierce:
I think this shouldn't even be too nasty from a dependency perspective -- simply mark the kotlin stdlib and coroutine dependencies as `optional` in maven parlance, and users who want to use them will already have those things on their classpath anyway.

lukaseder:
We have a new module starting from jOOQ 3.14: `jooq-kotlin` (https://github.com/jOOQ/jOOQ/issues/6256). Let's see what can be done in there.

lukaseder:
Looking into this now. It seems that the main purpose of the suggested approach is to replace our call to `blocking(() -> { ... })`, which invokes `ForkJoinPool::managedBlock` by Kotlin's `suspend` syntax. The content that is inside of the `transactionResult` method here: https://github.com/marshallpierce/jOOQ/commit/72b0b3b0127afc758094016e124b0f7d1da98cd4#diff-588eae5d26e55b16e058f61c3a1ce658 is the same as that of `DefaultDSLContext`.

At least, that seems to be it for these `DSLContext.transaction()` methods. But wouldn't the expectation here be for any database interaction to be "suspending", even without a transaction? Surely, leveraging this language feature for jOOQ's transaction API is better than nothing, but I'd say that a lot of people upvoting this issue would expect more.

lukaseder:
I don't want to rush anything here (as with reactive support, too: #6298). There seems to be a significant investment, making all sorts of API `suspend`-enabled for Kotlin users, and I don't fully understand what this means yet.

marshallpierce:
My initial desire was to be able to call `suspend` functions inside a closure passed to `transactionResult` (and ideally propagate coroutine context from the caller as well). 

Use case:
- in a transaction, query some data
- call some other thing (which `suspend`s) with that data
- update data with the response from the suspending call
- commit

[...]

Issue #6256, opened 2017-05-20
Add Kotlin infix and extension functions to a new jOOQ-kotlin extensions module


lukaseder:
# Ideas

### Infix functions

There is some potential of improving the Kotlin / jOOQ experience by providing out-of-the-box `infix` functions in Kotlin, which allow for creating expressions like these:

```kotlin
where (
    (FIRST_NAME eq "John") and (LAST_NAME eq "Doe")
)
```

The fluency of writing jOOQ queries might improve. However, unfortunately, all these infix functions have the same operator precedence and are left associative. This could prove to be very confusing when mixing `and` and `or`, or even more complex expressions.

We should probably not rush to add this functionality, and observe these discussions instead:

- https://youtrack.jetbrains.com/issue/KT-1440
- https://discuss.kotlinlang.org/t/specify-operator-precedence-for-infix-functions/2685/5

### Extension Functions

Another low hanging fruit are extension functions, which may simplify the jOOQ / Kotlin experience, for example:

```kotlin
@Support
inline fun <reified T: Any> Select<Record1<T>>.`as`(alias: String): Field<T> {
    return field(this).`as`(alias);
}
```

Which allows for calling:

```kotlin
// this
select(T.X).from(t).`as`("x")

// rather than this
field(select(T.X).from(T)).`as`("x")
```

# Roadmap

The following improvements are under review (gathered from comments here and elsewhere) and will be implemented via this issue:

### Implemented:

- [x] Add auxiliary functions to wrap `Select<Record1<T>>` in `field(Select)` for some common methods, like `eq`, `ne`, `asc`, `desc` etc.
- [x] Add extension methods to `Field<Boolean>`, from `Condition`
- [x] Array access `array_column[index]`: https://github.com/jOOQ/jOOQ/issues/229

### Rejected:

- [x] ~Moving `componentN()` methods (added with #6245) into this extension module. That's a backwards incompatible change, but acceptable to reduce the API surface of Java applications~: The out of the box experience that works without adding the `jOOQ-kotlin` module is much more important.
- [x] ~Add auxiliary functions that help prevent using `KotlinClass::class.java`, reducing them to `KotlinClass::class`~: Won't  be implemented, see: https://github.com/jOOQ/jOOQ/issues/6256#issuecomment-625175287
- [x] ~Some extensions that allow for a more native `set { a = x; b = y }` usage. Details: https://github.com/jOOQ/jOOQ/issues/6256#issuecomment-628010674~
[...]

PR #11483, opened 2021-02-19
Add various small Kotlin extensions

JohannesZick:
Hi,

first off, I've read the contributing guide. And no, this is not a bugfix. But the extensions are so small that I felt this is easier than explaining everything in text on [insert medium of choice here]. So, feel free to close this, but maybe consider adding some of these into the next release. I'm sure a lot of people keep writing these extensions all the time in their projects.

If you have questions or suggestions, please just ask.

### DIFF
--- a/jOOQ-kotlin/src/main/kotlin/org/jooq/kotlin/Extensions.kt
+++ b/jOOQ-kotlin/src/main/kotlin/org/jooq/kotlin/Extensions.kt
@@ -2,6 +2,10 @@ package org.jooq.kotlin
 
 import org.jooq.*
 import org.jooq.impl.DSL.*
+import java.util.concurrent.Executor
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+import kotlin.coroutines.suspendCoroutine
 
 // ----------------------------------------------------------------------------
 // Extensions to make Field<Boolean> a Condition
@@ -82,6 +86,15 @@ inline fun Field<Boolean>.not(): Condition = condition(this).not()
 // ----------------------------------------------------------------------------
 // Extensions to extract fields from Tables
 // ----------------------------------------------------------------------------
+
+@Support
+inline fun <reified T: Any> TableLike<*>.field(index: Int) = this.field(index, T::class.java)
+
+@Support
+inline fun <reified T: Any> TableLike<*>.field(name: Name) = this.field(name, T::class.java)
+
+@Support
+inline fun <reified T: Any> TableLike<*>.field(name: String) = this.field(name, T::class.java)
 
 @Support
 inline operator fun TableLike<*>.get(index: Int) = this.field(index)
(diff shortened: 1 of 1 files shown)


lukaseder:
Thanks a lot for your suggestions.

I've never thought about PRs this way, but I agree that a PR to convey some ideas is a very good approach, as long as there is no expectation for it to be merged. I've updated the contribution guidelines. Thanks for that idea.

I will comment on the individual suggestions directly in the commit.

jOOQ-kotlin/src/main/kotlin/org/jooq/kotlin/Extensions.kt
```diff
@@ -83,6 +87,15 @@ inline fun Field<Boolean>.not(): Condition = condition(this).not()
...
 
+@Support
+inline fun <reified T: Any> TableLike<*>.field(index: Int) = this.field(index, T::class.java)
+
+@Support
+inline fun <reified T: Any> TableLike<*>.field(name: Name) = this.field(name, T::class.java)
+
+@Support
+inline fun <reified T: Any> TableLike<*>.field(name: String) = this.field(name, T::class.java)
+
```
lukaseder, line 98: These types of suggestions have been made before by Jake Wharton
https://github.com/jOOQ/jOOQ/issues/6256#issuecomment-512314479

And later rejected after thorough analysis:
https://github.com/jOOQ/jOOQ/issues/6256#issuecomment-625175287

TL;DR: While looking very enticing at first, it's too late to add such convenience to the vast jOOQ API, as:

- It would be a *huge* change with many many methods involved
- Some methods could not have this kotlin-idiom applied consistently, leading to more cognitive friction than benefit

jOOQ-kotlin/src/main/kotlin/org/jooq/kotlin/Extensions.kt
```diff
@@ -220,3 +233,59 @@ inline fun <reified T: Any> Select<Record1<T>>.notIn(vararg values: T): Conditio
...
+inline fun <reified T: Any> Fetchable<*>.fetchOneInto() = this.fetchOneInto(T::class.java)
+
+@Support
+inline fun <reified T: Any> Fetchable<*>.fetchOptionalInto() = this.fetchOptionalInto(T::class.java)
+
+@Support
+inline fun <reified T: Any> Fetchable<*>.fetchSingleInto() = this.fetchSingleInto(T::class.java)
+
+@Support
+inline fun <reified T: Any> Fetchable<*>.fetchStreamInto() = this.fetchStreamInto(T::class.java)
```
lukaseder, line 263: Same as above.

jOOQ-kotlin/src/main/kotlin/org/jooq/kotlin/Extensions.kt
```diff
@@ -220,3 +233,59 @@ inline fun <reified T: Any> Select<Record1<T>>.notIn(vararg values: T): Conditio
...
+suspend fun <R: Record> Query.executeAwait(executor: Executor): Int = suspendCoroutine {
+        this.executeAsync(executor)
+                .handle { result, error -> if (error != null) it.resumeWithException(error) else it.resume(result) }
+}
+
+@Support
+suspend fun Query.executeAwait(): Int = suspendCoroutine {
+        this.executeAsync()
+                .handle { result, error -> if (error != null) it.resumeWithException(error) else it.resume(result) }
+}
```
lukaseder, line 291: These kinds of things are being discussed here: https://github.com/jOOQ/jOOQ/issues/9335

I don't feel at ease with the status quo of having:

- Classic blocking API (which will "soon" be supercharged by Loom)
- Asynchronous API via JDK `CompletionStage`
- Reactive API via reactive-streams or JDK `Flow`
- Kotlin specific coroutine API

All of jOOQ's API needs to be duplicated many times, with a lot cognitive overhead for users and maintenance overhead on our side.

I'm still betting on Loom, hoping to sit out all of the above fads, which are *much more* thoroughly addressed by Loom, without the need to change a single thing to jOOQ's API (or our internal JDBC usage).

lukaseder:
Since all the suggestions are already addressed in other issues, I'm closing this PR as a duplicate.

Thanks again for your suggestions. If you have more ideas, just let me know. I'll be very happy to discuss them!

jOOQ-kotlin/src/main/kotlin/org/jooq/kotlin/Extensions.kt
JohannesZick: Ah, thanks for the link. These are the only ones I was aware of, but I really don't have a good oversight of the full API. At some point it just becomes too much, I agree.

jOOQ-kotlin/src/main/kotlin/org/jooq/kotlin/Extensions.kt
JohannesZick: Thanks again for the link. I'll add my two cents there.

hantsy: Currently the Select is a `Publisher`, it is easy to convert to Kotlin Coroutines via Reactor kotlin extensions.

lukaseder: Can you show an example, @hantsy ?

Issue #11492, opened 2021-02-22
Add jbang example to the manual


lukaseder:
Apparently, running the jOOQ code generator with jbang is super easy:
https://twitter.com/maxandersen/status/1363803581397278721

We should document this in the manual/tutorial.

### Tasks

- [ ] Document this in the manual / tutorial
- [x] Add a "jbang-catalog" (see tweet) to further simplify this

maxandersen:
if you create a jbang-catalog repo I can submit PR with how I think this should/could work ;)

lukaseder:
That sounds great, though is a repo the right place for this? Usually, I prefer modules or example projects, not a top level repo. Is there a reason why a repo is better, except to simplify the PR?

maxandersen:
its just that if you put it in lets say `jooq/jooq` then it would be `jbang codegen@jooq/jooq` instead of just `jbang codegen@jooq`.

I can definitely start with a pr on `jooq/jooq` if you want that.

maxandersen:
Made a PR against jooq/jooq to show what I'm talking about. Details in #11492

I do recommend using `jooq/jbang-catalog`, but up to you.

maxandersen:
btw. I used the word `codegen` as that seem to be the word used in docs; but maybe `gentool` is more apropriate.

And we could also add most common jdbc drivers as direct `//DEPS` but I thought nice to have explicit control using `--deps`.

maxandersen:
I meant https://github.com/jOOQ/jOOQ/pull/11499 - copy/paste error

lukaseder:
> its just that if you put it in lets say `jooq/jooq` then it would be `jbang codegen@jooq/jooq` instead of just `jbang codegen@jooq`.

Oh, so these coordinates are github based? I'll have to do some reading first, but given that `jOOQ/jOOX` and other libraries exist, I guess it wouldn't be correct to assume a `jOOQ/jbang-catalog` github project is jOOQ's only, as opposed to any other's.

On the other hand, currently, there's no conflict in our namespace with respect to `codegen`, so this might still be changed later on if some other library ever does introduce code generation...

> btw. I used the word `codegen` as that seem to be the word used in docs; but maybe `gentool` is more apropriate.

The module is called `jooq-codegen`, so I think it's a good idea to reuse that term.

maxandersen:
You can also always introduce a repo specific jbang catalog or just prefix them to avoid conflicts in future.

lukaseder:
[...]

PR #11499, opened 2021-02-22
[jOOQ/jOOQ#11492] Add jbang catalog

maxandersen:
This PR adds jbang catalog so you can run jooq codegen using jbang.dev

i.e. given a `library.xml` with settings for postgres the simplest way
to run using postgres driver is as follows:

`jbang --deps org.postgresql:postgresql:42.2.14 codegen@maxandersen/jooq/jbang library.xml`

Notice the `codegen@maxandersen/jooq/jbang` - that is just where this PR lives right not:
`codegen` script at `maxandersen` github user's `jooq` git repo in the `jbang` branch.

If you push this commit it gets shorten to:

`jbang --deps org.postgresql:postgresql:42.2.14 codegen@jooq/jooq library.xml`

And if we make a `jooq/jbang-catalog` repo it can be as short as:

`jbang --deps org.postgresql:postgresql:42.2.14 codegen@jooq library.xml`

In case users want to use a specific version of the codegenerator they can do that
passing in `-Djooq.version=1.2.3` as follows:

`jbang --fresh -Djooq.version=1.2.3 codegen@jooq library.xml`

Note: the `--fresh` is to just ensure it will do a fresh resolve/build in case you used a previous version that is cached.

### DIFF
--- a/codegen.java
+++ b/codegen.java
@@ -1,0 +1,11 @@
+///usr/bin/env jbang "$0" "$@" ; exit $?
+//DEPS org.jooq:jooq-codegen:${jooq.version:3.14.7}
+
+import static java.lang.System.*;
+
+public class codegen {
+
+    public static void main(String... args) throws Exception {
+        org.jooq.codegen.GenerationTool.main(args);
+      }
+}
--- a/jbang-catalog.json
+++ b/jbang-catalog.json
@@ -1,0 +1,8 @@
+{
+  "catalogs": {},
+  "aliases": {
+    "codegen": {
+      "script-ref": "codegen.java"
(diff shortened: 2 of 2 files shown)


lukaseder:
Finally got some time to try this out. Fantastic, it works! Great for tutorials! 😀

Will merge, and adapt a few things, then integrate into the jOOQ 3.15 manual, see https://github.com/jOOQ/jOOQ/issues/11492. Will probably follow your repo suggestion, rather than putting this in the main repository.

maxandersen:
Glad you like it and thanks for publishing!

Issue #11304, opened 2021-01-26
ExecuteContext.exception() should accept null exceptions and translate them to an unspecified DataAccessException


bwajtr:
### Expected behavior
To be honest, I'm not sure what should be the correct behaviour, but JooqExceptionTranslator should not fail with NPE. The org.springframework.jdbc.core,.JdbcTemplate implementation solves this problem by throwing org.springframework.jdbc.UncategorizedSQLException when the exception translation returns null. So I think that the JooqExceptionTranslator should behave the same...

### Actual behavior
- The DB error is caught by org.springframework.boot.autoconfigure.jooq.JooqExceptionTranslator, which in turn calls SQLErrorCodeSQLExceptionTranslator to translate the error. However, in 5.3.0  version of spring-jdbc the contract of SQLErrorCodeSQLExceptionTranslator#translate() changed from @NotNull to @Nullable (see here https://github.com/spring-projects/spring-framework/commit/e9cded560d6ecb73aa5bfca57e06b09fa0013294#diff-c6aa1c6a4b11e8a722808e945c4c5b6ef471c42871c7ce830554dde81ad7f2c2L68). In the case of the custom errors (or any non-standard error), the SQLErrorCodeSQLExceptionTranslator now returns null (instead of UncategorizedSQLException) and JooqExceptionTranslator fails with NullPointerException, because it does not expect null.
 
```
java.lang.NullPointerException
	at org.jooq.impl.DefaultExecuteContext.exception(DefaultExecuteContext.java:737)
	at org.springframework.boot.autoconfigure.jooq.JooqExceptionTranslator.handle(JooqExceptionTranslator.java:83)
	at org.springframework.boot.autoconfigure.jooq.JooqExceptionTranslator.exception(JooqExceptionTranslator.java:55)
	at org.jooq.impl.ExecuteListeners.exception(ExecuteListeners.java:274)
	at org.jooq.impl.AbstractQuery.execute(AbstractQuery.java:390)
	at org.jooq.impl.AbstractResultQuery.fetch(AbstractResultQuery.java:333)
	at org.jooq.impl.AbstractResultQuery.fetchInto(AbstractResultQuery.java:1545)
	at org.jooq.impl.SelectImpl.fetchInto(SelectImpl.java:3936)
```

### Steps to reproduce the problem

- Have any PostgreSQL database
- Use jOOQ in Spring Environment (reproduced with latest Spring Boot 2.4.2, where spring-jdbc was in version 5.3.3)
[...]

PR #11574, opened 2021-03-05
ExecuteContext.exception() translates to DataAccessException with cause in case of null

jirkait:
Enhances [jOOQ/jOOQ#11304]


lukaseder:

We cannot merge a PR that modifies code in subtle ways without an explanation of why this modification would be useful, and ideally a test case or some other kind of sample code illustrating how jOOQ behaves before / after the change.

Please see these resources for more details:

- https://github.com/jOOQ/jOOQ/blob/main/CONTRIBUTING.md
- https://github.com/jOOQ/jOOQ-mcve

PR #11612, opened 2021-03-11
Make information schema protected for XMLDatabase

devkanro:
This PR just makes the info method of XMLDatabase can be overridden.

In my case, I need to read XML from jar resource in the classpath, I try to extend it, but I found I can't override the info method.

And I try to refactor it, make it more extendable, but I found a bug in using XSL case.
The `reader` always is `null` in line 202, and it will not load the XML file.

### DIFF
--- a/jOOQ-meta/src/main/java/org/jooq/meta/xml/XMLDatabase.java
+++ b/jOOQ-meta/src/main/java/org/jooq/meta/xml/XMLDatabase.java
@@ -133,7 +133,7 @@ public class XMLDatabase extends AbstractDatabase {
 
     InformationSchema               info;
 
-    private InformationSchema info() {
+    protected InformationSchema info() {
         if (info == null) {
 
             // [#8118] Regardless of failure, prevent NPEs from subsequent calls


lukaseder:

We tend to not make things non-final or non-(package)-private just like that. See:

- https://blog.jooq.org/2014/09/15/this-is-the-final-discussion/
- https://blog.jooq.org/2017/03/20/the-open-closed-principle-is-often-not-what-you-think-it-is/

These things don't scale for API designers. We try to make everything as backwards compatible as possible, and you're trying to access internal API, which would expose not only the internal representation of the `XMLDatabase`, but also the "protocol" (e.g. at what point in time the objects are initialised, etc.). There are so many things that can go wrong if we publish this internal API in the future, there would be no end to discussing all the "bugs" and "regressions" on it, let alone all the other things that people may wish to have opened.

Instead (given that this seems to be a case of https://xyproblem.info), let's discuss both your *actual* use-case (because there might be a useful feature request hidden in your request), as well as the bug (which can probably be fixed).

Both of these measures will lead to *much* better solutions for everyone, than a quick and dirty opening of jOOQ's internals.

PR #11613, opened 2021-03-11
Fix read XML from classpath and transforming by XSL

devkanro:
This PR fix 2 bugs of XMLDatabase.

1. Reader not be set when the XSL file be used in XML transforming. (Mentioned in the #11612)
2. FilePattern can't load the classpath resource correctly when using the XML file inside a jar.

### DIFF
--- a/jOOQ-meta/src/main/java/org/jooq/meta/xml/XMLDatabase.java
+++ b/jOOQ-meta/src/main/java/org/jooq/meta/xml/XMLDatabase.java
@@ -187,6 +187,7 @@ public class XMLDatabase extends AbstractDatabase {
                                 }
                                 else {
                                     InputStream xslIs = null;
+                                    reader = source.reader();
 
                                     try {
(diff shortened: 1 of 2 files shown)


lukaseder:

As I mentioned in your previous PR (https://github.com/jOOQ/jOOQ/pull/11612#issuecomment-796580809), I'd really love to discuss the bug first, before discussing a potential solution. So far, I don't know how the bug can be reproduced. Ideally, an MCVE would be created (template here: https://github.com/jOOQ/jOOQ-mcve), but any other set of instructions to reproduce the problem will be fine, as well.

Once we can reproduce the bug, we can investigate the options to solving it. Perhaps you found the right fix, perhaps not. It's not really possible to tell from your diff alone:

1. What the problem is
2. How the problem can be reproduced
3. If your fix is correct

I can't merge (or even review) such PRs without a clear problem description, very sorry.

PR #12441, opened 2021-09-17
Fix generating multi-key DAOs with Java records

octylFractal:
The single-key case was correctly mapped for this, but not the multi-key case.


lukaseder:
Thank you very much for your suggested fix. I have created an issue for this: https://github.com/jOOQ/jOOQ/issues/12442. I'll review shortly.

lukaseder:
Can confirm, this fixes the issue. Thank you very much!

Issue #5713, opened 2016-12-20
Add Configuration.dataTypes() as a runtime lookup registry for data types


lukaseder:
This idea has been around for a while (e.g. https://github.com/jOOQ/jOOQ/issues/650) as it is rather difficult to:

- predict jOOQ's current data type registry behaviour
- change this behaviour

A driving motivation includes also a flag that should already be available (but isn't) in order to implement #4429. Right now, the code generator needs to manually rewire all temporal data types to their JSR 310 equivalents. This should be done automatic and centralised.

josephglanville:
Is this also the tracking issue for being able to configure the registry?
Would this allow us to influence what `Field` definitions jOOQ generates when doing metadata queries? i.e register a DataType and have `DSLContext.meta()` return appropriately typed `Field`s?

This behaviour would be extremely useful for us as we are doing a lot of very dynamic data type things right now and are probably using APIs that aren't suitable/stable for our use-cases, i.e ` DefaultDataType.getDataType()`

Being able to register custom types and also retrieve the data type again from the registry for reconstruction from a string form would be ideal. i.e an API like this: `Configuration.dataTypes().getTypeFor(dialect: SQLDialect, typeName: String) : DataType`.

If there is a way to do this right now with the static registry that would be super useful as a work around even if it's going away soon.

For now what I am doing is not using the `DSLContext.meta()` API for discovering types and instead falling back to JDBC metadata API direct access as jOOQ unfortunately swallows the original type information when creating a `SQLDataType.OTHER`.

lukaseder:
Thanks for your message, @josephglanville. It's likely that your use-case will be covered in some way.

> If there is a way to do this right now with the static registry that would be super useful as a work around even if it's going away soon.

Not any documented way, but you can obviously try to reverse engineer how the static type registry works currently, and hack your way around it, including accessing it via reflection, or relying on `DefaultDataType` (which is `public`). At your own risk, of course.

josephglanville:
I was able to get most of the way there with `DefaultDataType` so I think we are happy for now.

Looking forward to a better supported mechanism though. We are happy to rely on internals for now because our application is tightly coupled to jOOQ anyway and we can afford to have to handle breaking changes but it's not an ideal state.

lukaseder:
There are numerous reasons why this should be implemented. The most important reason is clearly that a lot of folks are struggling to debug the static type registry warnings that started appearing with jOOQ 3.19, e.g.:

[...]

PR #12474, opened 2021-09-29
added postgres regclass type

cdalexndr:
Casting an `oid` to `regclass` results in human readable value.
Ex: `select relation::regclass from pg_locks`


lukaseder:
Thanks for your suggestion. A bit of history. Looking at the [1.6.6 version of the `PostgresDataType`](https://www.jooq.org/javadoc/1.6.6/org/jooq/util/postgres/PostgresDataType.html) (the oldest whose Javadoc is still available online), we can see that the `REGPROC` type has been added there, but others have not. In all these years, this has never been an issue, which hints much more at:

1. This not being a requirement for most users
2. These vendor specific types should be supported via custom data type bindings, a feature that wasn't available at the time of jOOQ 1.6.6

If anything needs to be changed in this area, I think it would be much more to remove such esoteric PG specific types from `PostgresDataType` (e.g. `MONEY`, `OID`, `OIDVECTOR`, `REGPROC`, `TID`, `XID`), which itself is also already deprecated.

So, future tasks involve:

- Removing these dialect specific data type APIs (from the public APIs)
- Replacing the static data type registry by a configurable one: https://github.com/jOOQ/jOOQ/issues/5713, where users can add global type support rather than ad-hoc type support or code generation type support.

As such, I'm rejecting this PR here as "won't fix"

Issue #12703, opened 2021-12-06
Add DefaultConverterProvider support for conversions between TIME[TAMP]TZ strings and Local[Date]Time


lukaseder:
We currently cannot directly convert from `String` to `Timestamp`/`Time` or `LocalDateTime`/`LocalTime`, if the string version of the timestamp has a time zone:

```java
ConverterProvider cp = new DefaultConverterProvider();

// This prints null
Converter<String, LocalDateTime> c = cp.provide(String.class, LocalDateTime.class);
System.out.println(c.from("2000-01-01 10:00:00+02"));
```

However, if we convert first to `OffsetDateTime`/`OffsetTime`, then it works:

```java
// This prints 2000-01-01T09:00, when running in CET
Converter<String, OffsetDateTime> c1 = cp.provide(String.class, OffsetDateTime.class);
Converter<OffsetDateTime, LocalDateTime> c2 = cp.provide(OffsetDateTime.class, LocalDateTime.class);
System.out.println(c1.andThen(c2).from("2000-01-01 10:00:00+02"));
```

We've been patching the internal temporal parsing logic time and again, it seems that these problems can only be fixed thoroughly by completely re-implementing parsing on our own, rather than patching things all the time.

This has already been done for performance reasons, because the `java.time` parser is very slow, even for the very popular ISO 8601 formats! See #7986, https://bugs.openjdk.java.net/browse/JDK-8213243. With just a little more effort, we can retrofit this parser for all temporal data types.

lukaseder:
This should fix https://github.com/jOOQ/jOOQ/issues/12699

katzyn:
I'm just wondering how this client-side conversion can take time zone of database session or transaction into account? Even if driver knows it internally, it isn't exposed via JDBC API.

katzyn:
`Timestamp` internally holds UTC-based value, similar to `java.time.Instant`, but with various bugs. Conversion from `TIMESTAMP WITH TIME ZONE` to `java.sql.Timestamp` doesn't depend on time zones at all. But conversion to `LocalDateTime` does.

lukaseder:
Yes that's right. The epoch timestamp remains unchanged, but `Timestamp::toString`, `Timestamp::toLocalDateTime` as well as both `Timestamp::valueOf` use the client JVM's default time zone, so in the context of this task here, is there really any issue?

katzyn:
[...]

Issue #12699, opened 2021-12-04
`DateTimeParseException` when mapping a row containing a timestamp with PostgreSQL 14.1


huuff:
### Expected behavior
The inner object is created correctly with the `LocalDateTime` object.
### Actual behavior
The code throws 
```java
java.time.format.DateTimeParseException: Text '2021-12-04 13:25:09.691023+01' could not be parsed at index 10
```
### Steps to reproduce the problem

The following minimal test produces the issue:

```java
    @Test
    void dateTimeFormatExceptionTest() throws Exception {
        try (Connection conn = DriverManager.getConnection("jdbc:tc:postgresql:14.1:///")) {
            var dsl = DSL.using(conn, SQLDialect.POSTGRES);
            var query = dsl.select(
                    row(currentLocalDateTime()).mapping(InnerTestObject::new)
            );

            var result = query.fetchOne(mapping(OuterTestObject::new));
        }
    }
```

Where `InnerTestObject` only contains a `LocalDateTime` and `OuterTestObject` only contains an `InnerTestObject`. I've provided an MCVE [here](https://github.com/huuff/jooq-mcve) (haven't followed exactly the template, since I'm using testcontainers but I hope it works)

### Versions

- jOOQ: 3.15.5
- Java: 17
- Database (include vendor): PostgreSQL 14.1
- OS: GNU/Linux 5.10.76
- JDBC Driver (include name if inofficial driver): postgresql 42.3.1

lukaseder:
Thanks a lot for your report. I can reproduce this, even without the mappings. This doesn't work:

```java
Record1<Record1<LocalDateTime>> q = ctx
    .select(row(currentLocalDateTime()))
    .fetchOne();
```

Will look into a fix right away. Note, this seems to happen only at the top level where we use native `ROW()` expressions in PostgreSQL, and then deserialiase the PostgreSQL string format. There are probably quite a few other such bugs related to serialisation edge cases that we haven't covered in tests yet, good reminder to look into this more thoroughly in the near future, e.g. when we implement native `MULTISET` implementations and `MULTISET` emulations using `ARRAY` (see https://github.com/jOOQ/jOOQ/issues/12021).

The reason why this doesn't work, currently, is because `currentLocalDateTime()` produces a `CURRENT_TIMESTAMP` expression, which is of type `TIMESTAMPTZ`, but we're expecting it to return a `TIMESTAMP` type, and it seems we currently cannot convert between the two formats. This also doesn't work:

```java
[...]

Issue #12547, opened 2021-10-22
Internal patchIso8601Timestamp() isn't safe for 5 digit years


lukaseder:
For simplicity and speed, the internal `patchIso8601Timestamp()` utility was implemented with a hard-coded 4-digit year in mind. Though, ISO 8601 supports 5-digit years as well. We need to take this into account. Current implementation:

```java
private static final String patchIso8601Timestamp(String string, boolean t) {
    if (string.length() > 11)
        if (t && string.charAt(10) == ' ')
            return string.substring(0, 10) + "T" + string.substring(11);
        else if (!t && string.charAt(10) == 'T')
            return string.substring(0, 10) + " " + string.substring(11);

    return string;
}
```

GeorgiMinkov:
Dear Lucas,

Do you know if this ticket was taken? If not I will like to give it a try as a first issue for the jOOQ community.

Additional question - the _"Convert"_ class was migrated to the org.jooq.impl from version 3.15.0, do we need to update the org.jooq.tools implementation of _"Convert"_.

What is the timeline for this update?

lukaseder:
Thanks for your offer @GeorgiMinkov. This is a very low priority issue. You can only fix the very obvious parts in this method, but you don't have access to the integration tests where the bulk of this work will take place (make sure this works correctly in all dialects).

While I appreciate the offer, I prefer you don't fix this right now, given the low priority and the effect this has on the roadmap's re-prioritisation, which I think isn't worth it right now (until a significant amount of users have this problem).

GeorgiMinkov:
Thank you for your answer @lukaseder. Several questions for you:

- Is there a board or a list in which I can track what task needs to be implemented?
- If there is no such list, what is the level of priority that you are aiming?
- Is there a group chat in which you discuss different problems and approaches? (like Discord, Google Chat)
- About the current one:
-- As this converter by design is "less type-safety", we can update it to be universal between the two ISO upgrades
-- Example:

`

        private static final String patchIso8601Timestamp(String string, boolean t) { 
            return t ? string.contains("T")
                        ? string
                        : string.replaceFirst(" ", "T")
                    : string.replaceFirst("T", " ");
`
Of course this logic can be bound with the specific indexes

If you are satisfied with the above approach I can create a corresponding branch and you can use it (merge it) whenever you need to.

In the mean time I will search for better tasks

lukaseder:
I think the expectations here are off.

[...]

PR #12705, opened 2021-12-07
Fix issue 12547 for Iso8601 time conversion

zaak71:


lukaseder:
Thanks for your suggestion. As I made clear (I think) in https://github.com/jOOQ/jOOQ/issues/12547#issuecomment-976228479 and https://github.com/jOOQ/jOOQ/blob/main/CONTRIBUTING.md, I don't want this particular issue to be fixed by the community

In fact, there are new plans of fixing this much more thoroughly as related discussions have shown in:

- https://github.com/jOOQ/jOOQ/issues/12699
- https://github.com/jOOQ/jOOQ/issues/12703

This proves once more that ad-hoc contributors don't have the big picture. To illustrate this a bit more, I'll comment on your suggested changes.

jOOQ/pom.xml
```diff
@@ -115,24 +115,11 @@
...
-
-
-
-
-
-
-
-
-
-
```
lukaseder, line 135: Please don't remove this whitespace, as documented in the contribution guidelines: https://github.com/jOOQ/jOOQ/blob/main/CONTRIBUTING.md

jOOQ/pom.xml
```diff
@@ -115,24 +115,11 @@
...
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
+        </dependency>
```
lukaseder, line 122: Our tests are not open source, and this won't change. It's not an accident you didn't find a single test in our repository. See again: https://github.com/jOOQ/jOOQ/blob/main/CONTRIBUTING.md

jOOQ/src/main/java/org/jooq/tools/Convert.java
```diff
@@ -1219,14 +1219,16 @@ private static final String patchIso8601Time(String string) {
...
-                else if (!t && string.charAt(10) == 'T')
-                    return string.substring(0, 10) + " " + string.substring(11);
-
-            return string;
+        private static String patchIso8601Timestamp(String string, boolean t) {
+            if (t) {
+                if (string.contains("T")) {
+                    return string;
+                } else {
+                    return string.replaceFirst(" ", "T");
```
lukaseder, line 1227: This is a regular expression replacement and despite there being an optimisation for expressions that are "obviously" not regular expressions, there is an `int[]` allocation in many JDK versions (haven't checked 17 yet), which is very costly.

PR #12733, opened 2021-12-10
Update log4j to 2.15.0 because of CVE-2021-44228

snuyanzin:
Update log4j2 version to 2.15.0 , because log4j2 have a RCE zero day.
more details at https://www.lunasec.io/docs/blog/log4j-zero-day/

### DIFF
--- a/jOOQ-examples/jOOQ-academy/pom.xml
+++ b/jOOQ-examples/jOOQ-academy/pom.xml
@@ -45,7 +45,7 @@
         <dependency>
             <groupId>org.apache.logging.log4j</groupId>
             <artifactId>log4j-slf4j-impl</artifactId>
-            <version>2.11.0</version>
+            <version>2.15.0</version>
         </dependency>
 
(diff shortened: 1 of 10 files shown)


lukaseder:
Thanks a lot for your report. I had seen the CVE yesterday and planned for doing this later, today. All the better if there's already a PR.

However, I'll fix this slightly differently via a new managed dependency in the parent `pom.xml`, so the version doesn't have to be changed everywhere.

Only examples on the github repository are affected. We don't otherwise have a direct dependency on log4j anymore, only on slf4j, so no new releases are required to fix this.

lukaseder:
Alternative fix here: https://github.com/jOOQ/jOOQ/commit/d07c1ad2e5bc9d36346d30c2b9b5457fc1dd94cb

lukaseder:
Even better, let's upgrade the examples to log4j 2.16.0: https://logging.apache.org/log4j/2.x/changes-report.html#a2.16.0

PR #12735, opened 2021-12-10
Adds @JvmStatic annotation to expose reference instance to other JVM …

mattjohnson:
This PR adds `@JvmStatic` annotation to expose reference instance to other JVM languages for improved interoperability for KotlinGenerator.

Other JVM languages like Java or Groovy can't access Kotlin companion object properties or functions unless the `@JvmStatic` compiler annotation is present. This will instruct the Kotlin compiler that additional static getter/setter methods should be generated.

### DIFF
--- a/jOOQ-codegen/src/main/java/org/jooq/codegen/JavaGenerator.java
+++ b/jOOQ-codegen/src/main/java/org/jooq/codegen/JavaGenerator.java
@@ -3058,6 +3058,7 @@ public class JavaGenerator extends AbstractGenerator {
             out.println();
             out.println("public companion object {");
             out.javadoc("The reference instance of <code>%s</code>", udt.getQualifiedOutputName());
+            out.println("@JvmStatic");
             out.println("public val %s: %s = %s()", getStrategy().getJavaIdentifier(udt), className, className);
             out.println("}");
         }


lukaseder:
Why not just use the `JavaGenerator` in such a case?

mattjohnson:
Our production code is in Kotlin, but using Spock (Groovy) for tests.

lukaseder:
I'm not convinced this is a road worth taking:

1) There would be dozens of these annotations by kotlin/groovy or kotlin/xyz users, which are hard to maintain (e.g. you've done this only for UDTs, not for anything else. How do we judge whether this fix is sound?)
2) We'd have to test/maintain these, also when we evolve the API / code generator

Cost benefit is off here, at least as far as this particular PR is concerned. I'll be happy to discuss this more thoroughly in a new issue before looking at specific PRs, but I'm not convinced this is something that needs to be rushed.

mattjohnson:
Appreciate the consideration, and agree that only doing it for UDTs was a mistake (I should have been more thorough before opening a PR). I would expect it be done for all usages of companion objects. If jOOQ supports generator configuration it would be nice to have a flag to enable it.

lukaseder:
I would really like a deeper discussion about the problem, and the various possible solutions first. E.g. with pointers to best practices in this area, to how big of a problem this really is in the broader community.

The current approach sounds like busywork. And I'll get it wrong every time I touch the `KotlinGenerator`, because adding Kotlin/Groovy interoperability tests is a lot of work, so, more busywork for little value to the community (at least, that's how I currently perceive it).

In fact, why not request Groovy or Spock fix this thoroughly? That would fix it for every use case, not just the jOOQ `KotlinGenerator` use-case.

PR #12943, opened 2022-01-25
Use more common form of 'unofficial' in issue templates

perlun:
I was about to say "fix typo" at first, but `inofficial` is actually [present in Merriam-Webster](https://www.merriam-webster.com/dictionary/inofficial). Regardless, it's definitely the more uncommon form and my browser (Firefox) marks `inofficial` as a typo, so... I think we are better off using the more common form here.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -22,4 +22,4 @@
 - Java:
 - Database (include vendor):
 - OS:
-- JDBC Driver (include name if inofficial driver):
+- JDBC Driver (include name if unofficial driver):
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -23,4 +23,4 @@ Please describe the possible workarounds you've implemented to work around the l
 - Java:
 - Database (include vendor):
 - OS:
-- JDBC Driver (include name if inofficial driver):
+- JDBC Driver (include name if unofficial driver):
--- a/.github/ISSUE_TEMPLATE/question.md
+++ b/.github/ISSUE_TEMPLATE/question.md
@@ -15,4 +15,4 @@ Before asking your question, please check out the FAQ too: https://www.jooq.org/
 - Java:
 - Database (include vendor):
 - OS:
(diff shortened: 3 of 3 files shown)


lukaseder:
Since it's valid English, I prefer to not change it.

perlun:
> Since it's valid English, I prefer to not change it.

Well, it's ultimately up to you of course. I personally tend to prefer mainstream forms in most cases (I did a quick google and `unofficial` has about 10 times the number of search hits, shows a dictionary definition when googling etc, which is not the case with `inofficial`). So it's less likely to cause people to stop and think "hey, is this misspelled or what?" => easier to focus on the more important things, like the question/bug report/feature request they were really _intended_ to focus on. :slightly_smiling_face:

perlun:
For the record, I still think this is a mild annoyance. My browser marks it as a typo when creating new issues and I'm reluctant to add an exception for this word.

[Sarah Madden writes](https://qr.ae/pvPCw4) (on Quora):

> The difference between “inofficial” and “unofficial” is that “inofficial” is not a word; only “unofficial” is an official English word. Even Chrome’s spellcheck flagged “inofficial” as being deficient in the spelling arena. (However, in an alternate universe, if it were an actual word, it would probably mean the same thing as “unofficial.”)
>
> EDIT: The word “unofficial” is appropriate in American English. I have no clue whether or where “inofficial” would be used.

Rickard Westman [summarizes it even better](https://qr.ae/pvPCbX), I think (emphasis is mine):

> I’m not a native speaker of English, and everyone else is saying that it’s not an English word, but I note that Merriam-Webster’s dictionary actually has it as a synonym for unofficial:
>
> [Definition of INOFFICIAL](https://www.merriam-webster.com/dictionary/inofficial)
>
> Can’t find it in any other reputable dictionary, though, so **perhaps it’s just a really uncommon/obscure synonym that is better avoided**?

lukaseder:
Workaround: make a blank issue:

image

lukaseder:
OK, I'll do this. I can see it's very important to you.

perlun:
> OK, I'll do this. I can see it's very important to you.

:slightly_smiling_face: :+1: Perhaps more important than what is healthy. But thanks for listening! I just happened to be reminded about it when creating a new issue someday ago.

Wishing you a nice weekend!

Issue #12992, opened 2022-02-04
`indexes = true` in code generation should enable `tables` creation automatically


perlun:
### Steps to reproduce the problem
With the following `generate` config snippet:

```groovy
generate {
    deprecated = false
    fluentSetters = true
    immutablePojos = true
    records = false
    tables = false // <-- this is the problem
}
```

...and `indexes` using the default setting (`true`), the code generator generates an `Indexes` class which doesn't compile:

![image](https://user-images.githubusercontent.com/630613/152531605-b1e5c312-36e1-4790-b50b-a52aab62480a.png)

The reason for this is simple; from what I can tell, the `Indexes` code is dependent on `tables` being generated. In other words, just like `records` or `daos` causing `tables` to be auto-enabled, the same thing should happen for `indexes`.

Since this seems to be trivial to fix, I'll give it a look and see if I can get a fix submitted. (it's Friday, after all. :sunglasses:)

### Versions

- jOOQ: 3.16.2
- Java: Java 8
- Database (include vendor): PostgreSQL
- OS: Debian GNU/Linux
- JDBC Driver (include name if inofficial driver): `org.postgresql:postgresql`

lukaseder:
Thanks for your report and PR. Fixed in jOOQ 3.17.0, 3.16.4 (https://github.com/jOOQ/jOOQ/issues/12995), and 3.15.8 (https://github.com/jOOQ/jOOQ/issues/12996)

lukaseder:
The suggested fix has introduced a curious regression in a different code generation test:

```xml
<generate>
    <relations>false</relations>
    <tables>false</tables>
    <routines>false</routines>
    <udts>false</udts>
    <sequences>false</sequences>
    <links>false</links>
    <queues>false</queues>
    <records>false</records>
    <pojos>true</pojos>
</generate>
```

This configuration is supposed to generate POJOs only, but now also generates tables because it doesn't specifically configure `<indexes>` (defaulting to `true`). The test generates a table without any index, so enabling the `<tables>` flag in this edge case seems wrong

lukaseder:
Edge case fixed and backported again to jOOQ 3.17.0, 3.16.4 (https://github.com/jOOQ/jOOQ/issues/12995), and 3.15.8 (https://github.com/jOOQ/jOOQ/issues/12996)

PR #12993, opened 2022-02-04
[jOOQ/jOOQ#12992] Automatically enable table generation if indexes are enabled

perlun:
Fixes jOOQ/jOOQ#12992.

### DIFF
--- a/jOOQ-codegen/src/main/java/org/jooq/codegen/JavaGenerator.java
+++ b/jOOQ-codegen/src/main/java/org/jooq/codegen/JavaGenerator.java
@@ -459,7 +459,8 @@ public class JavaGenerator extends AbstractGenerator {
         log.info("  table-valued functions", generateTableValuedFunctions());
         log.info("  tables", generateTables()
               + ((!generateTables && generateRecords) ? " (forced to true because of <records/>)" :
-                ((!generateTables && generateDaos) ? " (forced to true because of <daos/>)" : "")));
+                ((!generateTables && generateDaos) ? " (forced to true because of <daos/>)" :
+                ((!generateTables && generateIndexes) ? " (forced to true because of <indexes/>)" : ""))));
         log.info("  udts", generateUDTs());
         log.info("  relations", generateRelations()
             + ((!generateRelations && generateTables) ? " (forced to true because of <tables/>)" :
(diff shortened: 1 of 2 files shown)


jOOQ-codegen/src/main/java/org/jooq/codegen/JavaGenerator.java
```diff
@@ -459,7 +459,8 @@ public final void generate(Database db) {
         log.info("  table-valued functions", generateTableValuedFunctions());
         log.info("  tables", generateTables()
               + ((!generateTables && generateRecords) ? " (forced to true because of <records/>)" :
-                ((!generateTables && generateDaos) ? " (forced to true because of <daos/>)" : "")));
+                ((!generateTables && generateDaos) ? " (forced to true because of <daos/>)" :
+                ((!generateTables && generateIndexes) ? " (forced to true because of <indexes/>)" : ""))));
```
perlun, line 463: (I looked but I couldn't find any tests to update to ensure the new behavior works correctly, I presume they are elsewhere. Nonetheless, should be pretty much a no-brainer anyway.)

lukaseder: Thanks for your PR. The tests are not open source, see https://github.com/jOOQ/jOOQ/blob/main/CONTRIBUTING.md

perlun: Thanks. :+1: I suspected this could be the case in fact. I read the whole CONTRIBUTING guide now, thanks for the link - it'll help me balance between "fix myself" and "only file an issue" moving forward.

lukaseder: > it'll help me balance between "fix myself" and "only file an issue" moving forward.

Yeah, non-trivial fixes are probably not practical given the lack of access to tests... Also, a lot of things aren't as trivial as they may first seem.

perlun: Yep, true that. :+1: Btw, I noticed (after reading the fine CONTRIBUTING guide) that I should have added my own name to `jOOQ/src/main/resources/META-INF/ABOUT.txt` as well. :see_no_evil: Feel free to fix it if you like.

lukaseder: > Feel free to fix it if you like.

It's optional, only if you want to

perlun: Yes, I wouldn't mind. :+1: https://github.com/jOOQ/jOOQ/pull/13012

Issue #13008, opened 2022-02-07
Compilation error in KotlinGenerator output when implicit join path cache conflicts with keyword


owyke:
### Expected behavior
Expected to get Kotlin code that compiles.
### Actual behavior
The name of the one to many mapping is an invalid identifier which fails compilation.

### Steps to reproduce the problem
Running jOOQ codegen with `org.jooq.codegen.KotlinGenerator` having a one to many relation named as a kotlin keyword results in a compilation error.

See MCVE @ https://github.com/owyke/jOOQ-mcve

### Versions

- jOOQ: 3.16.3
- Java: N/A
- Database (include vendor): N/A
- OS: N/A
- JDBC Driver (include name if inofficial driver): N/A

lukaseder:
Thanks a lot for your report and MCVE. I can reproduce the problem. Will investigate.

lukaseder:
Thanks for your suggested PR. I agree with your fix.

Would you mind signing the transfer of rights agreement which allows for dual licensing your fix? https://www.jooq.org/legal/contributions

owyke:
Glad to be able to contribute to such an awesome project.
I just signed the agreement.

lukaseder:
Thanks, @owyke, and thanks again for your report!

lukaseder:
Fixed in jOOQ 3.17.0, 3.16.4 (https://github.com/jOOQ/jOOQ/issues/13019), and 3.15.8 (https://github.com/jOOQ/jOOQ/issues/13020)

lukaseder:
This fix was incomplete at the time, as the assumption was that backticks `` ` `` are introduced to support reserved words as identifiers / method names, etc. But they're also used for quoting of identifiers containing special characters, which is still not supported:

- https://github.com/jOOQ/jOOQ/issues/18654

PR #13009, opened 2022-02-07
[#13008] Use kotlin lazy delegate when mapping one to many relation

owyke:
This changes the api slightly. Another solution is possible (https://github.com/jOOQ/jOOQ/pull/13010) although this approach is more consistent with code generated for  scala.


lukaseder:
I prefer this solution, which doesn't change things as much: https://github.com/jOOQ/jOOQ/pull/13010

PR #13303, opened 2022-03-17
Fix #13302: Reverse order of nvl in H2 TYPE_NAME selection

### LINKED ISSUE #13302: H2TableDefinition with H2 2.0+ produces wrong code generation output when the run mode is Oracle
### Expected behavior

When querying the schema for an H2 table (in order to use a Generator against an H2 database), I expect that all types are appropriately determined.

### Actual behavior

It seems that all types are determined to be array types. I believe this is because of an incorrect `nvl` statement in the `H2TableDefinition`:

https://github.com/jOOQ/jOOQ/blob/c3a739680b093870595699e46c773fee542126ef/jOOQ-meta/src/main/java/org/jooq/meta/h2/H2TableDefinition.java#L114-L117

If my understanding of `nvl` is correct, then it picks the first non-null value. However, because of the `concat` with `" ARRAY"` the first value is always non-null, and therefore all types get labeled as arrays.

I've verified this by just setting up break points and seeing that no matter what I do, the definedType set in the DefaultColumnDefinition is set to:

```
DataType [ t=ARRAY]
```

And the generated code always looks like the following:

```
    /**
     * The column <code>PUBLIC.INVENTORY.ID</code>.
     */
    public final TableField<InventoryRecord, Object[]> ID = createField(DSL.name("ID"), org.jooq.impl.DefaultDataType.getDefaultDataType("\"\"").getArrayDataType(), this, "");
```

### Steps to reproduce the problem

Run `GenerationTool` against any H2 2.0+ database.

### Versions

- jOOQ: 3.16.3 (planning to upgrade to 3.16.5 as soon as I can)
- Java: 11
- Database (include vendor): H2 embedded 2.1.210
- OS: OSX
- JDBC Driver (include name if inofficial driver): H2

I'm happy to commit a fix here, I believe it's just reversing the order of arguments in the nvl.

mglazer:


lukaseder:
Thanks for your suggestion.

I'm afraid, I don't see how this fixes anything rather than making things worse. If `ELEMENT_TYPES.DATA_TYPE` is non-null, that means we *do* have an array type at hand, and we *must* place this expression first in the `NVL()` function. You just moved it out of the way because it was wrong for you, but that's hardly a viable fix.

I'm rejecting this PR. Let's discuss directly on the issue.

PR #13537, opened 2022-05-07
Automatically register PlainSQLChecker and SQLDialectChecker.

db2222:
Added the ```AutoService``` annotation so that META-INF\services\javax.annotation.processing.Processor is generated for the checkers.

This has 2 benefits:
- It is not necessary anymore to define the ```annotationProcessor``` in the ```pom.xml```. It is automatically detected. So only the dependency needs to be added.
- More importantly it does not break other frameworks that also use annotation processing. For instance our company uses Lombok and after adding ```PlainSQLChecker``` like it is documented in https://www.jooq.org/doc/latest/manual/tools/checker-framework/#plainsqlchecker compilation broke as ```log``` from Lombok wasn't found anymore. This is understandable as using an ```annotationProcessors``` block means that it simply should use the explicitly defined and not auto detect. Somebody not knowing the details doesn't know what's going on as Lombok states to simply add the dependency and that's it (see https://projectlombok.org/setup/maven). At first it's unclear which ```annotationProcessors``` need to be added for Lombok to work again as this is not normally needed.

For reference:
- https://www.baeldung.com/java-annotation-processing-builder#5-using-the-google-auto-service-library

If this is ok, can it please also be merged to version-3.16.0-branch.

Thanks :-)

### DIFF
--- a/jOOQ-checker/src/main/java/org/jooq/checker/PlainSQLChecker.java
+++ b/jOOQ-checker/src/main/java/org/jooq/checker/PlainSQLChecker.java
@@ -39,11 +39,14 @@ package org.jooq.checker;
 
 import static com.sun.source.util.TreePath.getPath;
 
+import com.google.auto.service.AutoService;
 import org.jooq.PlainSQL;
 
 import org.checkerframework.framework.source.SourceVisitor;
 
 import com.sun.source.tree.MethodInvocationTree;
+
+import javax.annotation.processing.Processor;
 
 /**
  * A checker to disallow usage of {@link PlainSQL} API, except where allowed
@@ -51,6 +54,7 @@ import com.sun.source.tree.MethodInvocationTree;
  *
  * @author Lukas Eder
  */
+@AutoService(Processor.class)
 public class PlainSQLChecker extends AbstractChecker {
 
     @Override
--- a/jOOQ-checker/src/main/java/org/jooq/checker/SQLDialectChecker.java
+++ b/jOOQ-checker/src/main/java/org/jooq/checker/SQLDialectChecker.java
@@ -39,6 +39,7 @@ package org.jooq.checker;
 
 import static com.sun.source.util.TreePath.getPath;
 
+import com.google.auto.service.AutoService;
 import org.jooq.Require;
 import org.jooq.SQLDialect;
 import org.jooq.Support;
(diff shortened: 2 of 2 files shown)


db2222:
@lukaseder Any feedback? :-) Is anything wrong or missing in this pull request? Happy to extend/correct.

lukaseder:
Thanks for the effort appreciated. I'll look into this eventually.

lukaseder:
> Just so that it's documented for anyone who needs it.

It's already documented now. If every PR and issue on github warranted a section in the manual, we'd have a 5000 page manual.

If you think this documentation is insufficient, why not add a Stack Overflow Q&A? It's explicitly encouraged:
https://stackoverflow.blog/2011/07/01/its-ok-to-ask-and-answer-your-own-questions/

db2222:
Like I wrote. It was just a suggestion for a 1 liner with the intention of avoiding questions/support in the future.
But why should I stress about it :-)

nicchongwb:
Hi @db2222 , apologies for reopening this issue. I do face this issue too on projects that uses lombok annotation processor and jooq-checker annotation processor. Can I just check with you, based on the changes you made in this PR, it is also possible to only autoservice `PlainSQLChecker` only? 

Let's say these changes are made. Does this mean when applying jooq-checker, we no longer need to specify `annotationProcessor("org.jooq.checker...")` ?

db2222:
Hi @nicchongwb,

I hope I have understood your problem correctly.

Yes, it would be possible to only annotate ```PlainSQLChecker``` with ```AutoService```.
But this change would need to be be accepted by @lukaseder.
I personally don't need it anymore as there is a workaround.

Alternatively you would have to build your own version of jOOQ with this change. But I personally wouldn't recommend that at all.

Can't you use the solution I mentioned at the bottom of my comment from 18th May of 2022?

You just have to add the following construct in the ```pom.xml``` if you have this specific problem with ```Lombok``` & ```jOOQ```:
```
<plugin>
  <artifactId>maven-compiler-plugin</artifactId>
  <version>3.11.0</version>
  <configuration>
    <source>1.8</source>
    <target>1.8</target>
    <fork>true</fork>
    <annotationProcessors>
      <annotationProcessor>org.jooq.checker.PlainSQLChecker</annotationProcessor>
      <annotationProcessor>lombok.launch.AnnotationProcessorHider$AnnotationProcessor</annotationProcessor>
      <annotationProcessor>lombok.launch.AnnotationProcessorHider$ClaimingProcessor</annotationProcessor>
    </annotationProcessors>
    <compilerArgs>
      <arg>-Xbootclasspath/p:1.8</arg>
    </compilerArgs>
  </configuration>
</plugin>
```

I described it more generic in my comment of last year so it's clearer if someone has the problem with another package than ```Lombok```.
Also maybe ```Lombok``` could change their ```annotationProcessors``` in the future so different values maybe are then needed.

Hope this helps 😀

Issue #13635, opened 2022-06-03
Test container does not close in example


kny78:
### Expected behavior
Test container is closed after the build

### Actual behavior
The test container is living after the build

### Steps to reproduce the problem
mvn install
Check for running containers.

### Versions

- jOOQ: 3.16.6
- Java: 17
- Database (include vendor): PostgreSQL
- OS: Linux
- JDBC Driver (include name if inofficial driver):

lukaseder:
Fixed as per your PR: https://github.com/jOOQ/jOOQ/pull/13636.

Thanks again

PR #13636, opened 2022-06-03
Close testcontainer in example

kny78:
Added code to close test-container.

Fixes [jOOQ/jOOQ#13635]


lukaseder:
Thanks for your suggestion. I agree that this could be one way to do it.

Issue #12021, opened 2021-06-17
Address known limitations of MULTISET support


lukaseder:
jOOQ 3.15 introduced support for nested collection emulations via JSON/XML: #3884

This issue addresses known limitations of the 3.15 implementation, including:

### Bugs

- [x] https://github.com/jOOQ/jOOQ/issues/12135
- [x] https://github.com/jOOQ/jOOQ/issues/12136
- [x] The SQL Server `JSON` encoding doesn't support multiple identical keys yet because our JSON parser can't handle it, see https://github.com/jOOQ/jOOQ/issues/12101.

### Limitations, missing features

- [x] Support combining nesting `ROW` in `MULTISET`
  - [x] The nested `ROW` expressions introduced in jOOQ 3.15 via #11812 must also be emulated using `JSON` or `XML` when being wrapped in a `MULTISET`. See: https://github.com/jOOQ/jOOQ/issues/12116
  - [x] The same is true for embeddable records, which are typed nested `ROW` expressions: https://github.com/jOOQ/jOOQ/issues/12237
  - [x] When a `ROW` contains a `MULTISET`, then it must render itself according to the `MULTISET` emulatio rules. See also: https://github.com/jOOQ/jOOQ/issues/12706
- [x] Support nesting `ROW` in `MULTISET_AGG`
- [x] Support nesting `MULTISET` in `MULTISET_AGG` (already works)
- [x] Aliasing of `MULTISET` expressions as well as content should not be necessary but automatic.
- [x] Data types
  - [x] `TIMESTAMP` https://github.com/jOOQ/jOOQ/issues/12130
- [x] We currently don't test what happens when XML/JSON content is nested in MULTISET, i.e.
  - [x] JSON content in a MULTISET/JSON emulation

lukaseder:
Nesting `MULTISET` in `MULTISET_AGG` seems to work already, nothing needs to be done.

lukaseder:
Aliasing of `MULTISET` content is no longer necessary, because for serialisation reasons, we're already auto-aliasing all contents

lukaseder:
Further work in jOOQ 3.17 is planned as https://github.com/jOOQ/jOOQ/issues/12734

hantsy:
I encountered an issue to use `multiset` in an unionAll sql, the `multiset` query can not be processed as other native datatype fields.https://stackoverflow.com/questions/77306180/how-to-create-a-union-query-with-multiset-fields-in-jooq

PR #13641, opened 2022-06-07
[jOOQ/jOOQ#12021] Multiset limitations are now being tracked by issue…

tivrfoa:
… 13589


lukaseder:
Thanks for the pointer. I see your point. I'd rather just remove the reference, though, rather than having to constantly update it...

Issue #13811, opened 2022-07-18
Not correctly mapping 'native query' (resultQuery()) fetch result (PostgreSQL integration)


mgrebnev:
### Your question:

**Example:**

I have simple table and record generated by jooq-codegen plugin:

```
/**
 * This class is generated by jOOQ.
 */
@SuppressWarnings({ "all", "unchecked", "rawtypes" })
public class RoleTable extends TableImpl<RoleEntity> {

    private static final long serialVersionUID = 1L;

    /**
     * The reference instance of <code>ROLE</code>
     */
    public static final RoleTable ROLE = new RoleTable();

    /**
     * The class holding records for this type
     */
    @Override
    public Class<RoleEntity> getRecordType() {
        return RoleEntity.class;
    }

    /**
     * The column <code>ROLE.ID</code>.
     */
    public final TableField<RoleEntity, Integer> ID = createField(DSL.name("ID"), SQLDataType.INTEGER.nullable(false).identity(true), this, "");
    
    ...
    ...
```
Also, i have same table:

```
CREATE TABLE ROLE (
   id serial primay key,
   name varchar not null
);
```

I need execute native query, simple select for getting rows from database:

`DSL.resultQuery("SELECT * FROM role").fetchInto(ROLE);`

I getting result in console 
```
Id  |   name
1   |   TEST_ROLE
```
Important, that columns in lower case
Next, after fetchInto, i have entity with **null** values

BUT, if i try to write query in jooq style:

```
DSL
  .selectFrom(ROLE)
  .fetchInto(RoleRecord.class);
```
i see correctly result with true row data from database and getting in logs
```
ID  |   NAME
1   |   TEST_ROLE
```
See, columns in uppercase - and mapping correctly finally

**What i think?**

**Firstly**
Codegen plugin generate, table, and fields in uppercase - i use LiquibaseDatabase:

part of configuration
```
<database>
    <name>org.jooq.meta.extensions.liquibase.LiquibaseDatabase</name>
    <properties>
        <property>
            <key>sort</key>
            <value>liquibase</value>
        </property>
        <property>
            <key>dialect</key>
            <value>Postgres</value>
        </property>
        <property>
            <key>scripts</key>
            <value>db/changelog/scripts/liquibase_package.xml</value>
        </property>
        <property>
            <key>defaultNameCase</key>
            <value>lower</value>
        </property>
    </properties>
</database>
```
any value in **defaultNameCase** doesn't has correctly result

[...]

PR #13815, opened 2022-07-20
[jOOQ#13811] Add setting 'fieldsInLowerCase' for change case fields i…

mgrebnev:
Related problem described in https://github.com/jOOQ/jOOQ/issues/13811

I try to reverse engineering codegen plugin and class `LiquibaseDatabase`
I think that developer may set lower or upper case fields in generated tables and entities

**How enable setting?**
```
<database>
  <name>org.jooq.meta.extensions.liquibase.LiquibaseDatabase</name>
  <properties>
    ....
      <property>
          <key>fieldsInLowerCase</key>
          <value>true</value>
      </property>
  </properties>
</database>
```

In true value, we have dsl fields in generated entities in lowercase:

```
public final TableField<RoleEntity, Integer> ID = createField(DSL.name("id"), SQLDataType.INTEGER.nullable(false).identity(true), this, "");

/**
 * The column <code>ROLE.CODE</code>.
 */
public final TableField<RoleEntity, String> CODE = createField(DSL.name("code"), SQLDataType.VARCHAR(2147483647).nullable(false), this, "");
```

if parameter **fieldsInLowerCase** is empty or false, we have default settings:
```
public final TableField<RoleEntity, Integer> ID = createField(DSL.name("ID"), SQLDataType.INTEGER.nullable(false).identity(true), this, "");

/**
 * The column <code>ROLE.CODE</code>.
 */
public final TableField<RoleEntity, String> CODE = createField(DSL.name("CODE"), SQLDataType.VARCHAR(2147483647).nullable(false), this, "");
```

I try to execute 
`DSL.resultQuery("SELECT id FROM role").fetchInto(ROLE);`
and getting expected correctly result

Also, default query work correctly:
`DSL.resultQuery("SELECT * FROM role").fetchInto(ROLE);`

Thanks a lot for checking this PR

### DIFF
--- a/jOOQ-codegen/src/main/java/org/jooq/codegen/JavaGenerator.java
+++ b/jOOQ-codegen/src/main/java/org/jooq/codegen/JavaGenerator.java
@@ -38,6 +38,7 @@
 package org.jooq.codegen;
 
 
+import static java.lang.Boolean.parseBoolean;
 import static java.util.Arrays.asList;
 import static java.util.function.Function.identity;
 import static java.util.stream.Collectors.counting;
@@ -84,6 +85,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
+import java.util.Properties;
 import java.util.Set;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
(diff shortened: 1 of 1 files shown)


lukaseder:
Thanks for showing your suggestion on how to work around https://github.com/jOOQ/jOOQ/issues/13811. I can see how this works around the problem for you.

However, this approach isn't robust enough for a core jOOQ code generation feature. There are two possible extension points that make sense:

1. A way to influence how the `LiquibaseDatabase` is simulated, i.e. fixing the problem at the root cause (which is that something intended for PostgreSQL is currently being executed on H2). This is already on the roadmap somewhere.
2. A way to influence *all* identifiers in `jOOQ-meta`, i.e. a way to translate arbitrary `Definition::getInputName` to `Definition::getOutputName`. We already have a feature for this called catalog and schema mapping: https://www.jooq.org/doc/latest/manual/code-generation/codegen-advanced/codegen-config-database/codegen-database-catalog-and-schema-mapping. In a way, we could extend the feature to allow for transforming *all* identifiers, not just those of catalogs and schemas. I've created a feature request for this: https://github.com/jOOQ/jOOQ/issues/13817

Given that this case sensitivity issue is quite frequent among users of the `LiquibaseDatabase`, `DDLDatabase`, and `JPADatabase`, I'm positive that it will be fixed thoroughly, soon (hopefully in jOOQ 3.18). In the meantime, you've found a viable workaround using your patch.

mgrebnev:
Thanks!

PR #13990, opened 2022-09-14
[#12230] Add Settings.includeSqlStringInException to allow disabling full sql query in DataAccessException

### LINKED ISSUE #12230: Consider not putting the SQL string into DataAccessException by default
jOOQ, by default, enhances its `DataAccessException::getMessage` with the SQL string for better debugging context. This is *very* useful, but at the same time creates a security risk in environments with highly sensitive personal information, where log files have to be scrutinised and anonymised. An accidental leak of inlined bind value could be a GDPR issue in such an environment. As such, our defaults might be ill chosen.

An alternative would be to offer the SQL string as a property of `DataAccessException` for those interested, but not include it in the exception message by default.

Here are different error messages form RDBMS vendors going from Oracle (most obscure) to PostgreSQL (complete information): https://twitter.com/lukaseder/status/1416358174119104514

A decision has not yet been made about the defaults, but clearly, it should be possible to opt in/out of the current behaviour.

vooft:
Fixes #12230 
`Settings` object seems to be available from pretty much any place where `Tools.translate` is called.
Added a simple boolean switch that defaults to `true` to preserve the current behavior.

Another option would be passing the whole `Settings` object into `translate`.

### DIFF
--- a/jOOQ/src/main/java/org/jooq/conf/Settings.java
+++ b/jOOQ/src/main/java/org/jooq/conf/Settings.java
@@ -276,6 +276,8 @@ public class Settings
     @XmlSchemaType(name = "string")
     protected ThrowExceptions throwExceptions = ThrowExceptions.THROW_ALL;
     @XmlElement(defaultValue = "true")
+    protected Boolean includeSqlStringInException = true;
+    @XmlElement(defaultValue = "true")
     protected Boolean fetchWarnings = true;
     @XmlElement(defaultValue = "0")
     protected Integer fetchServerOutputSize = 0;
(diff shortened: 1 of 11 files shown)


lukaseder:
Thanks for your suggestion.

> Another option would be passing the whole `Settings` object into `translate`.

There are tons of options to implement anything, each with their own tradeoffs. A PR implies maintainers have to prioritise not only a certain feature request, but also a *specific* implementation suggestion review.

By default, most PRs will be rejected on the grounds that such re-prioritisation from the community doesn't scale, see:

For example, I could now try to explain to you how `Settings` is generated from the XSD, to do it "right." We could now discuss whether passing `boolean` flags around to exceptions is the right design, or if the exception should have access to more context, likewise with the `Settings`.

But why prioritise this? Why did you choose to implement this? Please discuss things first, before sending a PR.

lukaseder:
... I just realised, you already commented on here: https://github.com/jOOQ/jOOQ/issues/12230#issuecomment-1164275471

Sure, there might be a `Settings`, but again, there are other options. I currently don't want to prioritise neither the feature itself, nor potential solutions...

vooft:
Thank you @lukaseder, this PR was just an attempt to demonstrate that it is technically possible to have a fix via `Settings`, there are really no places where `Tools.translate` is invoked and we don't have access to this flag.

PR #14026, opened 2022-09-24
Fix "Use case" -> "Possible solution" field name in feature request template

vlsi:
feature_request.yaml had two "use case" fields

Credits go to @pmouawad https://github.com/apache/jmeter/pull/5710#pullrequestreview-1119291869


lukaseder:
Thanks a lot!

PR #14033, opened 2022-09-27
Fix incorrect grammar in XSD documentation

perlun:
This wasn't me, the IDE made me do it. :wink: 

(I think "helps in identifying" would also be correct.)

image


lukaseder:
Thanks

PR #14433, opened 2023-01-01
feat : Added pessimistic lock test for springboot example

hgs-study:
I had a hard time finding an example using Jooq's pesimistic lock on the springboot. So I made a test of the stock quantity by referring to the official website. I'd appreciate it if you could review it!
jooq_pessimistic_lock_test


lukaseder:
Thanks for your suggestion.

I don't see the value of presenting how to use arbitrary SQL feature X (`FOR UPDATE`) with arbitrary integration platform Y (Spring Boot). You could generate a huge matrix with millions of entries for all possible combinations of X and Y, and there would still be someone having "a hard time finding an example" for a new combination X' and Y'.

The only value of a Spring Boot / jOOQ example is to show how to set up jOOQ with Spring Boot.

hgs-study:
Thank you for your kind words. I will try to contribute again next time.

Issue #14699, opened 2023-02-23
PostgreSQL interval parser handle fractional seconds with interval style iso_8601


rocketraman:
### Expected behavior

I have a PostgresQL database with `intervalstyle` set to `iso_8601` (see https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-INTERVAL-OUTPUT).

The parser for this type does not handle fractional seconds -- it assumes all components of the time value are integers:

https://github.com/jOOQ/jOOQ/blob/5ebcc5c0531e3eeee95ce809ef9602c5a2e0974e/jOOQ/src/main/java/org/jooq/util/postgres/PGInterval.java#L98

An example input to this function is `PT6M32.417S`.

### Actual behavior

```
Exception in thread "main" org.jooq.exception.DataAccessException: SQL[...]; Error while reading field: "public"."table"."field", at JDBC index: 83
	at org.jooq_3.17.6.POSTGRES.debug(Unknown Source)
	at org.jooq_3.17.6.POSTGRES.debug(Unknown Source)
	at org.jooq.impl.Tools.translate(Tools.java:3313)
	at org.jooq.impl.DefaultExecuteContext.sqlException(DefaultExecuteContext.java:678)
	at org.jooq.impl.CursorImpl$CursorIterator.fetchNext(CursorImpl.java:1464)
	at org.jooq.impl.CursorImpl$CursorIterator.hasNext(CursorImpl.java:1425)
	at org.jooq.impl.CursorImpl.fetchNext(CursorImpl.java:238)
	at org.jooq.impl.AbstractCursor.fetch(AbstractCursor.java:177)
	at org.jooq.impl.AbstractCursor.fetch(AbstractCursor.java:88)
	at org.jooq.impl.AbstractResultQuery.execute(AbstractResultQuery.java:265)
	at org.jooq.impl.AbstractQuery.execute(AbstractQuery.java:341)
	at org.jooq.impl.AbstractResultQuery.fetch(AbstractResultQuery.java:290)
	at org.jooq.impl.SelectImpl.fetch(SelectImpl.java:2837)
	at org.jooq.impl.Tools$3$1.block(Tools.java:5854)
	at java.base/java.util.concurrent.ForkJoinPool.unmanagedBlock(ForkJoinPool.java:3463)
	at java.base/java.util.concurrent.ForkJoinPool.managedBlock(ForkJoinPool.java:3434)
[...]

PR #14700, opened 2023-02-23
Parse fractional seconds on postgresql ISO8601 intervals

rocketraman:


lukaseder:
Thanks for your suggestion.

As commented here: https://github.com/jOOQ/jOOQ/issues/14699#issuecomment-1443697758, in order to avoid any unnecessary deviation from the upstream source of this class (see https://github.com/pgjdbc/pgjdbc/blob/master/pgjdbc/src/main/java/org/postgresql/util/PGInterval.java), I suggest supplying this PR directly to pgjdbc.

Issue #14801, opened 2023-03-15
JavaGenerator creates broken code for arrays when kotlinNotNullPojoAttributes is true


xoliver:
### Expected behavior

Given a DB table:
```
create table if not exists public.something
(
    id              bigint                                 not null
            primary key,
    things      text[]                                 not null
);
```

and the flag kotlinNotNullPojoAttributes being active (as per [this issue's comment](https://github.com/jOOQ/jOOQ/issues/10212#issuecomment-1252351193) )

the generated Kotlin POJO looks like this:

```
@Suppress("UNCHECKED_CAST")
data class SomethingPojo(
    var id: Long,
    var things: Array<String?>
): Serializable {

   (...)

    override fun hashCode(): Int {
        val prime = 31
        var result = 1
        result = prime * result + this.id.hashCode()
        result = prime * result + Arrays.deepHashCode(things)
        return result
    } 
```

And it compiles.

### Actual behavior

This is what the POJO looks like:
```
@Suppress("UNCHECKED_CAST")
data class SomethingPojo(
    var id: Long,
    var things: Array<String?>
): Serializable {

   (...)

    override fun hashCode(): Int {
        val prime = 31
        var result = 1
        result = prime * result + this.id.hashCode()
        result = prime * result + things.deepHashCode(this.Arrays)
        return result
    } 
```

This code does not compile, because `deepHashCode` (in `hashCode()`) is not implemented for `Array` and also `this.Arrays` is undefined.

### Steps to reproduce the problem

Troubleshooting:

As part of issue https://github.com/jOOQ/jOOQ/issues/10212
[Commit](https://github.com/jOOQ/jOOQ/commit/b9ef00717e56f779069bec5c93ec3d39631fc2f4), two `println` lines [have an extra string interpolation placeholder](https://github.com/jOOQ/jOOQ/blob/main/jOOQ-codegen/src/main/java/org/jooq/codegen/JavaGenerator.java#L5785-L5788) which results in the wrong values being generated.

PR incoming, given it's a small fix.

### jOOQ Version

jOOQ professional 3.18.0

### Database product and version

PostgreSQL 12

### Java Version

openjdk version "17.0.2" 2022-01-18

### OS Version

OSX 13.12.1

### JDBC driver name and version (include name if unofficial driver)

postgresql:postgreql:42.4.0

lukaseder:
Thanks a lot for your report and for your suggested fix. Will merge and backport right away.

[...]

PR #14802, opened 2023-03-15
[#14801] Fix Java String generation for non-nullable arrays

xoliver:
Fixes https://github.com/jOOQ/jOOQ/issues/14801 by removing an extra parameters for String interpolation when generating Java code, which resulted in code that wouldn't compile.


xoliver:
Sorry for the force-push, I misspelled my own surname 💩

lukaseder:
Thanks a lot for your suggested fix. I agree with it.

PR #14854, opened 2023-03-26
refactor: Use diamond operator

timtebeek:
Figured run a quick static analysis fix recipe; Could run more, but wanted to keep this easy to review (or turn down). :)

### DIFF
--- a/jOOQ-meta/src/main/java/org/jooq/meta/jaxb/CatalogMappingType.java
+++ b/jOOQ-meta/src/main/java/org/jooq/meta/jaxb/CatalogMappingType.java
@@ -102,7 +102,7 @@ public class CatalogMappingType implements Serializable, XMLAppendable
 
     public List<SchemaMappingType> getSchemata() {
         if (schemata == null) {
-            schemata = new ArrayList<SchemaMappingType>();
+            schemata = new ArrayList<>();
         }
         return schemata;
     }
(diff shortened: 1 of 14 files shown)


lukaseder:
Thanks for your suggestion.

Many of those classes are generated either using XJC or our internal code generator, so the fix (if necessary at all) is a the wrong location. In the `AbstractContext` case, I think there was a subtle javac bug that was worked around this way.

> Could run more, but wanted to keep this easy to review (or turn down). :)

Thanks for that. The immediate reaction is usually to turn these down without review because they cause only work with no immediate benefit. In this case, sure, diamond operator may be slightly better in most instances for merely stylistic reasons, not logic reasons, let alone correctness ones, but then again, it wasn't all that bad before, and as I said, there are subtle or not so subtle reasons for the status quo.

I don't personally think static code analysis outside of actual compiler errors is useful. It tends to produce *a ton* of false positives, and thus busywork before creating value, *especially* in library code, which may be more challenging for the analyser than ordinary business logic.

As a general guideline, we have our contributing page here, showing also why most PRs are likely to be rejected if there's no prior discussion (and even then, they have little chance of acceptance):

timtebeek:
No problem at all! Understand your point of view and will not open more unless discussed beforehand, or minor bugs are involved.
Let me know if I can help in some other way in the future! :)

PR #15059, opened 2023-05-11
Maven and Maven build related updates

cstamas:
Changes:
* fix jooq-maven-plugin, it emits warnings in 3.9.x (deps in bad scope), set prerequisite (same as ASF Maven plugins: 3.2.5, adjust if needed)
* update plugins where applicable (others may be updated as well, but did not touch non-ASF Maven plugins)
* remove obsolete or old things
* set project encoding to UTF8
* enforce minimum "modern" Maven 3.8.x for build

Ultimate goal is to get rid of Maven 3.9.x plugin validation warnings when building a project that uses jooq-maven-plugin.

### DIFF
--- a/jOOQ-codegen-maven/pom.xml
+++ b/jOOQ-codegen-maven/pom.xml
@@ -13,6 +13,10 @@
     <name>jOOQ Codegen Maven</name>
     <packaging>maven-plugin</packaging>
 
+    <prerequisites>
+        <maven>3.2.5</maven>
+    </prerequisites>
+
     <licenses>
         <license>
             <name>Apache License, Version 2.0</name>
@@ -52,11 +56,6 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-plugin-plugin</artifactId>
-                <configuration>
-
-                    <!-- see http://jira.codehaus.org/browse/MNG-5346 -->
-                    <skipErrorNoDescriptorsFound>true</skipErrorNoDescriptorsFound>
-                </configuration>
                 <executions>
                     <execution>
                         <id>mojo-descriptor</id>
@@ -73,15 +72,6 @@
                         </goals>
                     </execution>
                 </executions>
-                
-                <!-- Workaround for https://issues.apache.org/jira/browse/MPLUGIN-369 -->
-                <dependencies>
-                    <dependency>
-                        <groupId>org.apache.maven.plugin-tools</groupId>
-                        <artifactId>maven-plugin-tools-annotations</artifactId>
-                        <version>3.6.1</version>
-                    </dependency>
-                </dependencies>
             </plugin>
             
             <plugin>
@@ -103,26 +93,20 @@
             <groupId>org.jooq</groupId>
             <artifactId>jooq-codegen</artifactId>
         </dependency>
-
         <dependency>
             <groupId>org.apache.maven</groupId>
             <artifactId>maven-plugin-api</artifactId>
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.apache.maven</groupId>
             <artifactId>maven-core</artifactId>
-
-            <!-- junit 4.8.2 sneaking in here -->
-            <exclusions>
-                <exclusion>
-                    <groupId>junit</groupId>
-                    <artifactId>junit</artifactId>
-                </exclusion>
-            </exclusions>
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.apache.maven.plugin-tools</groupId>
             <artifactId>maven-plugin-annotations</artifactId>
+            <scope>provided</scope>
         </dependency>
     </dependencies>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -19,6 +19,11 @@
     <url>http://www.jooq.org</url>
 
     <properties>
+        <!--
+        File encodings.
+        -->
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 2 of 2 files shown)


lukaseder:
Thanks a lot for your suggestion. I've created an issue to track what will be done: https://github.com/jOOQ/jOOQ/issues/15065

Please note that our CI has several tools already to keep dependencies as updated as possible:

- Dependabot / owasp plugin to keep an eye out for urgent upgrades due to vulnerabilities
- A call to `mvn versions:display-dependency-updates` to get a report of what non-urgent upgrades are possible

The decision when to upgrade is a delicate one, as we support various JDK versions, so we don't just upgrade because we can.

Additionally, unrelated changes should be avoided to minimise efforts of reviewing PRs as well as undesired side effects that aren't easy to predict (see our stance on PRs here: https://github.com/jOOQ/jOOQ/blob/main/CONTRIBUTING.md). This may or may not include the UTF 8 related changes, the enforcer changes, the resources plugin changes, etc, which I will reject.

lukaseder:
The actual dependency upgrades will be documented in https://github.com/jOOQ/jOOQ/issues/15065. Please let me know if that addresses your observed issues. If not, can you please create a new issue documenting the problems (ideally with a reproducer). I'll then review which parts of your additional suggestions might be worth including.

jOOQ-codegen-maven/pom.xml
```diff
@@ -13,6 +13,10 @@
     <name>jOOQ Codegen Maven</name>
     <packaging>maven-plugin</packaging>
 
+    <prerequisites>
+        <maven>3.2.5</maven>
+    </prerequisites>
+
```
lukaseder, line 19: Not sure why this is needed, won't be included in https://github.com/jOOQ/jOOQ/issues/15065

jOOQ-codegen-maven/pom.xml
```diff
@@ -103,26 +93,20 @@
...
-                    <groupId>junit</groupId>
-                    <artifactId>junit</artifactId>
-                </exclusion>
-            </exclusions>
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.apache.maven.plugin-tools</groupId>
             <artifactId>maven-plugin-annotations</artifactId>
+            <scope>provided</scope>
```
lukaseder, line 109: These various `provided` scopes don't really look correct to me? What is the rationale? If someone wants to build a plugin based on the jOOQ plugin, wouldn't the jOOQ plugin rather already expose its transitive dependencies, formally?

pom.xml
```diff
@@ -19,6 +19,11 @@
     <url>http://www.jooq.org</url>
 
     <properties>
+        <!--
+        File encodings.
+        -->
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
```
lukaseder, line 26: This change seems unrelated to the actual set of changes

pom.xml
```diff
@@ -611,10 +616,15 @@
         <finalName>${project.artifactId}-${project.version}</finalName>
         <pluginManagement>
             <plugins>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-enforcer-plugin</artifactId>
+                    <version>3.3.0</version>
+                </plugin>
```
lukaseder, line 623: We've lived very long without this, so I'd like to know why this would be needed / beneficial. Otherwise, I'd rather not include new plugins.

jOOQ-codegen-maven/pom.xml
cstamas: These provided deps are provided by Maven when plugin runs. If you don't put them in provided scope, maven will download them but not use them (as provided ones replaces them). Also, maven 3.9.x due this warns when plugin is executed

pom.xml
cstamas: Jooq emits warnings about encoding not set, this makes them go away

jOOQ-codegen-maven/pom.xml
lukaseder: OK, but that's just one more thing to keep an eye out for whether it's up to date or not, right? The Maven plugins probably have higher "meta standards" here, but for jOOQ, this isn't really so important. E.g. we don't have integration test checking if stuff works with various maven versions.

jOOQ-codegen-maven/pom.xml
lukaseder: Can you show a reproducer / log output?

lukaseder: OK, I can see it now *after* the dependency upgrades:

```
15:16:29,163 [WARNING]

Some dependencies of Maven Plugins are expected to be in provided scope.
Please make sure that dependencies listed below declared in POM
have set '<scope>provided</scope>' as well.

The following dependencies are in wrong scope:
 * org.apache.maven:maven-plugin-api:jar:3.9.2:compile
 * org.apache.maven:maven-model:jar:3.9.2:compile
 * org.apache.maven:maven-artifact:jar:3.9.2:compile
 * org.apache.maven:maven-core:jar:3.9.2:compile
 * org.apache.maven:maven-settings:jar:3.9.2:compile
 * org.apache.maven:maven-settings-builder:jar:3.9.2:compile
 * org.apache.maven:maven-builder-support:jar:3.9.2:compile
 * org.apache.maven:maven-repository-metadata:jar:3.9.2:compile
 * org.apache.maven:maven-model-builder:jar:3.9.2:compile
 * org.apache.maven:maven-resolver-provider:jar:3.9.2:compile
```

PR #15102, opened 2023-05-23
Add argument to provide coroutine context for the transactionCoroutine

trietsch:


lukaseder:
It's easy to check that this particular suggestion isn't backwards compatible. Try this:

```kotlin
suspend fun insertActorTransaction(): ActorRecord {
    return ctx.transactionCoroutine(::insertActor)
}
```

From here:
https://github.com/jOOQ/demo/blob/main/jOOQ-demo-oss/jOOQ-demo-kotlin/src/test/kotlin/org/jooq/demo/kotlin/Demo13Reactive.kt#L102

This no longer compiles when we have that `context: CoroutineContext = EmptyCoroutineContext` parameter. Also, given the fact that there had already been a replacement from `CoroutineScope` to `CoroutineContext` (whose history I'm unaware of), I'd like to avoid exposing users to such things.

Since the extension function is really very trivial, I think it is safe for now to expect users to write their own context aware extension function. Maybe we'll address this some time in the future, but I'd like to collect more user feedback first, and then, make a more informed decision.

trietsch:
Another option to make sure that that code compiles, is to make an overloaded function of `transactionCoroutine`, that has the following signature:
```kotlin
suspend fun <T> DSLContext.transactionCoroutine(context: CoroutineContext, transactional: suspend (Configuration) -> T): T {
    @Suppress("UNCHECKED_CAST")
    return transactionPublisher { c ->
        mono {
        mono(context) {
            transactional.invoke(c)
        }
    }.awaitFirstOrNull() as T
```

Then the original function would be unmodified. But we'll keep it in our own code for now. Thanks for reviewing!

lukaseder:
I thought of overloading indeed. I can't think of any syntax that would break this way, even though I've found lambda resolution with overloads treacherous terrain in the past in Java. Not sure about kotlin. It's certainly less risky.

PR #15363, opened 2023-07-06
fix deprecation version and replacement

robstoll:


lukaseder:
Hmm, you're right. I thought we didn't do this, but it's quite ubiquitous (and inconsistently so). Never mind. I just didn't want PRs to mix concerns, it takes much more time to review...

PR #15831, opened 2023-11-15
add table-column's comment to pojo's field

nianien:
now for jooq 3.18.7, pojo class generated is still missed the comment of table-column,  the pull request aims to add column's comment  to pojo's field.

for  example ,  sql script as follows: 
```h2
CREATE TABLE `user_info`
(
    `id`          bigint      NOT NULL AUTO_INCREMENT COMMENT 'user's id',
    `name`        varchar(16) NOT NULL DEFAULT '' COMMENT 'user's name',
    `phone`       varchar(20) NOT NULL DEFAULT '' COMMENT 'user's telephone',
    `email`       varchar(20) NOT NULL DEFAULT '' COMMENT 'user's email',
    PRIMARY KEY (`id`)
);
COMMENT ON TABLE `user_info` IS 'user information';
```

the pojo class generated:
```java
**
 * user's information
 */
@SuppressWarnings({ "all", "unchecked", "rawtypes" })
public class UserInfo implements Serializable {

    private static final long serialVersionUID = 1L;


    /**
     * user's id
     */
    private Long id;

    /**
     * user's name
     */
    private String name;

    /**
     * user's telephone
     */
    private String phone;

    /**
     * user's email'
     */
    private String email;
 ...
}
```

### DIFF
--- a/jOOQ-codegen/src/main/java/org/jooq/codegen/JavaGenerator.java
+++ b/jOOQ-codegen/src/main/java/org/jooq/codegen/JavaGenerator.java
@@ -5519,11 +5519,15 @@ public class JavaGenerator extends AbstractGenerator {
             out.println();
 
             if (!generatePojosAsJavaRecordClasses())
-                for (TypedElementDefinition<?> column : getTypedElements(tableUdtOrEmbeddable))
+                for (TypedElementDefinition<?> column : getTypedElements(tableUdtOrEmbeddable)){
+                    //add column's comment to pojo's field
+                    out.javadoc("[[%s]]", list(escapeEntities(comment(column))));
                     out.println("private %s%s %s;",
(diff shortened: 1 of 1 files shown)


lukaseder:
Thanks for your suggestion.

I will reject this PR (as ever so often, see our contributing guidelines: https://github.com/jOOQ/jOOQ/blob/main/CONTRIBUTING.md), because it is overlooking a lot of additional work that would need to be done, should we choose to do it.

The fact that we don't have these comments on properties / members isn't an oversight, but just didn't seem necessary so far. There may be arguments in favour or against it, but if we do this, this must be done:

- For all supported languages
- For all types of POJO
- Made configurable with a sensible default

If you think this is worth pursuing, please create a feature request stating why you'd like this.

PR #15837, opened 2023-11-15
Code gen add field comment for pojo

nianien:
I am a huge fan of joq, and I use it heavily in my work. I hope that joq continues to improve. As a complement for 'https://github.com/jOOQ/jOOQ/pull/15831', I have submitted a new PR, which supports kotlin,scala and java record class, hoping that it will be accepted.


lukaseder:
Thanks for your suggestion. Can you please read this first:

The essential part of my feedback from the above link and the previous PR (https://github.com/jOOQ/jOOQ/pull/15831#issuecomment-1812269103) is the fact that I won't necessarily choose to prioritise this feature, or choose to implement it at all.

Having a PR available won't change that, au contraire. Reviewing PRs can be much more work than just doing it myself, and I don't want to spend the time on this, nor prioritise it right now.

nianien:
thank you for replying, apologize for wasting you time~

lukaseder:
No worries. If github allowed to turn off the PR feature, or require a more complex workflow, then we'd do that. Regrettably, github allows for anyone to create a PR without prior process. It's easy to overlook the contributing guidelines...

PR #16021, opened 2024-01-07
Add example for using jOOQ in Spring Boot with generated DAO

SightStudio:
I spent some time dealing with the auto-generated Spring Dao class.

So I added examples for using auto-generated Spring Dao class in Spring Boot
to prevent others from spending time like me.

I moved the generated DSL to ```src/main/java```.
So others can see how it goes.


jOOQ-examples/jOOQ-spring-boot-example/src/main/java/org/jooq/example/spring/repository/BookRepository.java
```diff
@@ -0,0 +1,22 @@
...
+import org.jooq.DSLContext;
+import org.jooq.example.db.h2.tables.daos.BookDao;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public class BookRepository extends BookDao {
+
+  private final DSLContext dslContext;
+
+  public BookRepository(Configuration configuration, DSLContext dslContext) {
```
SightStudio, line 13: This generator must be required.

or Genreated method in dao will throw ```org.jooq.exception.DetachedException``` like below

image

jOOQ-examples/jOOQ-spring-boot-example/src/main/java/org/jooq/example/spring/repository/BookRepository.java
```diff
@@ -0,0 +1,22 @@
...
+    super(configuration);
+    this.dslContext = dslContext;
+  }
+
+  @Override
+  public DSLContext ctx() {
```
SightStudio, line 19: this one was more tricky

Every time we call the dao-generated method. it creates a new DSLContext instance.

How about removing ```final``` in DefaultConfiguration?
https://github.com/jOOQ/jOOQ/blob/main/jOOQ/src/main/java/org/jooq/impl/DefaultConfiguration.java#L347

it makes other people configure more easy.
chose loading DSL like this.

image

lukaseder:
Thanks for your suggestion.

This suggests using inheritance rather than composition, which isn't something I'd endorse, and as such, I wouldn't want to maintain an example based on this approach. A generated `DAO` can be injected into a `Repository`, which seems better in my opinion.

SightStudio:
@lukaseder okay. I got it.

but this one remains a problem. 
https://github.com/jOOQ/jOOQ/pull/16021#discussion_r1444039363

how do I deal with this problem? 
it will create a new DSLContext instance either having generate dao in composition way

lukaseder:
Why would you override `DefaultConfiguration` to let it reference a `DSLContext` in the first place? You don't need that. A `DSLContext` *wraps* a `Configuration`, not the other way round.

SightStudio:
@lukaseder 

I thought it was the easy way to solve the problem that prevent creating a new DSLContext instance.
because I already made an instance of DSLContext as spring bean.

the proper way is adding DslContext filed as composition in DaoImpl and injecting it by the constructor.
but JavaGenerator was quite complicated.

image

lukaseder:
> I thought it was the easy way to solve the problem that prevent creating a new DSLContext instance.

Why do you want to prevent that? It's just a simple wrapper. It has no cost. What made you think this would be a problem at all?

SightStudio:
@lukaseder 

thanks a lot. I thought it had cost. 
that's all good for me. 👍 

----
It seems necessary to leave an example of what I received answer from Lukas.
can I make another PR in a composition way?

lukaseder:
> other people will be confused like me either can I make another PR in a composition way?

I don't like maintaining too many examples. There are so many combinations of features and tools that people may be interested in, we can't maintain an example for each combination. Why not just ask ([and answer](https://stackoverflow.blog/2011/07/01/its-ok-to-ask-and-answer-your-own-questions)) a question on stack overflow?

SightStudio:
ok. I'll post it on Stack Overflow. thanks for giving answer.

Issue #8439, opened 2019-03-22
PostgreSQL: array(<domain over UUID>) causes exception


pbillen:
jOOQ has problems with the following `array()` queries:

* `select array(<select query>)`
* `select array(<column in t>) from t`
* ...

This only happens when the columns are a domain over an UUID. Regular/direct UUIDs work fine. I did not verify, but maybe/probably this issue is also present in the `array_agg()` function.

We get the following exception:

```
org.jooq.exception.DataTypeException: Cannot convert from 4dd16c6d-8134-422e-9781-b4861b97f7a3 (class org.postgresql.util.PGobject) to class java.util.UUID
	at org.jooq.tools.Convert$ConvertAll.fail(Convert.java:1166)
	at org.jooq.tools.Convert$ConvertAll.from(Convert.java:1055)
```

### Versions:

- jOOQ: 3.11.9
- Java: 8
- Database (include vendor): PostgreSQL

pbillen:
Please find a self-contained example in the referenced repository (https://github.com/pbillen/playground-jooq-8439).

You need a (local) Docker daemon to run this. Executing `mvn clean install -U` should be enough.

The relevant files are:

* https://github.com/pbillen/playground-jooq-8439/blob/master/src/test/docker/postgresql/schema.sql
* https://github.com/pbillen/playground-jooq-8439/blob/master/src/test/java/T.java

Let me know if you need more info from me ;).

lukaseder:
Thank you very much for your report. We currently do not have any explicit support for PostgreSQL's domains yet, see #681. This means that our integration tests are unaware of any caveats that may appear when integrating with them.

In my opinion, however, the presence of a domain (which is really just a named, global check constraint) should not affect the data type that is produced by the JDBC driver. If the JDBC driver can produce `UUID` and `UUID[]` types without domains, I would suggest it is a bug for the JDBC driver to fall back to `PGobject` in this case.

Can you try to see what happens when you run similar queries with JDBC directly? And if you think there's wrong behaviour by the driver, perhaps report the issue here as well: https://github.com/pgjdbc/pgjdbc

As a workaround, you might probably be able to manually cast your `id_uuid` type to `uuid`, or `id_uuid[]` to `uuid[]`

lukaseder:
Finally looking into this now. `DOMAIN` types have long been supported (#681), so it would be safe to expect this to work. Yet, I'm still getting this stack trace in the logs:

```
[...]

PR #16189, opened 2024-01-26
Performance improvement: use getClass() to check if object is already a UUID

jrust:
When we upgraded from 3.16.0 to newer versions we had several of our performance tests start failing. Working through the pointfix versions of jOOQ we found that is the work done as part of #8439. The instanceof check combined with pattern matching to create a new variable results in a performance degradation when dealing with a large number of UUIDs. This solution uses the same method as `convert` by using `getClass` and doing the cast in the return statement.

After implementing this change our performance problems were resolved.

### DIFF
--- a/jOOQ/src/main/java/org/jooq/impl/DefaultBinding.java
+++ b/jOOQ/src/main/java/org/jooq/impl/DefaultBinding.java
@@ -5001,8 +5001,8 @@ public class DefaultBinding<T, U> implements Binding<T, U> {
 
                     if (o == null)
                         return null;
-                    else if (o instanceof UUID u)
-                        return u;
+                    else if (o.getClass() == UUID.class)
+                        return (UUID) o;
                     else
                         return Convert.convert(o.toString(), UUID.class);
                 }


lukaseder:
I'm not aware of any such significant problems with `instanceof` checks, as opposed to `getClass()` calls. Can you show profiling data, benchmark results, etc? How is it possible that such a simple thing causes a significant difference in an overall behaviour?

jrust:
Good question. I don't know the JVM internals and my use of `getClass()` was to go back to what `Convert.convert()` does. I'm not sure it's `instanceof` or the pattern matching that creates the new variable `u`. Both should be very fast, but in addition to our builds I created a test to profile just this change. I created a table with 50 million records with a UUID column. I then ran the following Kotlin program:
```
                val executor = DSL.using(connection)
                var sums = mutableListOf<Long>()
                for (n in 1..30) {
                    val sw = com.google.common.base.Stopwatch.createStarted()
                    var i = 0
                    executor.select(ITEMS.GUID)
                            .from(ITEMS)
                            .stream().use {
                                it.forEach {
                                    i++
                                }
                            }
                    sw.stop()
                    sums.add(sw.elapsed(TimeUnit.MILLISECONDS))
                }
                System.out.println("Average: ${sums.average()}")
```

I ran it against 3.19.3 and 3.19.3 with my patch. Without the patch the average was 2001.6ms and with the patch it was 1874.3ms yielding a 6.4% performance improvement.

lukaseder:
Well, it would be *very* surprising if such a simple change would have such a big impact compared to fetching 50M records over the wire (assuming PostgreSQL). I've quickly profiled your loop. This is the time spent on data transfer (57%):

image

And this is the time spent reading the data into UUIDs (within pgjdbc):

image

I don't see a bottleneck in the `DefaultBinding` class, though obviously, simple profiler sampling isn't accurate enough to be certain. But a 6.4% improvement seems unlikely. In fact, I can't produce reliable performance measurements with a simple loop. You can try JMH benchmarks, which are good to rule out measurement errors even for non-micro-benchmarks.

I'd like to know more about your performance tests. How do you measure things? Why did you conclude that this particular change would change everything?

In any case, I'm rejecting this PR. I don't think we can start tampering with such details in jOOQ's internals without *very clear* hints at them being *significant* problems. Otherwise, we'll be chasing single-line changes due to perceived "performance impact" all the time (while probably not solving actual problems at all).

lukaseder:
I mean, here's a JMH benchmark:

```java
@Fork(value = 3)
@Warmup(iterations = 3, time = 3)
@Measurement(iterations = 5, time = 3)
public class PGQueryBenchmark {

    @State(Scope.Benchmark)
    public static class BenchmarkState {

        Connection connection;
        DSLContext ctx;

        @Setup(Level.Trial)
        public void setup() throws Exception {
            try (InputStream is = BenchmarkState.class.getResourceAsStream("/config.postgres.properties")) {
                Properties p = new Properties();
                p.load(is);
                Class.forName(p.getProperty("db.postgres.driver"));
                connection = DriverManager.getConnection(
                    p.getProperty("db.postgres.url"),
                    p.getProperty("db.postgres.username"),
                    p.getProperty("db.postgres.password")
                );
            }

            ctx = DSL.using(connection, new Settings()
                .withExecuteLogging(false)
                .withRenderSchema(false));
        }

        @TearDown(Level.Trial)
        public void teardown() throws Exception {
            connection.close();
        }
    }

    @Benchmark
    public List<UUID> uuid(BenchmarkState state) {
        return state.ctx.fetchValues(select(DSL.uuid()).from(generateSeries(1, 10000)));
    }
}
```

Here are results before and after your change:

```
PGQueryBenchmark.uuid  thrpt   15  73.581 ┬▒ 1.796  ops/s
PGQueryBenchmark.uuid  thrpt   15  74.520 ┬▒ 1.378  ops/s
```

I can't reproduce any such significant change. The improvement is within the expected error range for such few run repetitions, so I wouldn't bet on it.

lukaseder:
And here's what I'm getting if I run the benchmark with jOOQ 3.16.0 instead of 3.20.0-SNAPSHOT:

```
PGQueryBenchmark.uuid  thrpt   15  75.566 ┬▒ 0.812  ops/s
```

There does indeed seem to be some marginal slowdown, which might be worth investigating, but it's probably not 6%, nor do I think it will be fixed by your suggestion.

jrust:
Ok, thanks for looking into it. Much appreciated.

PR #16559, opened 2024-04-08
[jOOQ/jOOQ] Adding nullOnRecordAllNull and nullOnRecordAnyNull functions to Functions class

kacetal:
Description:

This PR introduces two new utility functions, nullOnRecordAllNull and nullOnRecordAnyNull, to the Functions class. These functions provide convenient ways to handle scenarios where a Record object contains fields that may be null, allowing users to determine if all fields are null or if any field is null and return null accordingly.

nullOnRecordAllNull evaluates whether all fields within the Record object are null. If all fields are null, it returns null; otherwise, it executes the provided function.

nullOnRecordAnyNull, on the other hand, checks if any field within the Record object is null. If any field is null, it returns null; otherwise, it executes the provided function.

[Link](https://stackoverflow.com/questions/78263705/how-to-handle-null-values-in-mapping-with-jooqs-row-method-in-select-queries) to Stack Overflow question.

Please review and consider this contribution.
Thank you.

### DIFF
--- a/jOOQ/src/main/java/org/jooq/Functions.java
+++ b/jOOQ/src/main/java/org/jooq/Functions.java
@@ -48,6 +48,25 @@ public final class Functions {
 
     /**
      * A function that short circuits the argument function returning <code>null</code>
+     * if all record fields are <code>null</code>.
+     */
+    @NotNull
+    public static final <TR extends Record, R> Function1<TR, R> nullOnRecordAllNull(Function1<? super TR, ? extends R> function) {
+        return (t1) -> {
+            if (t1 == null) {
+                return null;
+            }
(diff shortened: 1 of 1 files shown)


lukaseder:
As often, a PR is premature when the requirements aren't properly understood yet. Additionally:

- A higher order function that is about `Record` types would belong in `Records`, not in `Functions`
- It is unclear to me why this higher order function is needed at all, given that my example was about using `Row[x].mapping()`, which accepts a `Function[x]<T1, T2, R>`, not a function accepting records

Additionally, I don't understand how my answer in the linked Stack Overflow question isn't enough:
https://stackoverflow.com/a/78267752/521799

Please provide a complete example showing exactly what you need to do.

PR #16668, opened 2024-05-09
Add jakarta.generated.annotation.Generated as possible GeneratedAnnotationType

tamaro-skaljic:


tamaro-skaljic:
What's missing is a check in DETECT_FROM_JDK whether the [jakarta.annotation-api](https://mvnrepository.com/artifact/jakarta.annotation/jakarta.annotation-api) dependency is in the classpath, but I think that's only possible when executing the JavaGenerator from inside the application. When used through the jooq-codegen-maven plugin, an option would be to run the https://maven.apache.org/plugins/maven-dependency-plugin/ through https://github.com/mojo-executor/mojo-executor. However, this is beyond my capabilities and would be a breaking change, as DETECT_FROM_JDK should then rather be renamed to DETECT_FROM_CLASSPATH.

jOOQ-checker/pom.xml
```diff
@@ -7,7 +7,7 @@
     <parent>
         <groupId>org.jooq</groupId>
         <artifactId>jooq-parent</artifactId>
-        <version>3.20.0-SNAPSHOT</version>
+        <version>3.20.1-SNAPSHOT</version>
```
lukaseder, line 10: Why upgrade the version?

1. It's unrelated to your actual change
2. It's an incorrect update. 3.20.0 hasn't been released

jOOQ-codegen/src/main/java/org/jooq/codegen/Generator.java
```diff
@@ -232,44 +231,37 @@ public interface Generator {
...
-     * Whether the {@link javax.annotation.processing.Generated} annotation
-     * should be generated.
+     * Whether a {@code Generated} annotation should be generated.
      */
     void setGenerateGeneratedAnnotation(boolean generateGeneratedAnnotation);
 
     /**
-     * Whether the {@link javax.annotation.processing.Generated} annotation
-     * should be generated.
+     * Whether a {@code Generated} annotation should be generated.
```
lukaseder, line 239: Please don't format code with a change. See also:

jOOQ-codegen/src/main/java/org/jooq/codegen/example/JVMArgsGeneratorStrategy.java
```diff
@@ -41,7 +41,7 @@
 import org.jooq.meta.Definition;
 
 /**
- * A generator strategy that prefixes / suffixes class names and other artefacts
+ * A generator strategy that prefixes / suffixes class names and other artifacts
```
lukaseder, line 44: artefact is a totally fine British English spelling of the word.

lukaseder:
Thanks for your suggestion. I'll reject it for it has too many non-essential changes. Though I do agree we can add another value to the enum to support this annotation type. I've created a feature request for this:

- https://github.com/jOOQ/jOOQ/issues/16670

> What's missing is a check in DETECT_FROM_JDK whether the [jakarta.annotation-api](https://mvnrepository.com/artifact/jakarta.annotation/jakarta.annotation-api) dependency is in the classpath

I don't think that's necessary or reasonable. The jakarta version of this annotation can be on the classpath for any reason, including transitive dependencies, so I don't think considering this will lead to any expected behaviour.

> However, this is beyond my capabilities and would be a breaking change, as DETECT_FROM_JDK should then rather be renamed to DETECT_FROM_CLASSPATH.

We could do this, but then again, it seems overengineered?

jOOQ-checker/pom.xml
tamaro-skaljic: Ah I'm sorry, I didn't know. :)

Issue #16318, opened 2024-02-21
Allow for specifying inputs to the gradle CodegenTask in order to participate in incremental builds


lukaseder:
It's currently not possible to tell the `jOOQ-codegen-gradle` plugin what its inputs are, it always assumes its outputs aren't up to date. See also:

- https://github.com/jOOQ/jOOQ/issues/16275
- https://github.com/jOOQ/jOOQ/issues/16316

It would be useful to make the plugin aware of:

- Liquibase
- Flyway
- Other inputs in some way

mklnwt:
@lukaseder the issue is in using the output to decide whether to update or not.
Gradle itself can clearly decide if the output has changed (i.e. file removed).

In this case you want to use the input to make the decision.

ThoSap:
Like the inputs where declared in the unofficial jOOQ Gradle plugin:
https://github.com/etiennestuder/gradle-jooq-plugin/blob/v9.0/example/configure_jooq_with_flyway/build.gradle#L62-L68

lukaseder:
The "unofficial" jOOQ Gradle plugin can still be used if it is preferred. There's no obligation to do everything like it did.

In this specific case, I think there's a better solution, one that knows more about jOOQ code generation.

- The default is to assume outputs are never up to date (because we don't know)
- If we're using the `DDLDatabase`, `XMLDatabase`, `JPADatabase` then the inputs are known to jOOQ and can be used by the gradle plugin
- If we're using Flyway or Liquibase (the most popular database change management utilities), then there's a possibility to auto-detect their outputs somehow (I'll have to investigate what their gradle plugins are doing)

mklnwt:
The default should assume that the inputs are never up-to-date, not the outputs.
The output check is there to make sure the files outputted by the code generation haven't been deleted or modified.

Otherwise I agree, the input checks should only trigger when using one of the file based generation.

lukaseder:
> The default should assume that the inputs are never up-to-date, not the outputs.

Thanks for the correction. We're all referring to this particular statement in `CodegenTask`:

```java
getOutputs().upToDateWhen(task -> false);
```

So, I suspect that we're saying the same thing, yes?

mklnwt:
The logic should be something like this:

```Kotlin
// Only if the generation is targeted against an actual database set this:
getOutputs().cacheIf(task -> isGenerationFromFile || isGenerationFromMigrationTool)

// DDLDatabase, XMLDatabase, JPADatabase
[...]

PR #16817, opened 2024-06-11
[#16318] Split the CodegenPlugin in two parts

martinqcz:
One plugin is for capabilities and another for conventions.

- jooq-codegen-gradle-base plugin provides all the capabilities
- jooq-codegen-gradle plugin applies base plugin and provides default task configuration

### DIFF
--- a/jOOQ-codegen-gradle/build.gradle.kts
+++ b/jOOQ-codegen-gradle/build.gradle.kts
@@ -39,5 +39,12 @@ gradlePlugin {
             tags.set(listOf("jooq"))
             implementationClass = "org.jooq.codegen.gradle.CodegenPlugin"
         }
+        create("jooqCodegenBasePlugin") {
+            id = "${group}.jooq-codegen-gradle-base"
+            displayName = "jooq-codegen-gradle-base"
+            description = "jOOQ code generation base plugin for Gradle"
+            tags.set(listOf("jooq"))
+            implementationClass = "org.jooq.codegen.gradle.CodegenBasePlugin"
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


lukaseder:
Thanks for your suggestion. I take this is following up with the discussion here:
https://github.com/jOOQ/jOOQ/issues/16318#issuecomment-2158986347

As ever so often, PRs are premature. Please, can you explain why this is needed, first.

martinqcz:
> Thanks for your suggestion. I take this is following up with the discussion here: [#16318 (comment)](https://github.com/jOOQ/jOOQ/issues/16318#issuecomment-2158986347)
> 
> As ever so often, PRs are premature. Please, can you explain why this is needed, first.

No problem. Take it as inspiration. Problem is that `upToDateWhen` adds the predicate to the list of predicates. That list is iterated at runtime and if any of the predicate returns false, the task will be considered as not up-to-date and will be executed. 
Unfortunately Gradle does not provide any way how to remove anything that was added to the list of upToDateWhen predicates, so if the plugin itself adds `false` into the list right after it is applied to project there is no possibility to change the task caching.

martinqcz:
> It will get better, I promise. We're not there yet.

I'm looking forward to

lukaseder:
Fun fact, I came across your "prime" example of convention vs base plugin again, the Java plugin. Look at what Etienne's plugin has to handle:
https://github.com/etiennestuder/gradle-jooq-plugin/blob/v9.0/src/main/groovy/nu/studer/gradle/jooq/JooqPlugin.java#L63

Deprecation, deprecation, deprecation. 🫠

I really fail to believe that there is anything such as "established" best practices anywhere within Gradle. Everyone is just hacking around, including the vendor themselves... So, this is clearly a hint towards keeping it simple, at least in our case.

PR #16982, opened 2024-07-22
Allow specifying multiple paths for DDLDatabase

atipls:
### Changes:

Currently DDLDatabase only allows specifying one ant-style path for scripts, which makes the following unmatchable:
```
src/main/resources/db/migration/01_generated_schema/*
src/main/resources/db/mysql/*
```

This PR allows specifying multiple paths to match, with a configurable delimiter


lukaseder:
Thanks for your suggestion.

I can see how this is a desireable feature, but I don't think this will be the way forward. There are numerous popular approaches to supporting multiple directories in different environments, and I'd like to research this a bit more, first. Additionally, I don't think it will be `DDLDatabase` specific.

For the future, it's usually better to make a feature request rather than a PR, see:

For now, I've created this feature request here:

- https://github.com/jOOQ/jOOQ/issues/16987

Please note that you can always use multiple code generation configurations as well, or copy files around with pre-processing Maven plugins prior to consuming them with `DDLDatabase`.

Issue #17515, opened 2024-10-30
Regression in code generation for MariaDB where unsignedTypes=false is ignored


guss77:
### Expected behavior

Given a table schema such as
```
CREATE TABLE IF NOT EXISTS `users` (
  `id` INT UNSIGNED NOT NULL AUTO_INCREMENT,
  `name` VARCHAR(45) NOT NULL,
  `email` VARCHAR(45) NOT NULL,
  PRIMARY KEY (`id`))
ENGINE = InnoDB;
```

and having the generator configured with
```
<database>
	<name>org.jooq.meta.mariadb.MariaDBDatabase</name>
	<includes>.*</includes>
	<unsignedTypes>false</unsignedTypes>
</database>
```

I'd expect the POJO to look like:
```
/**
 * This class is generated by jOOQ.
 */
@SuppressWarnings({ "all", "unchecked", "rawtypes", "this-escape" })
public class Users implements Serializable {

    private static final long serialVersionUID = 1L;

    private Integer id;
    private String name;
    private String email;
...
```

### Actual behavior

Instead, the generator create a POJO as thus
```
/**
 * This class is generated by jOOQ.
 */
@SuppressWarnings({ "all", "unchecked", "rawtypes", "this-escape" })
public class Users implements Serializable {

    private static final long serialVersionUID = 1L;

    private UInteger id;
    private String name;
    private String email;
...
```

### Steps to reproduce the problem

I'm using the vertx-jooq library, with it - and some custom converters and types, the generator configuration looks as thus:
```
<generator>
<name>io.cloudonix.lib.model.support.ExtendedClassicJDBCGuiceVertxGenerator</name>
<strategy>
	<name>io.cloudonix.lib.model.support.ExtendedVertxGeneratorStrategy</name>
</strategy>
<database>
	<name>org.jooq.meta.mariadb.MariaDBDatabase</name>
	<includes>.*</includes>
	<inputSchema>api_core</inputSchema>
	<outputSchema>api_core</outputSchema>
	<unsignedTypes>false</unsignedTypes>
	<forcedTypes>
		<!-- Convert tinyint to boolean -->
		<forcedType>
			<name>BOOLEAN</name>
			<includeTypes>(?i:TINYINT\(1\))</includeTypes>
		</forcedType>
		<!-- Convert varchar column with name 'profile' to a io.vertx.core.json.JsonObject -->
		<forcedType>
			<userType>io.vertx.core.json.JsonObject</userType>
[...]

PR #17516, opened 2024-10-30
Fix issue #17515 - do not ignore unsignedTypes=false for types with a precision specifier

guss77:
Any type with the `unsigned` specifier in their name - regardless of where it appears - should be mangled when `database.supportsUnsignedTypes()` is `false`. Specifically types with a precision specifier escaped mangling so far.

I think a cleaner approach will probably be to just try to mangle (if unsigned types are disabled) and it will just not do anything if it doesn't need to:

```
if (!database.supportsUnsignedTypes())
  typeName = P_UNSIGNED.matcher(typeName).replaceFirst("");
```

But that was too much of a change for a drive-by commit. Please let me know if you want it like that anyway.

### DIFF
--- a/jOOQ-meta/src/main/java/org/jooq/meta/DefaultDataTypeDefinition.java
+++ b/jOOQ-meta/src/main/java/org/jooq/meta/DefaultDataTypeDefinition.java
@@ -131,7 +131,7 @@ public class DefaultDataTypeDefinition implements DataTypeDefinition {
         this.schema = schema;
 
         // [#519] [#17135] Some types have unsigned versions
-        if (!database.supportsUnsignedTypes() && typeName.toLowerCase().endsWith("unsigned"))
+        if (!database.supportsUnsignedTypes() && P_UNSIGNED.matcher(typeName).find())
             typeName = P_UNSIGNED.matcher(typeName).replaceFirst("");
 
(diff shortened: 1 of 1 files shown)


guss77:
Sorry, forgot to mention in a comment body (otherwise Github can't see it): Fixes issue #17515 .

lukaseder:
Thanks for your suggestion.

I prefer to keep a cheaper check before regex pattern matching in general. It's probably not too expensive in this case, but it was, many times, elsewhere.

Replacing `endsWith()` by `contains()` should do the trick.

guss77:
Regular expressions are supposed to be very fast (at least in a lot of implementations I've used), so I did a small benchmark - under the best conditions the Java `Pattern` matchers is 3 times as slow as `String.contains()`. This is very annoying, and it shouldn't be that slow, but :shrug:

lukaseder:
> This is very annoying, and it shouldn't be that slow, but 🤷

Yeah, I know. Never call `String.replaceAll()` in hot loops. Always (if you must), pre-compile the pattern, first.

lukaseder:
Again, it doesn't matter here, probably (this being the code generator), but I've made it a habit always preventing unnecessary regex lookups given the issues jOOQ had in the past.

PR #17670, opened 2024-11-26
Performance improvement in BatchSingle class

ggsurrel:
Currently the Javadoc of `BatchBindStep` interface states that calling `BatchBindStep bind(Map<String, Object>... namedBindValues)` is the same as calling `BatchBindStep bind(Map<String, Object> namedBindValues)` several times: https://github.com/jOOQ/jOOQ/blob/d0cdd0c8200e7175477727615cf15425645f90b7/jOOQ/src/main/java/org/jooq/BatchBindStep.java#L140

This is not true performance wise, I noticed a significant performance boost when using the first one. This was due to the numerous calls to `extractBindValues` when using the second one several times.

### DIFF
--- a/jOOQ/src/main/java/org/jooq/impl/BatchSingle.java
+++ b/jOOQ/src/main/java/org/jooq/impl/BatchSingle.java
@@ -86,6 +86,7 @@ final class BatchSingle extends AbstractBatch implements BatchBindStep {
     final Map<String, List<Integer>> nameToIndexMapping;
     final List<Object[]>             allBindValues;
     final int                        expectedBindValues;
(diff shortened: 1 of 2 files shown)


lukaseder:
Thanks for your suggestion. There's a slight risk of introducing a regression in some edge cases, where users call `bind(Map)` consecutively while modifying the underlying `Query` (which is possible in jOOQ, regrettably). I've described the scenario in this issue here:
- https://github.com/jOOQ/jOOQ/issues/17671

With this risk (even if just very slight), I won't backport the fix. But I agree with your suggestion.

Issue #17873, opened 2025-01-15
Avoid case sensitive access to PgDatabaseMetaData columns due to regressions after driver version 42.7.5


rasa-app:
### Actual behavior

When I update postgresql jdbc drive to 42.7.5 code generator throws exception:

```
Caused by: java.lang.IllegalArgumentException: Field "key_seq" is not contained in row type (
  "PKTABLE_CAT",
  "pg_namespace"."PKTABLE_SCHEM",
  "pg_class"."PKTABLE_NAME",
  "pg_attribute"."PKCOLUMN_NAME",
  "FKTABLE_CAT",
  "pg_namespace"."FKTABLE_SCHEM",
  "pg_class"."FKTABLE_NAME",
  "pg_attribute"."FKCOLUMN_NAME",
  "KEY_SEQ",
  "UPDATE_RULE",
  "DELETE_RULE",
  "pg_constraint"."FK_NAME",
  "pg_class"."PK_NAME",
  "DEFERRABILITY"
)
```

### Steps to reproduce the problem

Upgrade your postgresql jdbc driver to 42.7.5

### jOOQ Version

JOOQ free 3.19.17

### Database product and version

PostgreSQL 16

### Java Version

zulu 22.0.2

### JDBC / R2DBC driver name and version (include name if unofficial driver)

org.postgresql:postgresql:42.7.5

lukaseder:
Thanks a lot for your report. I can reproduce this problem. It was introduced in 42.7.5 with this commit:

- https://github.com/pgjdbc/pgjdbc/commit/17cac52fe9cb80001cfd1fd72ec2b03fa4c5d64e

The commit contains numerous fixes to the case of JDBC `DatabaseMetaData` result columns, e.g.:

- https://github.com/pgjdbc/pgjdbc/commit/17cac52fe9cb80001cfd1fd72ec2b03fa4c5d64e#diff-0571f8ac3385a7f7bb34e5c77f8afd24810311506989379c2e85c6c16eea6ce4L2746

Knowing this is a thing, we must urgently switch to case insensitive reading of `DatabaseMetaData` results within jOOQ. I'll fix this and release patch versions later this week, as a ton of users will run into problems very soon after their upgrades of pgjdbc

lukaseder:
The runtime `MetaImpl` isn't affected, because there, we use column indexes to access `DatabaseMetaData` columns. So far, I haven't found any JDBC driver that didn't get those indexes right, as opposed to the column names.

[...]

PR #18150, opened 2025-03-13
Prevent swallowed Exceptions on source file generation for tables.

balk-sp:
Exceptions during source file generation do lead to missing files (e.g. `{MyTable}Record.java` would be missing), which is a critical issue and should be propagated to the call-site. I don't think there is a use-case where we want to continue and generate a partial file-set after the generator has executed.

We encountered this swallowed Exception when we updated our postgres-driver and ran into https://github.com/jOOQ/jOOQ/issues/17873 - We upgraded to jOOQ 3.19.21 and that fixed the root-cause, but the error-handling should fail-fast, abort the operation and make the call-site aware, so that it could, for example, fail a build-pipeline.

### DIFF
--- a/jOOQ-codegen/src/main/java/org/jooq/codegen/JavaGenerator.java
+++ b/jOOQ-codegen/src/main/java/org/jooq/codegen/JavaGenerator.java
@@ -1663,6 +1663,7 @@ public class JavaGenerator extends AbstractGenerator {
             }
             catch (Exception e) {
                 log.error("Error while generating table record " + table, e);
+                throw new RuntimeException("Error while generating table record " + table, e);
             }
         }
 


lukaseder:
Thanks for your suggestion. We cannot merge such PRs, because they're almost certainly wrong / incomplete / etc. See this resource for more details:

- https://github.com/jOOQ/jOOQ/blob/main/CONTRIBUTING.md

We can discuss the actual problem and potential solutionn

lukaseder:
> I actually see that the approach of merely logging, is done in quite a few methods in `JavaGenerator.java`, like `generateDaos()`, `generateInterfaces()`, `generatePojos()`, etc etc

Yeah, that's why I said the PR is incomplete. But rather than fixing (and discussing) the PR, let's discuss the issue.

> Would you agree that in all these cases throwing an Exception would be preferable?

No, I wouldn't. There's a reason for the status quo. We could argue that the configuration isn't applied correctly, though:
https://www.jooq.org/doc/latest/manual/code-generation/codegen-advanced/codegen-config-onerror/

> Like I said in my first comment, I don't quite see the use-case for leaving the output directory in a state that is known not to work, without feedback (beyond logging) to the call site.

The compiler will tell you. It's a philosophical question. Both approaches are valid.

> Which seems to indicate that we want to propagate issues to the call site, while the behavior of the `generate()` method is to swallowing these issues, even when `OnError.FAIL` is specified in the configuration.

Yes, please just create an issue for this. The PR is the wrong approach. Unfortunately, github won't let me turn off the PR functionality without also turning off the issue tracker.

balk-sp:
Thank you for your quick reply. I did read https://github.com/jOOQ/jOOQ/blob/main/CONTRIBUTING.md prior to making the PR, and read the following line as an invitation to use PRs as a conversation piece - there was no expectation of this getting merged as-is.

> Do you want to show some code in diff/PR form Sure! That's a clever usage of PRs, often better than prose. If there's no expectation of it to be merged, go for it!

My apologies for the misinterpretation. I will create an issue on the issue-tracker instead.

lukaseder:
> > Do you want to show some code in diff/PR form Sure! That's a clever usage of PRs, often better than prose. If there's no expectation of it to be merged, go for it!
> 
> My apologies for the misinterpretation. I will create an issue on the issue-tracker instead.

Ah I see my bad, I should delete that. Thanks for pointing that out.

Issue #18892, opened 2025-08-18
Invalid Require-Capability OSGi MANIFEST header in JDK 21 built artifacts


GeertZondervan:
Starting with jOOQ 3.20.0, the published artifacts include the following header in their `MANIFEST.MF`:

```
Require-Capability: osgi.ee;filter:="(osgi.ee=UNKNOWN)"
```

In earlier releases this was correctly set to something like:

```
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=17))"
```

The value `UNKNOWN` is not a valid OSGi EE identifier. This causes problems in an OSGi runtime, because the framework cannot map UNKNOWN to a valid execution environment version, leading to unresolved requirements and preventing the bundle from being resolved.

It’s unclear what triggered this change or if it is intended. From what I can tell, the Require-Capability header was never explicitly configured in the maven-bundle-plugin instructions in your POMs. (Perhaps the issue comes from the current build chain or JDK setup used to build the official releases?)

A fix is to explicitly configure the requirement in the maven-bundle-plugin instructions, for example:

```
<instructions>
    <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
    <Export-Package>*</Export-Package>

     <!-- [#12408] Some of these are resolved by introspecting reflective usage -->
     <Import-Package>
          org.slf4j;resolution:=optional;version="[1.7,2)",
           com.fasterxml.jackson.databind;resolution:=optional,
           com.fasterxml.jackson.module.kotlin;resolution:=optional,
           com.google.gson;resolution:=optional,
           *
       </Import-Package>
       <Require-Capability>
           osgi.ee;filter:="(&amp;(osgi.ee=JavaSE)(version=17))"
        </Require-Capability>
        <_versionpolicy>[$(version;==;$(@)),$(version;+;$(@)))</_versionpolicy>
</instructions>
```

This ensures that the generated bundle declares a valid execution environment requirement.

lukaseder:
Thanks for your report.

> It’s unclear what triggered this change or if it is intended

Certainly not intended. I need to be reminded with every OSGi issue, why I wouldn't just simply remove the whole meta data :)

> A fix is to explicitly configure the requirement in the maven-bundle-plugin instructions, for example:

I'd rather not implement a fix that simply restores previous accidental behaviour, let alone one that configures a trademark name that I haven't seen in ages (JavaSE). (though this does appear to be what Apache Felix generates out of the box).

[...]

PR #18895, opened 2025-08-19
[#18892] Specify JavaSE in maven-bundle-plugin instructions.

GeertZondervan:
In jOOQ 3.20.0 and upwards the OSGI MANIFEST.MF contains the following invalid header:
```
Require-Capability: osgi.ee;filter:="(osgi.ee=UNKNOWN)"
```

This PR fixes this by setting the header to the one used in previous releases (3.19.x), like:

```
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=17))"
```

### DIFF
--- a/jOOQ/pom.xml
+++ b/jOOQ/pom.xml
@@ -59,6 +59,9 @@
                             com.google.gson;resolution:=optional,
                             *
                         </Import-Package>
+                        <Require-Capability>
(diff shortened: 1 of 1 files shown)


lukaseder:
Thanks for your suggestion. We support more than one JDK version (see https://www.jooq.org/download/support-matrix-jdk), so even if this approach will be chosen, the specific fix won't be the one we'll implement.

PR #19620, opened 2026-01-28
fix: bill of materials project now has packaging=pom

alexanderankin:
This is the only pom referenced by spring-boot-dependencies that either doesn't have this or didn't need this.

### DIFF
--- a/jOOQ-bom/src/build/xsl/post-flatten.xsl
+++ b/jOOQ-bom/src/build/xsl/post-flatten.xsl
@@ -65,5 +65,10 @@
         <xsl:copy>
             <xsl:apply-templates select="@*|node()"/>
         </xsl:copy>
+
+        <xsl:if test="name() = 'description'">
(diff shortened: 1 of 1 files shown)


alexanderankin:
apologies i see that this is generated - how can i run the generation task in maven? so that i can test a fix properly...

alexanderankin:
ok, this is now fixed in the xslt (i remember reading tweets/linkedin about this, always top grade java content on there) - i did see that it is being published with the jar to maven repo but maven doesnt download that so perhaps this is safe. like if maven didnt care to download the jar anyways, maybe we can tell it that there is only a pom and its fine.

lukaseder:
Thanks for your suggestion. I've created an issue to track this:

- https://github.com/jOOQ/jOOQ/issues/19621

I don't agree with the attempted fix:

- It's not "clean" XSLT (can't believe I'm *that* guy, in 2026 😅)
- The fix should go into the source XML, not in the XSLT file. The `xsl:copy` element already takes care of copying all the source content to the target output

alexanderankin:
cheers Lukas, thanks so much!

